PR #50, opened 2020-06-25
MassTransit module registration

jakubfijalkowski:
The module registered periodic services but forgot to register the classes that will be run. Also, there was slight mismatch in `SaveChangesAsync` type and the MT docs weren't linked to, so I changed that.


jakubfijalkowski:
Aand `dependabot.yml` location fix.

marcin-dardzinski:
Wait, `services.AddHostedService` doesn't register it in the container? :open_mouth: Sounds exactly what the method will do

marcin-dardzinski (approved):

jakubfijalkowski:
It does register the `IHostedService`, but since you're registering `PeriodicHostedService<T>` that takes the `T` as ctor parameter, you need to register it also (the `T`). :)

lukaszgarstecki (approved):

marcin-dardzinski:
Oh, okay :stuck_out_tongue:

PR #63, opened 2020-07-08
Clean up EF repositories

### LINKED ISSUE #59: Cleanup in EF repositories
1. Currently we have two kind of repository base classes, https://github.com/leancodepl/corelibrary/blob/c591dc745e83dc2cc92f572d61386ee9a22c8d97/src/Domain/LeanCode.DomainModels.EF/EFRepository.cs#L11 and https://github.com/leancodepl/corelibrary/blob/c591dc745e83dc2cc92f572d61386ee9a22c8d97/src/Domain/LeanCode.DomainModels.EF/EFRepositoryNoOC.cs#L10. They are the same exept the first one works on `IAggregateRoot` and sets `IOptimisticConsurrency.DateModified` and the second on `IAggregateRootWithOptimisticConcurrency`. 
I suggest we remove the second one, and use plain old `x is IOptimisticConcurrency c` to check if currently saved entity is optimistically concurrent. 

2.Using UnitOfWorkBase has no advantages currently, as we haven't dealt with `ISoftDeletable`s for a long time.  https://github.com/leancodepl/corelibrary/blob/c591dc745e83dc2cc92f572d61386ee9a22c8d97/src/Domain/LeanCode.DomainModels.EF/EFUnitOfWorkBase.cs#L10
I suggest we move soft deletables handling logic to some extension method on the `DbContext` and implement `IUnitOfWork` in db contexts directly.

Dragemil:


lukaszgarstecki (approved):

jakubfijalkowski:
Open question - do we want this in 4.2 (used by some of our production projects) or should we target v5? IMO we should freeze v4.2 for now and just focus on v5. /cc: @marcin-dardzinski @Saancreed @lukaszgarstecki

marcin-dardzinski:
Yes, I'd merge it to v5, it's not critical in any way, so let's not disturb current projects.

marcin-dardzinski (approved):

src/Infrastructure/LeanCode.SmsSender/SmsApiClient.cs
```diff
@@ -78,7 +78,7 @@ public async Task SendAsync(string message, string phoneNumber)
                 }
             }
 
-            logger.Information("SMS to {PhoneNumber} sent successfully", phoneNumber);
+            logger.Information("SMS sent successfully");
```
jakubfijalkowski: This shouldn't really be in this PR.

jakubfijalkowski (changes requested):
The commit history is now convoluted. Let's rebase your work on v5 (without the SMS change) first.

src/Infrastructure/LeanCode.SmsSender/SmsApiClient.cs
Dragemil: True, it came with the GH rebase.

Dragemil:
Should be fine by now

jakubfijalkowski (approved):

PR #65, opened 2020-07-09
Cancellation tokens in CQRS pipeline

### LINKED ISSUE #24: `CancellationToken`s in pipelines
Currently, we're ignoring `CancellationToken`s in HTTP/event pipelines. It would be good to support them.

I see two options:
1. Pass `CancellationToken` in context. This is IMO preferred approach as it will mimic what we have in ASP.NET Core's `HttpContext`. It will be also easier to migrate to as it will _just_ require changes in a couple of places.
2. Pass it explicitly through execution calls. This will effectively force change in almost every file that touches pipelines, so... :)

marcin-dardzinski:
This PR adds cancellation tokens in the CQRS pipeline. The token is passed in `IPipelineContext` and is verified before each element execution. Also, cancellation support to repositories was added.

There is _a lot_ of places where cancellation tokens can be added too.
- [x] LeanCode.Dapper
- [x] LeanCode.Facebook
- [x] LeanCode.Mixpanel
- [x] LeanCode.PdfGenerator
- [x] LeanCode.SendGrid
- [x] LeanCode.SmsSender
- [x] LeanCode.Firebase.FCM
- [x] LeanCode.ViewRenderer (just the interfaces)
-  ~LeanCode.ViewRenderer.Razor~ - left for a separate PR

We will see later if we want to have them added in this PR too.


marcin-dardzinski:
Started adding cancellation tokens to other libraries too. Maybe a little out of scope of this PR, but I'll also clean up redundant http client classes for some api clients.

marcin-dardzinski:
OK, I've added cancellation support everywhere I saw possible, with two exceptions.
- `LeanCode.PushNotifications`/`LeanCode.PushNotifications.EF` - do we even want to keep those projects as we use FCM directly?
- `LeanCode.ViewRenderer.Razor` - it looks a little bit more complicated than plain mechanical work, I'd like to tackle it separately. Plus, I hear a lot we need to rewrite it, so we can handle it then. For now, tokens are on the interface level, but they're ignored.

Let me know if there's anything more I've missed.

jakubfijalkowski:
`LeanCode.PushNotifications` / `LeanCode.PushNotifications.EF` - yeah, let's get rid of them.

jakubfijalkowski:
> `LeanCode.PushNotifications` / `LeanCode.PushNotifications.EF` - yeah, let's get rid of them.

Ona second thought - let's get rid of them in a separate PR. Will open the issue.

marcin-dardzinski:
Yeah, that was my idea too :stuck_out_tongue:

jakubfijalkowski (approved):

src/Domain/LeanCode.CQRS.RemoteHttp.Server/BaseRemoteObjectHandler.cs
```diff
@@ -84,6 +84,11 @@ public async Task<ExecutionResult> ExecuteAsync(HttpContext context)
                     obj,
                     type);
             }
+            catch (TaskCanceledException ex)
```
jakubfijalkowski: `OperationCanceledException`, `TaskCanceledException` is when the task _scheduling_ is canceled (that won't occur here).

src/Domain/LeanCode.DomainModels.MassTransitRelay/Simple/SimpleEventsExecutor.cs
```diff
@@ -25,12 +26,16 @@ public SimpleEventsExecutor(IPipelineFactory factory, bool useOutbox)
             exec = PipelineExecutor.Create(factory, config);
         }
 
-        public Task HandleEventsOfAsync(Func<Task> action, Guid? correlationId = null) =>
+        public Task HandleEventsOfAsync(
+            Func<Task> action,
+            Guid? correlationId = null,
+            CancellationToken cancellationToken = default) =>
```
Saancreed: With multiline signature `{ return …; }` looks better than `=> …;` imho.

src/Infrastructure/LeanCode.Mixpanel/MixpanelAnalytics.cs
```diff
@@ -15,81 +17,103 @@ public class MixpanelAnalytics
...
         }
 
-        public Task TrackAsync(string userId, IMixpanelEvent mixpanelEvent, bool isImport = false)
-            => TrackAsync(userId, mixpanelEvent.EventName, mixpanelEvent.Properties, isImport);
+        public Task TrackAsync(
+            string userId,
+            IMixpanelEvent mixpanelEvent,
+            bool isImport = false,
+            CancellationToken cancellationToken = default)
+            => TrackAsync(userId, mixpanelEvent.EventName, mixpanelEvent.Properties, isImport, cancellationToken);
```
Saancreed, line 43: `{ return …; }` here and in a few other places in this file.

src/Domain/LeanCode.DomainModels/DataAccess/IRepositoryExtensions.cs
```diff
@@ -1,16 +1,20 @@
...
 
 namespace LeanCode.DomainModels.DataAccess
 {
     public static class IRepositoryExtensions
     {
-        public static async Task<TEntity> FindAndEnsureExistsAsync<TEntity, TIdentity>(this IRepository<TEntity, TIdentity> repository, TIdentity id)
+        public static async Task<TEntity> FindAndEnsureExistsAsync<TEntity, TIdentity>(
+                this IRepository<TEntity, TIdentity> repository,
+                TIdentity id,
+                CancellationToken cancellationToken = default)
```
Saancreed: Indentation.

Saancreed (approved):
Only a few codestyle nitpicks, otherwise looks fine.

src/Infrastructure/LeanCode.Mixpanel/MixpanelAnalytics.cs
marcin-dardzinski: In this file I'd keep it as is.

PR #66, opened 2020-07-09
Add strongly typed ids

### LINKED ISSUE #64: Add strongly typed Ids
`Id<TEntity>` has already been used in 2 or 3 projects, so it's time to move it to corelibrary :stuck_out_tongue: 
Along with should come:
1. Value converters for EF Core
2. Converters for `System.Text.Json`. I'd use the `JsonConverterAttribute`, so that we don't have to specify it in serializer settings each time

marcin-dardzinski:
Most of this is a just compilation from other projects, with the exception of json serialization.

An open question, since we add this to corelib, do we want to also add `SId` with a string backing field?

### DIFF
--- a/src/Domain/LeanCode.DomainModels/LeanCode.DomainModels.csproj
+++ b/src/Domain/LeanCode.DomainModels/LeanCode.DomainModels.csproj
@@ -1,6 +1,3 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
-    <ItemGroup>
-        <ProjectReference Include="../LeanCode.IdentityProvider/LeanCode.IdentityProvider.csproj" />
-    </ItemGroup>
 </Project>
--- a/src/Domain/LeanCode.DomainModels/LeanCode.DomainModels.csproj
+++ b/src/Domain/LeanCode.DomainModels/LeanCode.DomainModels.csproj
@@ -1,3 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
+    <ItemGroup>
+        <ProjectReference Include="../LeanCode.IdentityProvider/LeanCode.IdentityProvider.csproj" />
+    </ItemGroup>
 </Project>
--- a/src/Domain/LeanCode.DomainModels/Model/Id.cs
+++ b/src/Domain/LeanCode.DomainModels/Model/Id.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Diagnostics;
 using LeanCode.DomainModels.Serialization;
-using LeanCode.IdentityProvider;
 
 namespace LeanCode.DomainModels.Model
 {
@@ -19,7 +18,7 @@ namespace LeanCode.DomainModels.Model
             Value = value;
         }
 
-        public static Id<TEntity> New() => new Id<TEntity>(Identity.NewId());
+        public static Id<TEntity> New() => new Id<TEntity>(Guid.NewGuid());
         public static Id<TEntity> From(Guid id) => new Id<TEntity>(id);
         public static Id<TEntity>? From(Guid? id) => id is Guid v ? new Id<TEntity>(v) : (Id<TEntity>?)null;
 
--- a/src/Domain/LeanCode.DomainModels/Model/Id.cs
+++ b/src/Domain/LeanCode.DomainModels/Model/Id.cs
@@ -10,6 +10,8 @@ namespace LeanCode.DomainModels.Model
     public readonly struct Id<TEntity> : IEquatable<Id<TEntity>>, IComparable<Id<TEntity>>
         where TEntity : class, IIdentifiable<Id<TEntity>>
     {
+        public static readonly Id<TEntity> Empty = default;
+
         public Guid Value { get; }
 
         public Id(Guid value)
@@ -38,7 +40,7 @@ namespace LeanCode.DomainModels.Model
     public readonly struct IId<TEntity> : IEquatable<IId<TEntity>>, IComparable<IId<TEntity>>
         where TEntity : class, IIdentifiable<IId<TEntity>>
     {
-        public static readonly IId<TEntity> EmptyValue = IId<TEntity>.From(0);
+        public static readonly IId<TEntity> Empty = default;
 
         public int Value { get; }
 
--- a/src/Domain/LeanCode.DomainModels/Model/Id.cs
+++ b/src/Domain/LeanCode.DomainModels/Model/Id.cs
@@ -1,0 +1,64 @@
+using System;
+using System.Diagnostics;
+using LeanCode.DomainModels.Serialization;
+using LeanCode.IdentityProvider;
+
+namespace LeanCode.DomainModels.Model
+{
+    [DebuggerDisplay("{Value}")]
+    [TypedIdConverter]
+    public readonly struct Id<TEntity> : IEquatable<Id<TEntity>>, IComparable<Id<TEntity>>
+        where TEntity : class, IIdentifiable<Id<TEntity>>
+    {
+        public Guid Value { get; }
+
+        public Id(Guid value)
+        {
+            Value = value;
+        }
+
+        public static Id<TEntity> New() => new Id<TEntity>(Identity.NewId());
+        public static Id<TEntity> From(Guid id) => new Id<TEntity>(id);
+        public static Id<TEntity>? From(Guid? id) => id is Guid v ? new Id<TEntity>(v) : (Id<TEntity>?)null;
+
+        public bool Equals(Id<TEntity> other) => Value.Equals(other.Value);
+        public int CompareTo(Id<TEntity> other) => Value.CompareTo(other.Value);
+        public override bool Equals(object? obj) => obj is Id<TEntity> id && Value.Equals(id.Value);
+        public override int GetHashCode() => HashCode.Combine(Value);
+        public override string? ToString() => Value.ToString();
+
+        public static bool operator ==(Id<TEntity> left, Id<TEntity> right) => left.Equals(right);
+        public static bool operator !=(Id<TEntity> left, Id<TEntity> right) => !left.Equals(right);
+        public static implicit operator Guid(Id<TEntity> id) => id.Value;
+        public static explicit operator Id<TEntity>(Guid id) => new Id<TEntity>(id);
+    }
+
+    [DebuggerDisplay("{Value}")]
+    [TypedIdConverter]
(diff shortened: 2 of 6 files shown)


Dragemil (approved):

src/Domain/LeanCode.DomainModels/Model/Id.cs
```diff
@@ -0,0 +1,64 @@
...
+        public static implicit operator Guid(Id<TEntity> id) => id.Value;
+        public static explicit operator Id<TEntity>(Guid id) => new Id<TEntity>(id);
+    }
+
+    [DebuggerDisplay("{Value}")]
+    [TypedIdConverter]
+    public readonly struct IId<TEntity> : IEquatable<IId<TEntity>>, IComparable<IId<TEntity>>
+        where TEntity : class, IIdentifiable<IId<TEntity>>
+    {
+        public static readonly IId<TEntity> EmptyValue = IId<TEntity>.From(0);
```
Saancreed: I'm not sure if we need that at all, since it's very easy to obtain such `EmptyValue` with `default(IId<TEntity>)` (and sometimes you will be able to omit the type too, if it can be inferred automagically). Or, if there _is_ need for such static field, then maybe add one to `Guid`–based Id type too?

src/Domain/LeanCode.DomainModels/LeanCode.DomainModels.csproj
```diff
@@ -1,3 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
+    <ItemGroup>
+        <ProjectReference Include="../LeanCode.IdentityProvider/LeanCode.IdentityProvider.csproj" />
+    </ItemGroup>
```
Saancreed: Shouldn't MSBuild files be indented with 2 spaces?

src/Domain/LeanCode.DomainModels/Model/Id.cs
```diff
@@ -0,0 +1,64 @@
+using System;
+using System.Diagnostics;
+using LeanCode.DomainModels.Serialization;
+using LeanCode.IdentityProvider;
+
+namespace LeanCode.DomainModels.Model
+{
+    [DebuggerDisplay("{Value}")]
+    [TypedIdConverter]
```
Saancreed, line 8: Inb4 someone tries to (de)serialize one of these with `Newtonsoft.Json` :smile:

Saancreed (commented):
An additional suggestion: how about adding `type Id<T> = string;` and `type IId<T> = number;` to default preamble of TypeScript contract generator (and an equivalent for Dart as well) to have them correctly recognize strongly typed IDs out of the box?

src/Domain/LeanCode.DomainModels/Model/Id.cs
marcin-dardzinski: This part was copy pasted and I overlooked it :stuck_out_tongue: 
I guess we can add it to `Id` as well to mimic they way Guid behaves, but your point with `default` is valid.

marcin-dardzinski:
Currently we do not have typed ids propagated to clients and it could be tricky, as they'd have to be specified on the contracts level.  
To do this (in a way that has any benefit) we'd have to use the exactly same type both in contracts and domain. I had an idea how to get this, but I don't think we plan to use typed ids that way.

src/Domain/LeanCode.DomainModels/Model/Id.cs
marcin-dardzinski: Damn, MT Relay uses Newtonsoft currently. I may create a converter, but I don't think we'd like to hardcode an attribute here.

Saancreed:
> in a way that has any benefit

Isn't the primary benefit of all of this prevention of assigments of incompatible/distinct ID types? I'd be pretty hard to do so in TypeScript anyway, since it only cares about shape of an object. The preamble I suggested would at least ensure that `Id<Entity>` is correctly interpreted/serialized in web clients, and with strongly typed IDs in contracts we wouldn't have to convert them manually in command validators/handlers.

jakubfijalkowski:
I don't think these Ids should be published in contracts. These are your domain _things_ (well - value objects) and should stay there and only there. Contracts are "pure" and don't care whether the ID is strongly typed or not. It is the responsibility of services (i.e. handlers) to translate "primitive form" into the correct type.

Currently the "conversion process" is primitive (just a cast) but that doesn't mean that it has no value - it forces you to think about what you're doing. With the ID type exposed in contracts we lose the thought process and effectively go to square one - almost the same as using `Guid`s.

It would also introduce strong coupling between contracts and domain which is a terrible, terrible idea (been there, done that, that's PITA). It also gives contracts some "logic" since `Id` type (in the current form) is able to generate itself and contracts shouldn't be able to do that.

Consider another step with strongly-typed IDs - IDs that contain object type as part of the value, i.e. `alert-(HERECOMESGUID)`. Converting `string` into it has non-trivial validation connected to it and it has to be done _somewhere_.

src/Domain/LeanCode.DomainModels/Model/Id.cs
jakubfijalkowski: `LeanCode.DomainModels` don't (and shouldn't) reference `Newtonsoft.Json`. We can provide the converters Json.Net converters but let's use other assembly for that.

I have the same problem with these converters - they don't really belong in the domain.

marcin-dardzinski: Yeah, it might not be the best place, but the alternative is to force every user to add a proper converter factory each time they want to perform the serialization.

Saancreed:
> I don't think these Ids should be published in contracts. These are your domain _things_ (well - value objects) and should stay there and only there. Contracts are "pure" and don't care whether the ID is strongly typed or not. It is the responsibility of services (i.e. handlers) to translate "primitive form" into the correct type.

Ah, right, I forgot it would require contracts to reference domain.

Okay, then we probably don't need JSON converters at all here. Unless someone for some reason wants to store JSONs with IDs in database, but who even does that? :stuck_out_tongue:  

> Currently the "conversion process" is primitive (just a cast) but that doesn't mean that it has no value - it forces you to think about what you're doing. With the ID type exposed in contracts we lose the thought process and effectively go to square one - almost the same as using `Guid`s.

I'd say the equivalent 'thought process' of putting strongly typed ID in a contract has the same value as typing `(Id<T>)command.TId` one to four times per command.

> It would also introduce strong coupling between contracts and domain which is a terrible, terrible idea (been there, done that, that's PITA). It also gives contracts some "logic" since `Id` type (in the current form) is able to generate itself and contracts shouldn't be able to do that.

Well, `Guid` can also generate itself ¯\\\_(ツ)\_/¯

Just because a type is able to do something in some context doesn't mean it _should_ (and by extension, doesn't mean that the type should be forbidden from being used just because it has some additional capabilities).

src/Domain/LeanCode.DomainModels/Model/Id.cs
jakubfijalkowski: This has this advantage that you will need to think about the data (and format) that you're serializing.

I somewhat agree that having this here is just easier and don't introduce any complexity (thanks to it being internal) so I'm OK with it.

jakubfijalkowski:
> I'd say the equivalent 'thought process' of putting strongly typed ID in a contract has the same value as typing (Id<T>)command.TId one to four times per command.

I strongly disagree. Casting is more involved thought process than just passing anything anywhere because it's _explicit_.

> Well, Guid can also generate itself ¯\_(ツ)_/¯

There is a difference between `System`-provided type that you have to use (it's forced on you) and type that you deliberately introduce as part of your codebase.

> Just because a type is able to do something in some context doesn't mean it should

Exactly. And we shouldn't even allow that.

> (and by extension, doesn't mean that the type should be forbidden from being used just because it has some additional capabilities).

This is a really far-fetching "extension". Going further you might say that we should add mapping code to contracts because it shouldn't be used inappropriately.

`Guid` is a primitive type in this context because it is "given". `Id` on the other hand is our code, introduced to keep some invariants that we won't be able to enforce (or it would need to be done by some infra code - which is another terrible idea) if we just publish it.

Saancreed:
> There is a difference between `System`-provided type that you have to use (it's forced on you) and type that you deliberately introduce as part of your codebase.

How exactly is it forced if we can always introduce our own types which don't expose any 'problematic' capabilities? If we truly cared about this, we would have changed `Guid` in contracts to some more dumbed–down version of it long ago.

And another case in point: we are (afaik) fine with using our own `Date` and `Time` types in contracts even though they have `Today`, `UtcToday`, `Now` and `UtcNow` properties which _don't cause any harm if they are not abused_. We could remove them anytime and yet they are still there.

> This is a really far-fetching "extension". Going further you might say that we should add mapping code to contracts because it shouldn't be used inappropriately.

I wouldn't because it would not be the correct place for this logic (not to mention the fact that contracts would need to reference project with types we are mapping to, which is another unacceptable requirement, and one I've already mentioned too).

jakubfijalkowski:
> How exactly is it forced if we can always introduce our own types which don't expose any 'problematic' capabilities?

They are _common_, provided by the framework, used _everywhere_. _Id_ on the other hand is _our own type_, introduced _by us_, used only _by us_. Moreover, it is introduced to _limit our ability to do bad things_ and exposing it in contracts defies the point of using _our own custom type_ - we could just use `Guid`s like in the old days.

Saancreed:
`s/Id/Date/g` and `s/Guid/DateTime/g` and you get another one that doesn't quite hold up there. Just because we expose stuff like this in contracts doesn't mean it suddenly becomes unable to limit our ability to do bad things.

But since we already arrived at the conclusion that it's a bad idea to use `Id<T>` in contracts, let's just leave it at that.

jakubfijalkowski:
`Date` is another `System` provided type, so this argument is invalid :)

Saancreed:
Technically, it's a `System` type that _we_ provide (making it well within our abilities to remove some methods from it), but I see your point.

src/Domain/LeanCode.DomainModels/Model/Id.cs
Saancreed: > We can provide the Json.Net converters but let's use other assembly for that.

I agree, and I wouldn't mind if they came in separate PR.

Saancreed (approved):
There's still the case of `Newtonsoft.Json` converters, but we can leave that one for another time.

jakubfijalkowski (approved):

PR #76, opened 2020-07-17
Remove event handlers

### LINKED ISSUE #73: Remove outdated components
We have some components that we are no longer using (or at least shouldn't use):
1. `LeanCode.PushNotifications` & `LeanCode.PushNotifications.EF`,
2. `LeanCode.IdentityServer.PersistedGrantStore` (we can use IS-provided one now),
3. `LeanCode.DomainModels.EventsExecution` - this will be trickier since MT integration relies on it somewhat,
4. `LeanCode.IdentityProvider` - ?
5. Anything else?

marcin-dardzinski:
Part of https://github.com/leancodepl/corelibrary/issues/73

It removes the original in-proc domain events handling in favour of MT relay.
Most significant changes include:
- remove `LeanCode.DomainModels.EventsExecution` et al.
- remove `IEventsContext` - events interception is now done in the raising events middleware, there is no separate one for interception
- move `AsyncEventsInterceptor` and `SimpleEventsExecutor` to `LeanCode.DomainModels.MassTransitRelay`
- make Hangfire integration use MT Relay when consuming jobs
- allow to opt out from inbox and outbox

This was unit tested (somehow), but I feel we have to make a kind of a prerelease and test how it will behave in one of our projects (but this is a general remark regarding v5.0 version)


src/Domain/LeanCode.DomainModels.MassTransitRelay/Simple/SimpleEventsExecutor.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        private static ConfiguredPipeline<SimplePipelineContext, Func<Task>, ValueTuple> Classic() => Pipeline
+            .Build<SimplePipelineContext, Func<Task>, ValueTuple>()
+            .Use<EventsPublisherElement<SimplePipelineContext, Func<Task>, ValueTuple>>()
+            .Finalize<SimpleFinalizer>();
+
+        private static ConfiguredPipeline<SimplePipelineContext, Func<Task>, ValueTuple> Outboxed() => Pipeline
+            .Build<SimplePipelineContext, Func<Task>, ValueTuple>()
+            .Use<EventsPublisherElement<SimplePipelineContext, Func<Task>, ValueTuple>>()
+            .Finalize<SimpleFinalizer>();
```
Saancreed: This seems to be exactly the same as `Classic()`…

jakubfijalkowski (approved):

Saancreed (approved):

PR #80, opened 2020-07-21
Publish builds as artifacts

### LINKED ISSUE #47: Publish to internal feed only
Currently we're publishing _all_ versions to NuGet and the feed looks [rather bad](https://www.nuget.org/packages/LeanCode.Components) and is quite unusable.

Consider Dependabot's PRs - it creates one PR per dependency and that results in a couple of PRs merged at once. Moreover, we tend to break main branch temporarily (e.g. two consecutive PRs) and publishing everything makes using our library painful.

I propose to publish everything to public Feedz feed "just in case" and publish to NuGet only the versions that we manually mark as "publishable".

marcin-dardzinski:
First step of reworking Github Actions.
The new flow will be:
1. Every build on main branches will publish packages as build artifacts only (as a plain zipfile)
2. Only manually selected builds will be pushed to NuGet - it will be done via manual action dispatch.

Initially I wanted to make a separate workflows for building and publishing, but it has a drawback, you cannot share run numbers with other workflows, so I unified it for versioning consistency. So far, nothing is published to nuget, I have to create another job which would fetch the zip, tag the commit and push it to nuget.

### DIFF
--- a/.github/workflows/core_build_publish.yml
+++ b/.github/workflows/core_build_publish.yml
@@ -3,7 +3,7 @@ name: Corelibrary Build & Publish
 on:
   push:
     branches:
-      - 'v[0-9].[0-9]'
+      - 'v[0-9]+.[0-9]+'
   pull_request:
   workflow_dispatch:
 jobs:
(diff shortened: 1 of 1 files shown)


marcin-dardzinski:
I have to abandon this PR as action checks glitched when I changed the job name :(

PR #81, opened 2020-07-21
Rework CI

marcin-dardzinski:
First step of reworking Github Actions.
The new flow will be:
1. Every build on main branches will publish packages as build artifacts only (as a plain zipfile)
2. Only manually selected builds will be pushed to NuGet - it will be done via manual action dispatch.

Initially I wanted to make a separate workflows for building and publishing, but it has a drawback, you cannot share run numbers with other workflows, so I unified it for versioning consistency.
So now, if you trigger build manually, it will check if you are on the correct branch and publish to NuGet, otherwise it will build and upload packages as artifacts.

### DIFF
--- a/.github/workflows/core_build_publish.yml
+++ b/.github/workflows/core_build_publish.yml
@@ -101,8 +101,8 @@ jobs:
     - name: Create release
       uses: actions/create-release@v1
       with:
-        tag_name: v${{ needs.prepare.outputs.version }}
-        release_name: Release v${{ needs.prepare.outputs.version }}
+        tag_name: ${{ format('v{0}', needs.prepare.outputs.version) }}
+        release_name: ${{ format('Release v{0}', needs.prepare.outputs.version) }}
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         ZIP: ${{ needs.prepare.outputs.artifacts_name }}
--- a/.github/workflows/core_build_publish.yml
+++ b/.github/workflows/core_build_publish.yml
@@ -3,7 +3,7 @@ name: Corelibrary Build & Publish
 on:
   push:
     branches:
-      - 'v[0-9].[0-9]'
+      - 'v[0-9]+.[0-9]+'
   pull_request:
   workflow_dispatch:
 jobs:
--- a/.github/workflows/core_build_publish.yml
+++ b/.github/workflows/core_build_publish.yml
@@ -54,6 +54,7 @@ jobs:
         echo "::set-output name=publish_nuget::${PUBLISH_NUGET}"
         echo "::set-output name=artifacts_name::${ARTIFACTS_NAME}"      
   build:
+    name: Build
     runs-on: ubuntu-18.04
     needs: [prepare]
     env:
--- a/.github/workflows/core_build_publish.yml
+++ b/.github/workflows/core_build_publish.yml
@@ -2,38 +2,70 @@ name: Corelibrary Build & Publish
 
 on:
   push:
-    branches: 
-      - 'v[0-9]+.[0-9]+'
+    branches:
+      - 'v[0-9].[0-9]'
   pull_request:
+  workflow_dispatch:
 jobs:
-  build:
+  prepare:
+    name: Prepare & Version
     runs-on: ubuntu-18.04
-    env:
-      DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1
-      DOTNET_CLI_TELEMETRY_OPTOUT: 1
+    outputs:
+      version: ${{ steps.version.outputs.version }}
+      publish_artifacts: ${{ steps.version.outputs.publish_artifacts }}
+      publish_nuget: ${{ steps.version.outputs.publish_nuget }}
+      artifacts_name: ${{ steps.version.outputs.artifacts_name }}
     steps:
-    - uses: actions/checkout@v2
-    - name: Setup .NET Core
-      uses: actions/setup-dotnet@v1
-      with:
-        dotnet-version: 3.1.202
     - name: Version
+      id: version
       run: |
         BRANCH=${GITHUB_REF#refs/*/}
         if [[ $BRANCH =~ ^v[0-9]+.[0-9]+$ ]]
         then
           BUILD_NUMBER=$(( GITHUB_RUN_NUMBER + 500 )) # compensate for old jenkins CI
           VERSION="${BRANCH#v}.$BUILD_NUMBER"
-          IS_MASTER_BUILD=1
+          PUBLISH_ARTIFACTS=1
+          ARTIFACTS_NAME="LeanCode.CoreLibrary.$VERSION.zip"
         else
           VERSION="0.0.0"
-          IS_MASTER_BUILD=0
+          PUBLISH_ARTIFACTS=0
+          ARTIFACTS_NAME="<none>"
         fi
-        echo Building on "$BRANCH"
-        echo Building version: "$VERSION"
+        echo $BRANCH
+        echo $VERSION
+        echo "Artifacts will be saved as $ARTIFACTS_NAME" 
         
-        echo "::set-env name=VERSION::${VERSION}"
-        echo "::set-env name=IS_MASTER_BUILD::${IS_MASTER_BUILD}"
+        if [[ $GITHUB_EVENT_NAME == 'workflow_dispatch' ]]
+        then
+          echo "Packages will be published to NuGet"
+          PUBLISH_NUGET=1
+        else
+          PUBLISH_NUGET=0
+        fi
+        
+        if [ $PUBLISH_ARTIFACTS == 0 ] && [ $PUBLISH_NUGET == 1 ]
+        then
+          echo "Only vX.Y branches can be published to NuGet, failing"
+          exit 1
+        fi 
+                
+        echo "::set-output name=version::${VERSION}"
+        echo "::set-output name=publish_artifacts::${PUBLISH_ARTIFACTS}"
+        echo "::set-output name=publish_nuget::${PUBLISH_NUGET}"
+        echo "::set-output name=artifacts_name::${ARTIFACTS_NAME}"      
+  build:
+    runs-on: ubuntu-18.04
+    needs: [prepare]
+    env:
+      DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1
+      DOTNET_CLI_TELEMETRY_OPTOUT: 1
+      VERSION: ${{ needs.prepare.outputs.version }}
+    steps:
+    - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


marcin-dardzinski:
Okay, it wasn't a glitch, it's just a target branch protection rule - I'll deal with it one we actually want to merge it

.github/workflows/core_build_publish.yml
```diff
@@ -44,10 +77,36 @@ jobs:
...
-    #   run: find packed/ -name '*.nupkg' -exec dotnet nuget push -k "$NUGET_API_KEY" -s 'https://api.nuget.org/v3/index.json' -n true '{}' ';'
-    #   env:
-    #     NUGET_API_KEY: ${{ secrets.NUGET_API_KEY }}
\ No newline at end of file
+      if: ${{ needs.prepare.outputs.publish_artifacts == '1' }}
+      env:
+        ZIP: ${{ needs.prepare.outputs.artifacts_name }} 
+      run: |
+        dotnet pack --no-build -c Release -o $PWD/packed
+        zip -j $ZIP $PWD/packed/*.nupgk
```
Saancreed: nup**kg**

Also why use `zip` when better alternatives exist (like zstd)? You _should_ be able to compress to that with `bsdtar`, unless Ubuntu 18.04 is too ancient to support that.

.github/workflows/core_build_publish.yml
```diff
@@ -44,10 +77,36 @@ jobs:
...
+      uses: actions/create-release@v1
+      with:
+        tag_name: v${{ needs.prepare.outputs.version }}
+        release_name: Release v${{ needs.prepare.outputs.version }}
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        ZIP: ${{ needs.prepare.outputs.artifacts_name }}
+    - name: Push to NuGet
+      run: |
+        unzip $ZIP
```
Saancreed: Should be quoted to prevent potential word splitting (assuming that's a Bourne shell–like syntax).

.github/workflows/core_build_publish.yml
```diff
@@ -44,10 +77,36 @@ jobs:
...
+        dotnet nuget push -k "$NUGET_API_KEY" -s 'https://api.nuget.org/v3/index.json' *.nupkg
```
Saancreed: `dotnet nuget push` doesn't handle neither multiple input files nor wildcards correctly, which is the entire point of `find -exec` magic.

.github/workflows/core_build_publish.yml
```diff
@@ -2,38 +2,71 @@ name: Corelibrary Build & Publish
...
+    env:
+      DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1
+      DOTNET_CLI_TELEMETRY_OPTOUT: 1
+      VERSION: ${{ needs.prepare.outputs.version }}
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup .NET Core
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: 3.1.202
```
Saancreed: We have `3.1.302` now.

Saancreed (changes requested):

.github/workflows/core_build_publish.yml
jakubfijalkowski: Or even, `5.*`

.github/workflows/core_build_publish.yml
```diff
@@ -44,10 +77,36 @@ jobs:
...
-    #   run: find packed/ -name '*.nupkg' -exec dotnet nuget push -k "$NUGET_API_KEY" -s 'https://api.nuget.org/v3/index.json' -n true '{}' ';'
-    #   env:
-    #     NUGET_API_KEY: ${{ secrets.NUGET_API_KEY }}
\ No newline at end of file
+      if: ${{ needs.prepare.outputs.publish_artifacts == '1' }}
+      env:
+        ZIP: ${{ needs.prepare.outputs.artifacts_name }} 
+      run: |
+        dotnet pack --no-build -c Release -o "$PWD/packed"
+        zip -j "$ZIP" "$PWD/packed/*.nupkg"
```
Saancreed: And now the glob will also be escaped, resulting in nice `no such file: *.nupkg` :upside_down_face:

.github/workflows/core_build_publish.yml
marcin-dardzinski: Well, everyone is familiar with `zip` and probably has a tool to open it. Remember that builds will be also available for download, just in case. Also, what makes alternatives better, given a simple use case?

.github/workflows/core_build_publish.yml
marcin-dardzinski: :angry:

.github/workflows/core_build_publish.yml
marcin-dardzinski: `3.1.302` it is. Will switch when we update to .NET 5

Saancreed: Ah, fair enough, I haven't noticed that this PR targets 5.0 branch.

.github/workflows/core_build_publish.yml
Saancreed: Compression ratio and decompression speed, although I suspect it doesn't matter here as much anyway, so feel free to keep using `zip`.

.github/workflows/core_build_publish.yml
```diff
@@ -2,38 +2,71 @@ name: Corelibrary Build & Publish
...
         else
           VERSION="0.0.0"
-          IS_MASTER_BUILD=0
+          PUBLISH_ARTIFACTS=0
+          ARTIFACTS_NAME="<none>"
         fi
-        echo Building on "$BRANCH"
-        echo Building version: "$VERSION"
+        echo $BRANCH
+        echo $VERSION
```
Saancreed: I don't really see a reason for this change, it was nicely descriptive before.

.github/workflows/core_build_publish.yml
```diff
@@ -2,38 +2,71 @@ name: Corelibrary Build & Publish
...
-        echo "::set-env name=IS_MASTER_BUILD::${IS_MASTER_BUILD}"
+        if [[ $GITHUB_EVENT_NAME == 'workflow_dispatch' ]]
+        then
+          echo "Packages will be published to NuGet"
+          PUBLISH_NUGET=1
+        else
+          PUBLISH_NUGET=0
+        fi
+        
+        if [ $PUBLISH_ARTIFACTS == 0 ] && [ $PUBLISH_NUGET == 1 ]
```
Saancreed: I'd prefer `[[` over `[` or `test` like you used above.

.github/workflows/core_build_publish.yml
```diff
@@ -44,10 +77,36 @@ jobs:
...
-    #   run: find packed/ -name '*.nupkg' -exec dotnet nuget push -k "$NUGET_API_KEY" -s 'https://api.nuget.org/v3/index.json' -n true '{}' ';'
-    #   env:
-    #     NUGET_API_KEY: ${{ secrets.NUGET_API_KEY }}
\ No newline at end of file
+      if: ${{ needs.prepare.outputs.publish_artifacts == '1' }}
+      env:
+        ZIP: ${{ needs.prepare.outputs.artifacts_name }} 
+      run: |
+        dotnet pack --no-build -c Release -o "$PWD/packed"
+        zip -j "$ZIP" $PWD/packed/*.nupkg
```
Saancreed:
```suggestion
        zip -j "$ZIP" "$PWD"/packed/*.nupkg
```

Saancreed (commented):
Some final nitpicks and the rest looks fine to me (but I don't have much GH Actions experience so I'm going to trust you on that :see_no_evil:).

.github/workflows/core_build_publish.yml
marcin-dardzinski: Oh, actually it was copied from a playground repo on which I was testing everything, will revert.

.github/workflows/core_build_publish.yml
```diff
@@ -2,38 +2,71 @@ name: Corelibrary Build & Publish
...
+          ARTIFACTS_NAME="LeanCode.CoreLibrary.$VERSION.zip"
         else
           VERSION="0.0.0"
-          IS_MASTER_BUILD=0
+          PUBLISH_ARTIFACTS=0
+          ARTIFACTS_NAME="<none>"
         fi
-        echo Building on "$BRANCH"
-        echo Building version: "$VERSION"
+        echo $BRANCH
```
marcin-dardzinski:
```suggestion
        echo Building on "$BRANCH"
```

.github/workflows/core_build_publish.yml
marcin-dardzinski:
```suggestion
        echo Building version: "$VERSION"
```

.github/workflows/core_build_publish.yml
marcin-dardzinski:
```suggestion
        if [[ $PUBLISH_ARTIFACTS == 0  &&  $PUBLISH_NUGET == 1 ]]
```

.github/workflows/core_build_publish.yml
```diff
@@ -2,38 +2,71 @@ name: Corelibrary Build & Publish
...
-        echo "::set-env name=IS_MASTER_BUILD::${IS_MASTER_BUILD}"
+        if [[ $GITHUB_EVENT_NAME == 'workflow_dispatch' ]]
+        then
+          echo "Packages will be published to NuGet"
+          PUBLISH_NUGET=1
+        else
+          PUBLISH_NUGET=0
+        fi
+        
+        if [[ $PUBLISH_ARTIFACTS == 0  &&  $PUBLISH_NUGET == 1 ]]
```
Saancreed: Those spaces around the `&&` though :smile:

Saancreed (approved):

.github/workflows/core_build_publish.yml
marcin-dardzinski: Damn, the online editor should have a linter :upside_down_face:

marcin-dardzinski:
```suggestion
        if [[ $PUBLISH_ARTIFACTS == 0 && $PUBLISH_NUGET == 1 ]]
```

.github/workflows/core_build_publish.yml
```diff
@@ -2,38 +2,70 @@ name: Corelibrary Build & Publish
 
 on:
   push:
-    branches: 
-      - 'v[0-9]+.[0-9]+'
+    branches:
+      - 'v[0-9].[0-9]'
```
marcin-dardzinski:
```suggestion
      - 'v[0-9]+.[0-9]+'
```

PR #82, opened 2020-07-22
SendGrid exceptions

### LINKED ISSUE #72: SendGridClient does not throw when tries to send message with empty subject
Expected behaviour: failure

Actual behaviour: result returned has code 400 (no subject)

Saancreed:
Closes #72

### DIFF
--- a/src/Infrastructure/LeanCode.SendGrid/SendGridException.cs
+++ b/src/Infrastructure/LeanCode.SendGrid/SendGridException.cs
@@ -1,0 +1,78 @@
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Net;
+using System.Text.Json;
+using System.Threading.Tasks;
+using SendGrid;
+
+namespace LeanCode.SendGrid
+{
+    public class SendGridException : Exception
+    {
+        private static readonly JsonSerializerOptions SerializerOptions = new JsonSerializerOptions
+        {
+            PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+        };
+
+        public HttpStatusCode ResponseStatusCode { get; }
+        public ImmutableArray<string> ErrorMessages { get; } = ImmutableArray<string>.Empty;
+
+        private SendGridException(
+            HttpStatusCode responseStatusCode,
+            IEnumerable<SendGridError>? errors)
+            : base($"SendGrid request failed with status code {responseStatusCode}.")
+        {
+            ResponseStatusCode = responseStatusCode;
+
+            if (errors is object)
+            {
+                ErrorMessages = errors
+                    .SelectMany(EnumerateMessages)
+                    .ToImmutableArray();
+            }
+
+            static IEnumerable<string> EnumerateMessages(SendGridError error)
+            {
+                if (error.Message is string message)
+                {
+                    yield return message;
+                }
+            }
+        }
+
+        public static async Task ThrowOnFailureAsync(Response response)
+        {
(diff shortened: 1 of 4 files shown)


src/Infrastructure/LeanCode.SendGrid/SendGridException.cs
```diff
@@ -0,0 +1,78 @@
...
+            HttpStatusCode responseStatusCode,
+            IEnumerable<SendGridError>? errors)
+            : base($"SendGrid request failed with status code {responseStatusCode}.")
+        {
+            ResponseStatusCode = responseStatusCode;
+
+            if (errors is object)
+            {
+                ErrorMessages = errors
+                    .SelectMany(EnumerateMessages)
```
marcin-dardzinski: Why not a simple `Select(e => e.Message).Where(not null)`?

marcin-dardzinski (approved):

src/Infrastructure/LeanCode.SendGrid/SendGridException.cs
Saancreed: Because I like to overthink such simple problems :smile:

Saancreed: Also it will cause the type to be inferred as `IEnumerable<string?>`, which `SelectMany` nicely avoids.

marcin-dardzinski: `!` to the rescue!

src/Infrastructure/LeanCode.SendGrid/SendGridException.cs
```diff
@@ -0,0 +1,71 @@
...
+            if (errors is object)
+            {
+                ErrorMessages = errors
+                    .Select(e => e.Message!)
+                    .Where(m => m is object)
+                    .ToImmutableArray();
+            }
+        }
+
+        public static async Task ThrowOnFailureAsync(Response response)
```
jakubfijalkowski: Is this a common pattern? I haven't really seen having these kind of helper methods in `Exception` classes (and I don't think this is the place for these methods).

src/Infrastructure/LeanCode.SendGrid/SendGridException.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        private class SendGridResponse
+        {
+            public SendGridError[]? Errors { get; set; }
+        }
+
+        private class SendGridError
+        {
+            public string? Message { get; set; }
+            // public string? Field { get; set; }
```
jakubfijalkowski: If we don't plan to use these - let's get rid of them.

src/Infrastructure/LeanCode.SendGrid/SendGridException.cs
Saancreed: Maybe not, but where would you move such logic anyway?

src/Infrastructure/LeanCode.SendGrid/SendGridException.cs
Saancreed: Yeah, I added them in case we ever wanted to access those, but looks like it won't be the case anytime soon. Will remove.

src/Infrastructure/LeanCode.SendGrid/SendGridException.cs
jakubfijalkowski: In the client - it is responsible for doing requests and handling results.

jakubfijalkowski (approved):

PR #83, opened 2020-07-22
Asyncify all

marcin-dardzinski:
Initially I wanted to write a smartass regex, but it turns out regexes are not good at working with context-free grammars :thinking: 
So I wrote a simple Roslyn script that does the job for me. I had to correct some things manually, but at least if was fun

### DIFF
--- a/benchmarks/LeanCode.Benchmarks/RemoteCQRS.cs
+++ b/benchmarks/LeanCode.Benchmarks/RemoteCQRS.cs
@@ -61,19 +61,19 @@ namespace LeanCode.Benchmarks
 
         [Benchmark]
         public Task EmptyQuery() =>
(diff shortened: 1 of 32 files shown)


jakubfijalkowski:
There is this beast - [vs-threading](https://github.com/Microsoft/vs-threading) - that has `Async` suffix analyzers :)

marcin-dardzinski:
Well, I sensed that there are tools for it, e.g. ReSharper/Rider, but for the sake of the exercise I didn't look for them. The benefit is that I finally know how to force Roslyn to load MSBuild solution on .NET Core, which was the hardest part of the job :stuck_out_tongue:

jakubfijalkowski (approved):

Saancreed (approved):

PR #86, opened 2020-07-24
Fix artifacts upload

marcin-dardzinski:
image
:sob:


marcin-dardzinski:
After all, I misread the next line of the docs too: "If not provided, artifact will be used as the default name which will manifest itself in the UI after upload.", so we'd have to set the name no matter what

Saancreed (approved):

PR #93, opened 2020-08-06
Bump autofac

marcin-dardzinski:
Kind of a breaking change, but I need it in one of the projects due to other dependency :sob:


lukaszgarstecki (approved):
:cry:

PR #99, opened 2020-08-19
Open Telemetry integration

### LINKED ISSUE #29: OpenTelemetry
Instead of using custom 2x`Guid`-based correlation, we should switch to [OpenTelemetry](https://opentelemetry.io/). OpenTelemetry combines OpenCensus & OpenTracing into one standard that is becoming the de-facto telemetry standard in the industry.

This will also require to re-write Datadog's integration a little bit because it still uses OpenTracing libs (and... well... isn't that .NET Core-native).

marcin-dardzinski:
Dump our current `LeanCode.Correlation` project in favour of `LeanCode.OpenTelemetry` based on OpenTelemetry and `System.Diagnostics`.

Changes include:
- tracing pipeline element, which opens an activity/span for invocation of CQRS pipeline
- Serilog logs enricher enriching logs with current span id and trace id
- MassTransit instrumentation copied from open-telemetry-contrib repo
- propagating of parent activity in events outbox

Still to do:
Datadog exporter. Datadog does not have an exporter for open telemetry exporter for dotnet, we'd have to write one ourselves (not hard, but a little bit tricky though). Also Datadog only accepts 64-bit trace ids, compared to 128-bit standard, but this shouldn't be a problem as long we use Datadog only


src/Domain/LeanCode.DomainModels.MassTransitRelay/Inbox/ConsumedMessagesCleaner.cs
```diff
@@ -28,14 +30,24 @@ public ConsumedMessagesCleaner(IConsumedMessagesContext dbContext)
...
+                var time = Time.Now - KeepTime;
+                var deleted = await dbContext.Self.ExecuteScalarAsync<int>(
+                    $@"DELETE t FROM {tableName} t
                 WHERE t.[DateConsumed] < @time;",
-                new { time },
-                commandTimeout: 3600);
-            logger.Information("Deleted {Count} consumed messages", deleted);
+                    new { time },
+                    commandTimeout: 3600);
+                logger.Information("Deleted {Count} consumed messages", deleted);
```
jakubfijalkowski: Let's change this to `Debug`, spams too much.

jakubfijalkowski (approved):

marcin-dardzinski:
Okay, this is ready for finalization, I think. It is kinda a breaking change so we have to decide how do we handle that.

src/Domain/LeanCode.DomainModels.MassTransitRelay/Outbox/PeriodicEventsPublisher.cs
```diff
@@ -36,6 +39,8 @@ public class PeriodicEventsPublisher : IPeriodicAction
 
         public async Task ExecuteAsync(CancellationToken stoppingToken)
         {
+            using var fetchActivity = LeanCodeActivitySource.Start("outbox.fetch-unpublished");
```
marcin-dardzinski, line 42: On the other hand, activity start could be moved to `PeriodicHostedService` to have it covered once and for all for all actions.

jakubfijalkowski:
IMO we can go with v5.1 as it hasn't really stabilized yet.

PR #100, opened 2020-08-19
Bump identity server to 4.0 and update IdentityService

### LINKED ISSUE #46: Update IdentityServer to v4
There are some major changes under-the-hood but most of the ones that affect us are only visual (method rename). Should be a quick fix (we've already using v4 in one of our projects so if someone needs hints ping me).

Also, we should remove our implementation of PersistedGrantStore - IS.EF provides it OOTB (and we don't need to use in-database clients with it).

Dragemil:
closes #46

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -5,7 +5,7 @@
     <ExtensionsVersion>5.0.0-preview.7.*</ExtensionsVersion>
 
     <HangfireVersion>1.7.11</HangfireVersion>
-    <IdentityServerVersion>3.1.3</IdentityServerVersion>
+    <IdentityServerVersion>4.0.0</IdentityServerVersion>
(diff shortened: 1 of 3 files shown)


lukaszgarstecki (approved):

Directory.Build.targets
```diff
@@ -5,7 +5,7 @@
     <ExtensionsVersion>5.0.0-preview.7.*</ExtensionsVersion>
 
     <HangfireVersion>1.7.11</HangfireVersion>
-    <IdentityServerVersion>3.1.3</IdentityServerVersion>
+    <IdentityServerVersion>4.0.0</IdentityServerVersion>
```
jakubfijalkowski: `4.0.4`

jakubfijalkowski (changes requested):

jakubfijalkowski (approved):

PR #101, opened 2020-08-20
Update key vault configuration

Dragemil:


src/Core/LeanCode.Components.Startup/IWebHostBuilderExtensions.cs
```diff
@@ -16,29 +14,18 @@ public static class IWebHostBuilderExtensions
...
 
-        public const string VaultKey = "Secrets:KeyVault:VaultUrl";
-        public const string ClientIdKey = "Secrets:KeyVault:ClientId";
-        public const string ClientSecretKey = "Secrets:KeyVault:ClientSecret";
-        public const string TenantIdKey = "Secrets:KeyVault:TenantId";
         public const string MinimumLogLevelKey = "Logging:MinimumLevel";
         public const string EnableDetailedInternalLogsKey = "Logging:EnableDetailedInternalLogs";
 
-        public static IWebHostBuilder AddAppConfigurationFromAzureKeyVault(this IWebHostBuilder builder)
+        public static IWebHostBuilder ConfigureOnNonDevelopmentEnvironment(
```
jakubfijalkowski: This is too generic method (does not add much value). I would keep the old one intact (just use the now-extracted helper) and introduce new one that just adds the KeyVault. But don't be generic, let it operate on KV only.

jakubfijalkowski (changes requested):

Dragemil:
However integration tests refuse to use .NET 5, but that's for another day
```watch_test_1  | A compatible installed .NET Core SDK for global.json version [5.0.100-preview.7.20366.6] from [/app/code/global.json] was not found
watch_test_1  | Install the [5.0.100-preview.7.20366.6] .NET Core SDK or update [/app/code/global.json] with an installed .NET Core SDK:
watch_test_1  |   3.1.201 [/usr/share/dotnet/sdk]```

File: src/Core/LeanCode.Components.Startup/IWebHostBuilderExtensions.cs (hunk 1 above)
Comment (Dragemil) [reply src/Core/LeanCode.Components.Startup/IWebHostBuilderExtensions.cs]: But it neither takes away any value and using this feels fine.

File: src/Core/LeanCode.Components.Startup/IWebHostBuilderExtensions.cs [hunk 2]
```diff
@@ -43,6 +42,34 @@ public static IWebHostBuilder AddAppConfigurationFromAzureKeyVault(this IWebHost
...
         }
 
+        public static IWebHostBuilder AddAppConfigurationFromAzureKeyVaultOnNonDevelopmentEnvironment(
+            this IWebHostBuilder builder)
+        {
+            return builder.ConfigureAppConfiguration((context, builder) =>
+            {
+                if (!context.HostingEnvironment.IsDevelopment())
+                {
+                    var configuration = builder.Build();
```
jakubfijalkowski: DRY :) `AddAppConfigurationFromAzureKeyVault` should also throw.

jakubfijalkowski (changes requested):

jakubfijalkowski (approved):

PR #102, opened 2020-08-21
Feature/mass transit 7

marcin-dardzinski:
Bus configuration in MT 7 slightly changed. Instead of generic `AddBus` method now `UsingInMemory`/`UsingRabbitMQ` variants are preferred. As a result we have to expose the whole MT configurator method. This approach is more flexible, e.g. now we will have full control of MT registration.

### DIFF
--- a/src/Domain/LeanCode.DomainModels.MassTransitRelay/MassTransitRelayModule.cs
+++ b/src/Domain/LeanCode.DomainModels.MassTransitRelay/MassTransitRelayModule.cs
@@ -9,11 +9,12 @@ using LeanCode.DomainModels.MassTransitRelay.Outbox;
 using LeanCode.DomainModels.MassTransitRelay.Simple;
 using LeanCode.PeriodicService;
 using MassTransit;
+using MassTransit.AutofacIntegration;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace LeanCode.DomainModels.MassTransitRelay
 {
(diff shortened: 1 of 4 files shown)


src/Domain/LeanCode.DomainModels.MassTransitRelay/MassTransitRelayModule.cs
```diff
@@ -9,11 +9,12 @@
 using LeanCode.DomainModels.MassTransitRelay.Simple;
 using LeanCode.PeriodicService;
 using MassTransit;
+using MassTransit.AutofacIntegration;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace LeanCode.DomainModels.MassTransitRelay
 {
-    public delegate IBusControl BusFactory(IComponentContext context);
+    public delegate void BusFactory(IContainerBuilderBusConfigurator busConfig);
```
Saancreed: It doesn't look like a factory anymore :stuck_out_tongue: 

How about removing this delegate type and using some sort of `Action<IContainerBuilderBusConfigurator> configureBusAction` instead?

marcin-dardzinski: :stuck_out_tongue:

jakubfijalkowski (approved):

Saancreed (approved):

PR #105, opened 2020-08-27
Publish to FeedZ

### LINKED ISSUE #47: Publish to internal feed only

jakubfijalkowski:
We need to stop pushing every commit to NuGet. NuGet releases should be stable and our v5 currently is far from being one.

Closes #47

### DIFF
--- a/.github/workflows/core_build_publish.yml
+++ b/.github/workflows/core_build_publish.yml
@@ -93,14 +93,18 @@ jobs:
     runs-on: ubuntu-18.04
     name: Publish to NuGet
     needs: [prepare, build]
-    if: ${{ needs.prepare.outputs.publish_nuget == '1' }}
     steps:
       - name: Fetch build
         id: download
         uses: actions/download-artifact@v2
         with:
           name: ${{ needs.prepare.outputs.artifacts_name }}
+      - name: Push to Feedz
+        run: |
+          unzip "$ZIP"
+          find -name '*.nupkg' -exec dotnet nuget push -k "$FEEDZ_API_KEY" -s 'https://f.feedz.io/leancode/public/nuget/index.json' -n true '{}' ';'
       - name: Create release
+        if: ${{ needs.prepare.outputs.publish_nuget == '1' }}
         uses: actions/create-release@v1
         with:
           tag_name: ${{ format('v{0}', needs.prepare.outputs.version) }}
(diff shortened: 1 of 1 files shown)


lukaszgarstecki (approved):

marcin-dardzinski:
W8, do we? 
After I last changed it only manually triggered builds should be published to nuget.
Also, if prereleases are a thing, we can append alpha/beta/prerelease suffix and nuget will respect that. Since our new projects will use v5.0 anyway, it'd better to have them on nuget.

jakubfijalkowski:
> After I last changed it only manually triggered builds should be published to nuget.

Check how Emil uses it. He is on v5 already and pushes most of the changes there.

> Also, if prereleases are a thing, we can append alpha/beta/prerelease suffix and nuget will respect that.

That is not something we should do. Prereleases should have some sensible release cadence and at least be slightly stable. Pushing every commit on main branch there will just spam the NuGet page.

EDIT:
Also, this is mostly painless - most of our projects would use NuGet feed and only if someone wants to be cutting edge will use the Feedz feed (that is public).

marcin-dardzinski:
Well, I didn't mean publishing every merge to NuGet as preview, just the manually selected ones, but okay, no cost at having feedz too.

.github/workflows/core_build_publish.yml
```diff
@@ -91,26 +91,33 @@ jobs:
...
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - name: Push to NuGet
+        if: ${{ needs.prepare.outputs.publish_nuget == '1' }}
         run: |
           unzip "$ZIP"
           find -name '*.nupkg' -exec dotnet nuget push -k "$NUGET_API_KEY" -s 'https://api.nuget.org/v3/index.json' -n true '{}' ';'
         env:
           NUGET_API_KEY: ${{ secrets.NUGET_API_KEY }}
+          FEEDZ_API_KEY: ${{ secrets.FEEDZ_API_KEY }}
```
marcin-dardzinski: This `env` block is scoped to the Publish to NuGet step, so publishing to feedz will fail. I think it's possible to setup env on job level to avoid copying, or if not this would have to go to the correct step

marcin-dardzinski (changes requested):

Dragemil:
The v5.0 is still very unstable, but it is used by me, so this seems like a reasonable solution.

Dragemil (approved):

marcin-dardzinski (approved):

PR #108, opened 2020-09-16
Bump dotnet to RC1, upgrade packages and resolve NRT warnings

Dragemil:


src/Core/LeanCode.Components.Startup/SerilogExtensions.cs
```diff
@@ -32,7 +32,7 @@ internal static class SerilogExtensions
             return config.Enrich.WithProperty("AppName", appName);
         }
 
-        private static List<TType> SelectTypes<TType>(Assembly[] searchAssemblies)
+        private static List<TType?> SelectTypes<TType>(Assembly[] searchAssemblies)
```
jakubfijalkowski: Nope, that's not correct. This will never result in `null`.

src/Domain/LeanCode.CQRS.RemoteHttp.Client/HttpQueriesExecutor.cs
```diff
@@ -23,7 +23,7 @@ public HttpQueriesExecutor(HttpClient client)
             this.serializerOptions = serializerOptions;
         }
 
-        public virtual async Task<TResult> GetAsync<TResult>(IRemoteQuery<TResult> query)
+        public virtual async Task<TResult?> GetAsync<TResult>(IRemoteQuery<TResult> query)
```
jakubfijalkowski: That's not correct also. This should never return `null`. The response is malformed is null is returned here.

jakubfijalkowski (changes requested):

src/Domain/LeanCode.CQRS.RemoteHttp.Client/HttpQueriesExecutor.cs
Saancreed: I mean, technically it could, but only if `TResult` itself is a nullable type.

jakubfijalkowski: Ah, yeah, indeed. That's the only case.

jakubfijalkowski (approved):

Saancreed (approved):

PR #130, opened 2020-10-19
Fix `info`-level diagnostics

jakubfijalkowski:
Some are debatable (the defaults) but some are quite good I think (empty array, statics, etc.).


lukaszgarstecki (approved):
I like the idea of using Array.Empty

src/Domain/LeanCode.DomainModels.MassTransitRelay/AsyncEventsInterceptor.cs
```diff
@@ -1,10 +1,12 @@
 using System;
 using System.Collections.Concurrent;
+using System.Diagnostics.CodeAnalysis;
 using System.Threading;
 using LeanCode.DomainModels.Model;
 
 namespace LeanCode.DomainModels.MassTransitRelay
 {
+    [SuppressMessage("?", "CA1822", Justification = "Forcing these to be instance methods makes for better desing.")]
```
Saancreed: Desing?

test/Domain/LeanCode.DomainModels.MassTransitRelay.Tests/StoreAndPublishEventsTests.cs
```diff
@@ -89,9 +89,12 @@ private Task<List<RaisedEvent>> GetRaisedEvents()
                 .ToListAsync();
         }
 
-        public void Dispose() => dbContext.Dispose();
+        public void Dispose()
+        {
+            dbContext.Dispose();
+        }
```
Saancreed: This change seems to be quite unnecessary.

src/Core/LeanCode.Time/LeanCode.Time.csproj
```diff
@@ -10,6 +10,11 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <Compile Update="System/Strings.Designer.cs">
+      <DesignTime>True</DesignTime>
+      <AutoGen>True</AutoGen>
+      <DependentUpon>Strings.resx</DependentUpon>
+    </Compile>
```
Saancreed: Ah, so this is the magic invocation I was missing so the compiler doesn't complain about this file being included twice? Good to know.

LeanCode.CodeAnalysis.ruleset
```diff
@@ -9,6 +9,8 @@
         <!-- ConvertSwitchStatementToExpression -->
         <Rule Id="IDE1006" Action="None"/>
         <!-- NamingRule (interface with I) - we already have this rule in StyleCop -->
+        <Rule Id="CA1816" Action="None"/>
+        <!-- DisposeMethodsShouldCallSuppressFinalize - we don't use finalizers that much -->
```
Saancreed, line 13: Do we actually violate this rule so much?

Saancreed (approved):

src/Domain/LeanCode.DomainModels.MassTransitRelay/AsyncEventsInterceptor.cs
Saancreed: Oh, and it might be a good idea to actually set `category` to a correct value :upside_down_face:

test/Domain/LeanCode.DomainModels.MassTransitRelay.Tests/StoreAndPublishEventsTests.cs
jakubfijalkowski: Yep, I did add the `GC.SupressFinalize` and then removed it, forgetting about the layout change. Will change back.

LeanCode.CodeAnalysis.ruleset
jakubfijalkowski: Yeah, it fires up even if you don't have finalizer.

src/Core/LeanCode.Time/LeanCode.Time.csproj
jakubfijalkowski: Nope, that's not it unfortunately. I added it to test that and left it just in case.

LeanCode.CodeAnalysis.ruleset
Saancreed: Then maybe consider either suppressing it on each `Dispose` method we have or just call that `GC.SuppressFinalize(this)` to satisty it, even if it's a noop in most cases. You never know when the time for implementing a finalizer actually comes, and then such a rule might be useful.

src/Domain/LeanCode.DomainModels.MassTransitRelay/AsyncEventsInterceptor.cs
jakubfijalkowski: Yeah, `design`. `Category` is PITA :P

LeanCode.CodeAnalysis.ruleset
jakubfijalkowski: Up to the point when we really need the finalizer, this rule just clutters the code. I think we never used the finalizer in our code and I don't see it being used in the near future.

marcin-dardzinski (approved):

PR #132, opened 2020-10-22
Add ef date property builder

lukaszgarstecki:


jakubfijalkowski:
Adding converters for `Time` would be beneficial. :)

src/Domain/LeanCode.DomainModels.EF/DateConverter.cs
```diff
@@ -0,0 +1,13 @@
+using System;
+using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
+
+namespace LeanCode.DomainModels.EF
+{
+    public class DateConverter : ValueConverter<Date, DateTime>
+    {
+        public static readonly DateConverter Instance = new DateConverter { };
```
jakubfijalkowski: `new DateConverter();`

src/Domain/LeanCode.DomainModels.EF/DateConverter.cs
```diff
@@ -0,0 +1,13 @@
...
+        public DateConverter()
+            : base(d => d.ToDateTimeAtMidnight(), d => d.GetDate(), null)
```
jakubfijalkowski: Split these into three (well - four) lines, it will be more readable.

Also, you've missed a new line above.

jakubfijalkowski (changes requested):

src/Domain/LeanCode.DomainModels.EF/DateConverter.cs
Saancreed: I thought we skipped `()` for parametersless constructors?

jakubfijalkowski: IFF you set some properties using object initializer. If you have to choose between `new T {};` vs `new T();`, select the latter.

jakubfijalkowski (approved):

PR #133, opened 2020-10-23
Remove IUnitOfWork

Dragemil:
Since we tend to commit changes on the infrastructure level, the interface, being an abstraction for that, became redundant.


src/Domain/LeanCode.DomainModels.EF/EFRepository.cs
```diff
@@ -9,19 +9,20 @@
...
+    public abstract class EFRepository<TEntity, TIdentity, TContext> : IRepository<TEntity, TIdentity>
         where TEntity : class, IAggregateRootWithoutOptimisticConcurrency<TIdentity>
         where TIdentity : notnull
         where TContext : notnull, DbContext
-        where TUnitOfWork : notnull, IUnitOfWork
     {
+        private TContext DbContext { get; }
         protected DbSet<TEntity> DbSet { get; }
-        protected TUnitOfWork UnitOfWork { get; }
+        protected Task CommitAsync(CancellationToken cancellationToken = default) =>
```
jakubfijalkowski: Since this is EF-based implementation, let's stick to `SaveChangesAsync` naming.

src/Domain/LeanCode.DomainModels.EF/EFRepository.cs
```diff
@@ -9,19 +9,20 @@
...
 namespace LeanCode.DomainModels.EF
 {
-    public abstract class EFRepository<TEntity, TIdentity, TContext, TUnitOfWork> : IRepository<TEntity, TIdentity>
+    public abstract class EFRepository<TEntity, TIdentity, TContext> : IRepository<TEntity, TIdentity>
         where TEntity : class, IAggregateRootWithoutOptimisticConcurrency<TIdentity>
         where TIdentity : notnull
         where TContext : notnull, DbContext
-        where TUnitOfWork : notnull, IUnitOfWork
     {
+        private TContext DbContext { get; }
```
jakubfijalkowski: Just in case - let's make this `protected`.

jakubfijalkowski (changes requested):

jakubfijalkowski:
Aren't we using `IUnitOfWork` somewhere else in the CoreLib? Is this the only usage?

src/Domain/LeanCode.DomainModels.EF/EFRepository.cs
Dragemil: Should I remove this function if I'll make `DbContext` protected?

Dragemil:
Yes, this was the only usage, I've searched for that.

src/Domain/LeanCode.DomainModels.EF/EFRepository.cs
Dragemil: Or maybe make it public in order to have access to that?

jakubfijalkowski: Yeah, we can remove that if you find that redundant.

No, I think `public` will be too much. Let's make the final classes decide what they want to do with this.

src/Domain/LeanCode.DomainModels.EF/EFRepository.cs
```diff
@@ -9,19 +9,20 @@
...
+    public abstract class EFRepository<TEntity, TIdentity, TContext> : IRepository<TEntity, TIdentity>
         where TEntity : class, IAggregateRootWithoutOptimisticConcurrency<TIdentity>
         where TIdentity : notnull
         where TContext : notnull, DbContext
-        where TUnitOfWork : notnull, IUnitOfWork
     {
+        protected TContext DbContext { get; }
         protected DbSet<TEntity> DbSet { get; }
-        protected TUnitOfWork UnitOfWork { get; }
+        protected Task SaveChangesAsync(CancellationToken cancellationToken = default) =>
```
Saancreed: Empty line before this one please.
```suggestion

        protected Task SaveChangesAsync(CancellationToken cancellationToken = default) =>
```

Saancreed (commented):
Looks fine, just one minor request.

Also, having that `protected TContext DbContext { get; }` will be quite useful, which is nice.

src/Domain/LeanCode.DomainModels.EF/EFRepository.cs
jakubfijalkowski: Aaand let's move it after `public` methods.

Saancreed (approved):

jakubfijalkowski (approved):

PR #150, opened 2020-11-18
MassTransit inactivity, part 2

jakubfijalkowski:
Well... It's not _that_ simple.

`BusActivityMonitor` uses under the hood the `SemaphoreSlim` class (with single reentrance) which, IMO, is quite unfortunate. This has this nasty side effect that the code will lock forever (provided that there are no stray messages left):

```
await monitor.AwaitBusInactivity();
await monitor.AwaitBusInactivity();
```

Also, if you don't send any message to the bus and call the `AwaitBusInactivity`, the returned task will never end. There are also problems with thread-safety since the semaphore is released only once and therefore it will only unlock a single thread, but that is more of a "nice to know" than real problem for us.

In this PR, I introduce somewhat simplified version of the `IBusActivityMonitor` that is "resettable" and works more like a toggle than a semaphore (i.e. it behaves as `ManualResetEvent` and not `AutoResetEvent`). It counts the consumers and message receives since counting sends and publishes won't add much (in-memory bus is _fast_ in this regard). After there are no more outstanding consumers/receives, it triggers a timer that gives some time for the bus to stabilize.

Note that this won't exactly allow for consumer retries if they are not _instant_ - if the delay between restarts is longer than inactivity wait time, the monitor will be toggled in-between. I think this is a nice side effect, as we should not rely on restarts in integration tests, really. :)

### DIFF
--- a/src/Domain/LeanCode.DomainModels.MassTransitRelay/Testing/AsyncManualResetEvent.cs
+++ b/src/Domain/LeanCode.DomainModels.MassTransitRelay/Testing/AsyncManualResetEvent.cs
@@ -1,0 +1,93 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using GreenPipes.Internals.Extensions;
+
+namespace LeanCode.DomainModels.MassTransitRelay.Testing
+{
+    // Based on https://github.com/StephenCleary/AsyncEx/blob/master/src/Nito.AsyncEx.Coordination/AsyncManualResetEvent.cs
+    public sealed class AsyncManualResetEvent
+    {
+        private readonly object mutex;
+        private TaskCompletionSource<object?> tcs;
+
+        public AsyncManualResetEvent(bool set)
+        {
+            mutex = new object();
+            tcs = new TaskCompletionSource<object?>(TaskCreationOptions.RunContinuationsAsynchronously);
+
+            if (set)
+            {
+                tcs.TrySetResult(null);
+            }
+        }
+
+        public AsyncManualResetEvent()
+            : this(false)
+        { }
+
+        public bool IsSet
+        {
+            get
+            {
+                lock (mutex)
+                {
+                    return tcs.Task.IsCompleted;
+                }
+            }
+        }
+
+        public Task WaitAsync()
+        {
+            lock (mutex)
+            {
+                return tcs.Task;
+            }
+        }
+
+        public Task WaitAsync(CancellationToken cancellationToken)
+        {
+            var waitTask = WaitAsync();
+            if (waitTask.IsCompleted)
+            {
+                return waitTask;
+            }
+            else
+            {
+                return waitTask.OrCanceled(cancellationToken);
+            }
+        }
+
+        public async Task<bool> WaitAsync(TimeSpan timeout)
+        {
+            try
+            {
+                await WaitAsync().OrTimeout(timeout);
+                return true;
+            }
+            catch (TimeoutException)
+            {
+                return false;
+            }
+        }
+
+        public void Set()
+        {
+            lock (mutex)
+            {
+                tcs.TrySetResult(null);
+            }
+        }
+
+        public void Reset()
+        {
+            lock (mutex)
+            {
+                if (tcs.Task.IsCompleted)
+                {
+                    tcs = new TaskCompletionSource<object?>(TaskCreationOptions.RunContinuationsAsynchronously);
+                }
+            }
+        }
+    }
+}
--- a/src/Domain/LeanCode.DomainModels.MassTransitRelay/Testing/AsyncManualResetEvent.cs
+++ b/src/Domain/LeanCode.DomainModels.MassTransitRelay/Testing/AsyncManualResetEvent.cs
@@ -9,16 +9,16 @@ namespace LeanCode.DomainModels.MassTransitRelay.Testing
     public sealed class AsyncManualResetEvent
     {
         private readonly object mutex;
-        private TaskCompletionSource<object?> tcs;
+        private TaskCompletionSource<ValueTuple> tcs;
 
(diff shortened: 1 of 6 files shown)


lukaszgarstecki (approved):

src/Domain/LeanCode.DomainModels.MassTransitRelay/Testing/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,93 @@
...
+using System.Threading.Tasks;
+using GreenPipes.Internals.Extensions;
+
+namespace LeanCode.DomainModels.MassTransitRelay.Testing
+{
+    // Based on https://github.com/StephenCleary/AsyncEx/blob/master/src/Nito.AsyncEx.Coordination/AsyncManualResetEvent.cs
+    public sealed class AsyncManualResetEvent
+    {
+        private readonly object mutex;
+        private TaskCompletionSource<object?> tcs;
```
Saancreed: I'd prefer `TaskCompletionSource<ValueTuple>` over `TaskCompletionSource<object?>` if you don't use it to store an actual result, just signal completion of the task.

Dragemil (approved):

src/Domain/LeanCode.DomainModels.MassTransitRelay/Testing/ResettableBusActivityMonitor.cs
```diff
@@ -0,0 +1,148 @@
...
+        IReceiveObserver,
+        IConsumeObserver,
+        IDisposable
+    {
+        private readonly object mutex = new object();
+        private readonly AsyncManualResetEvent inactive = new AsyncManualResetEvent(true);
+        private readonly RollingTimer timer;
+
+        private int consumersInFlight;
+        private int receiversInFlight;
```
Saancreed: Since all writes to these variables are protected with a `lock` anyway, maybe you could mark them with `volatile` and get rid of usages of `Interlocked` and `Volatile` classes? Using both seems like an overkill, but I might be wrong here.

Saancreed (commented):
One request, one possible suggestion, looks good otherwise. Kind of disappointing we have to do it ourselves manually, but nice side effects are nice indeed.

src/Domain/LeanCode.DomainModels.MassTransitRelay/Testing/ResettableBusActivityMonitor.cs
jakubfijalkowski: Yep, you're right. I totally forgot about the keyword.

src/Domain/LeanCode.DomainModels.MassTransitRelay/Testing/AsyncManualResetEvent.cs
jakubfijalkowski: Using `object?` with `null` as a marker value is shorter and, at least now, more prevalent in .NET world. Also, it results in shorter code since `()` is not an empty tuple and you have to use `new ValueTuple()` (which, I think, will be boxed in the end, whereas `null` won't).

Saancreed: You could use `TrySetResult(default)`. And boxing shouldn't be an issue, unless `TaskCompletionSource` stores it internally as a reference anyway. Actually, I'm going to check this later, but for now `object?` is fine.

jakubfijalkowski: Okay, they [have one with `VoidResult`](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Threading.Channels/src/System/Threading/Channels/TaskCompletionSource.cs) in Channels, so I think even .NET team is going away from `object?` here.

jakubfijalkowski: And no, they don't store it directly, but only as part of `Task<T>`, so there won't be any additional allocation.

src/Domain/LeanCode.DomainModels.MassTransitRelay/Testing/AsyncManualResetEvent.cs
```diff
@@ -0,0 +1,93 @@
...
+            }
+        }
+
+        public void Reset()
+        {
+            lock (mutex)
+            {
+                if (tcs.Task.IsCompleted)
+                {
+                    tcs = new TaskCompletionSource<ValueTuple>(TaskCreationOptions.RunContinuationsAsynchronously);
```
Saancreed: One more thing, I just remembered; target typing should work in such cases: `tcs = new(TaskCreationOptions.RunContinuationsAsynchronously)`.

But I'm not sure if we want to use it yet.

Saancreed (approved):

src/Domain/LeanCode.DomainModels.MassTransitRelay/Testing/AsyncManualResetEvent.cs
jakubfijalkowski: Ha, right. Let's check how does that look like. :)

jakubfijalkowski: With a slight update to StyleCop - it works and looks quite nice. :)

Saancreed: Also it appears that this release of StyleCop is aware of positional records, so it won't be angry with `record A(T Thing);` anymore :tada:

I shall fix my `NoWarn`s soon.

PR #164, opened 2020-12-01
Support multiple Apple ClientIds

Saancreed:
Basically yesterday's change now applied to the other provider too.


lukaszgarstecki (approved):

jakubfijalkowski:
Why is it needed?

Saancreed:
Because, according to @mateuszwojtczak, Apple's docs explicitly mention that you might get either app ID _or_ service ID there. I was originally expecting `pl.leancode.project.tst.signin` there but I got `pl.leancode.project.tst`. Except swapping it is apparently not the correct solution because while signing in from iOS device results in the latter, signing in from web or Android could still give me the former.

Well, maybe I should rename the property to `Audiences` then…

jakubfijalkowski (approved):

Issue #56, opened 2020-06-25
Improve `IAsyncInitializable` and use it everywhere instead of `IHostedService` if the service depends on _anything_


jakubfijalkowski:
The problem is that `IHostedService`s are `Start`ed and `Stop`ped in _the same order_. Consider the case of integration tests:

Assumptions:
1. You create the database in `IAsyncInitializable` (registered as the initial one),
2. You use the database in the cleanup process (e.g. MassTransit),
3. You drop the database when the (1) initializable is stopped.

Then consider the following execution order:
1. You start the `WebApplicationFactory` in integration tests,
2. It starts the `AsyncInitializer` and creates the database,
3. You start MT cleanup services,
4. Tests finishes,
5. It then drops the database 'cause of assumption (3),
6. Cleanup service fails because the database no longer exists,
7. Process exits.

It is less evident in normal application runs (since the DB isn't temporary), but in this scenario it generates helluva of invalid logs.

jakubfijalkowski:
I will probably fix this very soon since it makes me cringe every time I run integration tests in one of our projects.

PR #168, opened 2020-12-07
Ordered hosted service

jakubfijalkowski:
This PR replaces the custom `IAsyncInitializable` with a slightly modified version of an `IHostedService` that supports number-ordering. This is mostly the same code but with slightly changed naming. I've also ported everything in the repo to use that.

I'm starting to think that using this kind of ordering is flawed. The more I think about it, the more I'm for using registration order. The only thing that makes me not want to go with plain `IHostedService` is that ASP.NET Core [does not reverse the services when stopping](https://github.com/dotnet/aspnetcore/blob/v5.0.0/src/Hosting/Hosting/src/Internal/HostedServiceExecutor.cs#L29-L65) and we *do* have cross-service dependencies. We could change that by splitting them in two (i.e. `StartAsync` implemented in one class, `StopAsync` in the other) and ensuring that they are placed correctly, but this will be terribly fragile. :(

Solves (partly) #56


lukaszgarstecki (approved):

Saancreed (approved):
While the code itself is fine, I don't really understand why continuing to use `IAsyncInitializable` would be a problem and why do we need this breaking change to fix it.

jakubfijalkowski:
I hated the naming. `Initializable` suggests that we're doing initialization only, but if you look at `PeriodicAction` or `MassTransitHostedService`, these are really background services, so the name doesn't match the purpose. Also, I want to somewhat re-use the `IHostedService` infrastructure if we ever (and I really hope so) wish to get rid of this.

Also, I'm not sure whether that is the correct approach. It seems to me it is better named now, but I'm not sure whether it should ever exist in the first place.

Saancreed:
Perhaps we could file an issue about the order of stopping hosted services to aspnetcore repo, maybe we will find some resolution there.

jakubfijalkowski:
Yeah, you're right, I will send one.

jakubfijalkowski:
Here you go: https://github.com/dotnet/aspnetcore/discussions/28540

jakubfijalkowski:
I'm merging this so that we get at least _some_ improvement here.

PR #170, opened 2020-12-11
Code coverage with codecov.io + unit test results as GH comment

jakubfijalkowski:


lukaszgarstecki (approved):

.github/workflows/core_build_publish.yml
```diff
@@ -69,13 +69,31 @@ jobs:
...
-        run: dotnet msbuild /t:RunTests /p:Configuration=Release
+        run: dotnet msbuild /t:RunTestsWithCoverage
         working-directory: test
+
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v1
+        if: always()
+
+      - name: Convert Test Results to jUnit
+        run: trx2junit --output TestResults $(find -name '*.trx')
```
Saancreed:
```suggestion
        run: find -name '*.trx' -exec trx2junit --output TestResults {} +
```
Otherwise path splitting can happen.

lukaszgarstecki (approved):

test/LeanCode.Tests.csproj
```diff
@@ -16,9 +18,19 @@
...
+
+  <Target Name="RunTestsWithCoverage" DependsOnTargets="Build">
     <exec
-      Command="dotnet vstest @(TestProjectsAssemblies -> '&quot;%(FullPath)&quot;', ' ') --logger:&quot;trx;LogFileName=$(LogFileName)&quot; --Blame"
-      WorkingDirectory="$(TargetDir)" />
+      ContinueOnError="ErrorAndContinue"
+      WorkingDirectory="$(TargetDir)"
+      Command="dotnet vstest \
+        '%(_ResolvedProjectReferencePaths.FullPath)' \
+        --logger:&quot;trx&quot; \
```
Saancreed: Are quotes actually needed around string literals? Assuming this will be executed by native OS shell, `bash` and similar shells will just consume them and do nothing, making them unnecessary if that's the case here.

Saancreed (changes requested):
One minor nitpick regarding path splitting, otherwise looks good.

.github/workflows/core_build_publish.yml
jakubfijalkowski: The TRX files don't have space in the name nor the path where everything lands. We're safe here.

test/LeanCode.Tests.csproj
jakubfijalkowski: That's not needed now, really - it's a leftover from when `LogFileName` was a thing.

.github/workflows/core_build_publish.yml
Saancreed: That's a bold assumption which I'd rather we didn't rely on.

Saancreed (approved):

PR #172, opened 2020-12-11
Update outdated packages & remove the hack for code actions

### LINKED ISSUE #142: Bump CodeAnalyzers to newest version
We're currently on `3.7.0` and we need to test whether `3.8.0` would work.

The main problem is with `System.Collections.Immutable` - Omnisharp, being Mono project, used to have problems with versions newer than `1.5.0`. This might be fixed already and we need to test that.

jakubfijalkowski:
We now don't need to override the `System.Collections.Immutable` version - it works as expected without the hack.

Closes #142

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
-    <EntityFrameworkCoreVersion>5.0.0</EntityFrameworkCoreVersion>
-    <AspNetCoreVersion>5.0.0</AspNetCoreVersion>
+    <EntityFrameworkCoreVersion>5.0.1</EntityFrameworkCoreVersion>
+    <AspNetCoreVersion>5.0.1</AspNetCoreVersion>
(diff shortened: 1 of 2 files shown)


Directory.Build.targets
```diff
@@ -84,7 +83,9 @@
     <PackageReference Update="Microsoft.Extensions.Logging.Debug" Version="$(ExtensionsVersion)" />
     <PackageReference Update="Microsoft.Extensions.Options" Version="$(ExtensionsVersion)" />
     <PackageReference Update="Microsoft.Extensions.Options.ConfigurationExtensions" Version="$(ExtensionsVersion)" />
-    <PackageReference Update="Microsoft.Extensions.WebEncoders" Version="$(ExtensionsVersion)" />
+    <!-- Warning: as of 2020.12.11, the rest of `Extension` packages is at version 5.0.0 and these two are at 5.0.1 :) -->
```
Saancreed, line 86: :confused:

Saancreed (approved):

PR #173, opened 2020-12-11
Use multiple Sqlite connections in MassTransitRelay tests

### LINKED ISSUE #139: Flaky/deadlocking tests
We have one flaky test that randomly fails in the MassTransit suite: https://github.com/leancodepl/corelibrary/pull/138/checks?check_run_id=1340421337#step:6:101 .

Also, sometimes the tests deadlock for unknown reasons (I suppose something changed in RC2), eg. this: https://github.com/leancodepl/corelibrary/actions/runs/314779817 or https://github.com/leancodepl/corelibrary/actions/runs/315471310

jakubfijalkowski:
This fixes the flakyness in MassTransit relay tests.

Previously, it used a single connection that was accessed from multiple threads (the tests and MT), making it race for the transaction (during `SaveChangesAsync`). Now we use different connections to the same in-mem database, making the tests thread-safe.

Closes #139


marcin-dardzinski:
Hmm, I swear I read somewhere that the sqlite connection does support multithread access, but I might missed some feature switch.

marcin-dardzinski (approved):

lukaszgarstecki (approved):

PR #175, opened 2020-12-14
Use generic host and get rid of `OrderedHostedService`

jakubfijalkowski:
Sooooo...

Long story short - I looked in the wrong place. It seems that we were still using the deprecated WebHost. This was the culprit - it doesn't reverse the `IHostedService`. The [correct generic Host](https://github.com/dotnet/runtime/blob/master/src/libraries/Microsoft.Extensions.Hosting/src/Internal/Host.cs#L94-L104) does exactly what we need.

When we use generic host, we don't really need `IOrderedHostedService`s - using `IHostedService` directly and ensuring that these are registered in the correct order will make our life easier (not custom!).

:)


Saancreed (approved):
Maybe we should have bumped the version to 5.1 already :)

jakubfijalkowski:
Yeah, the versioning... doesn't really work. :) 

Although we never said 5.0 is stable :P

jakubfijalkowski:
I've fixed the tests & moved the helper tests from "integration" to "unit" test runner, so that at least we can be semi-sure that changes to it work.

docs/general/02_building-and-testing.md
```diff
@@ -18,12 +18,11 @@ The framework can be unit-tested by `cd`ing into `test` folder and calling
 dotnet msbuild /t:RunTests
 ```
 
-Moreover, there are some integration-style tests that require external services. They can be tested with `docker` and `docker-compose` tool. There are two integration-test suites:
+Moreover, there are some integration-style tests that require external services. They can be tested with `docker` and `docker-compose` tool. There is one integration-test suite currently:
 
- 1. `test/IntegrationTests` - _real_ integration tests,
- 2. `test/Testing/LeanCode.IntegrationTestHelpers.Tests` - test for the integration test helpers.
+ 1. `test/LeanCode.IntegrationTests` - _real_ integration tests,
```
Saancreed: This clarification seems unnecessary now that there is only one test suite.

Also, what is `test/Testing/…` supposed to be now that it's not mentioned here anymore?

jakubfijalkowski: `test/Testing` is no longer integration - it is run normally, along with other tests.

Saancreed (approved):

PR #191, opened 2021-01-11
Retrieve tokens for multiple users and send single message to multiple users

### LINKED ISSUE #92: Send FCM message to multiple users, but not all
Something like:
`public Task SendToMultipleUsersAsync(IEnumerable<Guid> userIds, MulticastMessage message, CancellationToken cancellationToken = default(CancellationToken))`

### LINKED ISSUE #189: Fetch FCM tokens for multiple users in a single query
In `IPushNotificationTokenStore`

```csharp
public Task<Dictionary<string, List<string>>> GetTokensAsync(
    IEnumerable<Guid> userIds,
    CancellationToken cancellationToken = default(CancellationToken))
```

jakubfijalkowski:
The `GetTokensAsync(List<Guid>)` limits the number of users to max. 100 so that we don't generate terribly inefficient SQL statement with long `IN` clause.

Closes #189
Closes #92

### DIFF
--- a/src/Infrastructure/LeanCode.Firebase.FCM/EntityFramework/PushNotificationTokenStore.cs
+++ b/src/Infrastructure/LeanCode.Firebase.FCM/EntityFramework/PushNotificationTokenStore.cs
@@ -58,7 +58,7 @@ namespace LeanCode.Firebase.FCM.EntityFramework
             {
                 await dbContext.ExecuteAsync(
                 $@"
-                    BEGIN TRAN;
+                    BEGIN TRANSACTION;
 
                     -- Remove token from (possibly another) user
                     DELETE FROM {GetTokensTableName()} WHERE [Token] = @token;
@@ -67,7 +67,7 @@ namespace LeanCode.Firebase.FCM.EntityFramework
                     INSERT INTO {GetTokensTableName()} ([Id], [UserId], [Token], [DateCreated])
                     VALUES (@newId, @userId, @token, @now);
 
-                    COMMIT TRAN;
+                    COMMIT TRANSACTION;
                 ", new { newId = Identity.NewId(), userId, token, now = TimeProvider.Now },
                 cancellationToken: cancellationToken);
                 logger.Information("Added push notification token for user {UserId} from the store", userId);
--- a/src/Infrastructure/LeanCode.Firebase.FCM/EntityFramework/PushNotificationTokenStore.cs
+++ b/src/Infrastructure/LeanCode.Firebase.FCM/EntityFramework/PushNotificationTokenStore.cs
@@ -14,8 +14,7 @@ namespace LeanCode.Firebase.FCM.EntityFramework
     public sealed class PushNotificationTokenStore<TDbContext> : IPushNotificationTokenStore
         where TDbContext : DbContext
     {
-        private const int MaxBatchSize = 100;
-
+        private const int MaxTokenBatchSize = IPushNotificationTokenStore.MaxTokenBatchSize;
         private readonly Serilog.ILogger logger = Serilog.Log.ForContext<PushNotificationTokenStore<TDbContext>>();
 
         private readonly TDbContext dbContext;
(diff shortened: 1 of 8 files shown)


src/Infrastructure/LeanCode.Firebase.FCM/EntityFramework/PushNotificationTokenStore.cs
```diff
@@ -14,6 +14,8 @@ namespace LeanCode.Firebase.FCM.EntityFramework
     public sealed class PushNotificationTokenStore<TDbContext> : IPushNotificationTokenStore
         where TDbContext : DbContext
     {
+        private const int MaxBatchSize = 100;
```
Saancreed: Might be a good idea to make it `public` (and rename to `GetTokensMaxBatchSize` for clarity), because it's kind of part of the public contract right now.

src/Infrastructure/LeanCode.Firebase.FCM/EntityFramework/PushNotificationTokenStore.cs
```diff
@@ -32,13 +34,31 @@ public async Task<List<string>> GetTokensAsync(Guid userId, CancellationToken ca
             return res.AsList();
         }
 
+        public async Task<Dictionary<Guid, List<string>>> GetTokensAsync(List<Guid> userIds, CancellationToken cancellationToken = default)
```
Saancreed: `List<Guid> userIds` → `IReadOnlySet<Guid> userIds` or `IReadOnlyCollection<Guid> userIds` maybe?

src/Infrastructure/LeanCode.Firebase.FCM/FCMClient.cs
```diff
@@ -60,6 +62,25 @@ public async Task SendToUserAsync(Guid userId, MulticastMessage message, bool dr
...
         }
 
+        public async Task SendToUsersAsync(List<Guid> userIds, MulticastMessage message, bool dryRun, CancellationToken cancellationToken = default)
+        {
+            var tokens = await tokenStore.GetTokensAsync(userIds, cancellationToken);
+            message.Tokens = tokens.SelectMany(t => t.Value).ToList();
+
+            if (message.Tokens.Count == 0)
+            {
+                logger.Information("Cannot send push to user {UserId} - no tokens", userIds);
```
Saancreed: `userIds` is a collection while the template implies that only one user per message will be skipped.

src/Infrastructure/LeanCode.Firebase.FCM/EntityFramework/PushNotificationTokenStore.cs
```diff
@@ -32,13 +34,31 @@ public async Task<List<string>> GetTokensAsync(Guid userId, CancellationToken ca
...
+
+            var res = await dbContext.QueryAsync<UserToken>(
+                $"SELECT [UserId], [Token] FROM {GetTokensTableName()} WHERE [UserId] IN @userIds",
+                new { userIds },
+                cancellationToken: cancellationToken);
+            return res
+                .GroupBy(g => g.UserId)
+                    .ToDictionary(
+                        t => t.Key,
+                        t => t.Select(e => e.Token).ToList());
```
Saancreed: Unnecessary extra level of indentation.

Saancreed (commented):
One strange logger call that will probably need to be changes and some minor suggestions, looks good otherwise.

src/Infrastructure/LeanCode.Firebase.FCM/EntityFramework/PushNotificationTokenStore.cs
jakubfijalkowski: I disagree - the exact number is not part of the real _contract_. Exposing it just makes it the liability and it won't be used much - most of the time the code would depend on `IPushNotificationTokenStore` than this.

But... it might be a good idea to put in the interface and make it contractual there, so no matter the implementation, we can't expect it to accept more params.

src/Infrastructure/LeanCode.Firebase.FCM/EntityFramework/PushNotificationTokenStore.cs
jakubfijalkowski: `IReadOnlySet` is a better option, thanks!

Saancreed (approved):

marcin-dardzinski (approved):

PR #205, opened 2021-01-26
Facebook external login

jakubfijalkowski:
Adds Facebook to the `LeanCode.ExternalIdentityProviders`. Plus long-needed tests.

There is one place though that I am not sure about. We want to get rid of `LeanCode.Facebook` in the next version and integrate it directly into the provider (as that will be it's only use-case). To make the transition a little bit easier, I register `FacebookModule` along `ExternalIdentityProvidersModule` - this is the part I am not sure about. It will require no changes after the next version but it deviates from what we have up to this point. What do you think?


src/Infrastructure/LeanCode.ExternalIdentityProviders/Facebook/Constants.cs
```diff
@@ -0,0 +1,7 @@
+namespace LeanCode.ExternalIdentityProviders.Facebook
+{
+    public static class Constants
+    {
+        public const string GrantType = "google";
```
Saancreed: That doesn't feel correct.

lukaszgarstecki: namespace Facebook and grant type google :thinking:

src/Infrastructure/LeanCode.ExternalIdentityProviders/ExternalIdentityProvidersModule.cs
```diff
@@ -2,28 +2,77 @@
...
 namespace LeanCode.ExternalIdentityProviders
 {
+    [Flags]
+    public enum Providers
+    {
+        Facebook = 0b001,
+        Apple = 0b010,
+        Google = 0b100,
+
+        None = 0b000,
```
Saancreed, line 19: Attempting to register this module with `Providers.None` doesn't make much sense so I'd expect it to throw.

Saancreed (commented):
I think it's a good time to start v5.1 (removing `LeanCode.Facebook` and keep it unstable for a while) but we could stay on v5.0 and mark everything from the old module as `[Obsolete]`.

lukaszgarstecki (changes requested):

Saancreed:
Also:

> To make the transition a little bit easier, I register `FacebookModule` along `ExternalIdentityProvidersModule` - this is the part I am not sure about.

I feel like there is no point, we should avoid it if possible because it will go away soon. Anyone that uses that module now should migrate asap and if you are about to add Facebook integration just now, go with `ExternalIdentityProvidersModule` only. We will have to deal with two implementations for some time but this way the transition, while requiring some manual intervention, will be cleaner imho.

jakubfijalkowski:
Done, targeting v5.1 now, with `LeanCode.Facebook` removed completely.

lukaszgarstecki (approved):

src/Infrastructure/LeanCode.ExternalIdentityProviders/ExternalIdentityProvidersModule.cs
```diff
@@ -2,28 +2,76 @@
...
+    }
+
     public class ExternalIdentityProvidersModule<TUser> : AppModule
         where TUser : IdentityUser<Guid>
     {
+        private readonly Providers configuration = Providers.All;
+
+        public ExternalIdentityProvidersModule(Providers configuration)
+        {
+            if (configuration == Providers.None)
```
Saancreed:
```suggestion
            if ((configuration & Providers.All) == Providers.None)
```
This way we are also guarded against invalid enum values.

src/Infrastructure/LeanCode.ExternalIdentityProviders/ExternalIdentityProvidersModule.cs
```diff
@@ -2,28 +2,76 @@
...
+        Google = 0b100,
+
+        None = 0b000,
+        All = 0b111,
+    }
+
     public class ExternalIdentityProvidersModule<TUser> : AppModule
         where TUser : IdentityUser<Guid>
     {
+        private readonly Providers configuration = Providers.All;
```
Saancreed: Default value is unnecessary now that there is no parameterless constructor.
```suggestion
        private readonly Providers configuration;
```

Saancreed (approved):
LGTM. Feel free to either implement or ignore my last two suggestions.

PR #208, opened 2021-02-04
Add ADR about value objects

lukaszgarstecki:
Result of our internal discussion


docs/adrs/adr1_value_object.md
```diff
@@ -0,0 +1,32 @@
...
+## Consequences
+
+Abstract class was removed already. We need to provide implementation of that abstract record.
+
+## Side notes
+
+### How to use records as value objects
+
+* All properties must have `private init` setters.
+* Always remember that C# uses default value comparer (in case of arrays and collections this is reference quality).
```
Albert221: Did you want to use _comparator_ instead of the _comparer_?

Albert221: What do you mean by _reference quality_?

lukaszgarstecki: Tnat was supposed to be `equality`

docs/adrs/adr1_value_object.md
```diff
@@ -0,0 +1,32 @@
+# How to implement value objects in our domains
+
+While modelling domains we use concept of value object. The issue is how to achieve model of value object described in DDD.
```
jakubfijalkowski:
```suggestion
While modelling a domain we use the concept of value objects. The issue is how to represent value objects in code, as described in DDD.
```

docs/adrs/adr1_value_object.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Status
+
+To be implemented (04.02.21)
+
+## Context
+
+We see two options: Base abstract class and record.
```
jakubfijalkowski:
```suggestion
We see two options: base abstract class and record.
```

docs/adrs/adr1_value_object.md
```diff
@@ -0,0 +1,32 @@
...
+
+We used base class before records were introduced. The issue with base class is that it forced developer to implement `GetAttributesToIncludeInEqualityCheck` method in every implemented ValueObject and always contained all the properties that were defined.
```
jakubfijalkowski:
```suggestion
We used base class before records were introduced. The issue with the base class is that it forces developers to implement `GetAttributesToIncludeInEqualityCheck` method in every implemented `ValueObject` and always contained all the properties that were defined.
```

Also, what do you mean by "and always contained all the properties that were defined."?

docs/adrs/adr1_value_object.md
```diff
@@ -0,0 +1,32 @@
...
+
+When records came with C# 9.0 we decided to check how they suit our needs in regards to modeling value objects. In general, there were to differences in regards to previous option was that there were no abstract method that we need to implement (structural equality is more or less out of the box with records) and there were no marking interface or base class that clearly states that we are dealing with value object.
```
jakubfijalkowski:
```suggestion
When records came in C# 9.0 we decided to check how they suit our needs in regards to modeling value objects. In general, there were two differences compared to previous option: there were no abstract method that we need to implement (structural equality is more or less out of the box with records) and there were no marker interfaces or base classes that clearly state that we are dealing with a value object.
```

docs/adrs/adr1_value_object.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Decision
+
+We decided to use records that will inherit from abstract record `ValueObject`. This allows us to use all the benefits of records and do not loose the benefits of clearly stating what the record represents.
```
jakubfijalkowski:
```suggestion
We decided to use records that will inherit from a abstract record `ValueObject`. This allows us to use all the benefits of records and do not lose the benefits of clearly stating what the record represents.
```

docs/adrs/adr1_value_object.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Consequences
+
+Abstract class was removed already. We need to provide implementation of that abstract record.
```
jakubfijalkowski: [That's incorrect](https://github.com/leancodepl/corelibrary/blob/v5.0/src/Domain/LeanCode.DomainModels/Model/ValueObject.cs)

docs/adrs/adr1_value_object.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Side notes
+
+### How to use records as value objects
+
+* All properties must have `private init` setters.
+* Always remember that C# uses default value comparer (in case of arrays and collections this is reference quality).
+* We can use `with` only in private context. If anywhere in your codebase you can do `vo with { A = "B" }` you violated first point.
+* Use abstract base record (when it will be implemented).
```
jakubfijalkowski:
```suggestion
* Use the abstract base record (when it will be implemented).
```

jakubfijalkowski (changes requested):
Great job! I left you a couple of small suggestions. Additionally, may I ask you to add a note in the "normal" documentation? That is - start "domain" documentation and link that here?

docs/adrs/adr1_value_object.md
lukaszgarstecki: :open_mouth: will change to unused

docs/adrs/adr1_value_object.md
lukaszgarstecki: I was thinking about full property definition not like the short one like in records, but I guess it is not that crucial here, so I will remove this bit.

jakubfijalkowski (approved):

PR #216, opened 2021-02-19
Don't require FB `AppSecret`

jakubfijalkowski:
That was a mistake on my side - requiring `AppSecret` is painful on dev and in integration tests. :upside_down_face:

### DIFF
--- a/src/Infrastructure/LeanCode.ExternalIdentityProviders/Facebook/FacebookClient.cs
+++ b/src/Infrastructure/LeanCode.ExternalIdentityProviders/Facebook/FacebookClient.cs
@@ -19,7 +19,7 @@ namespace LeanCode.ExternalIdentityProviders.Facebook
         private readonly Serilog.ILogger logger = Serilog.Log.ForContext<FacebookClient>();
 
         private readonly HttpClient client;
-        private readonly HMACSHA256 hmac;
+        private readonly HMACSHA256? hmac;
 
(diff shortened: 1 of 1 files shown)


marcin-dardzinski (approved):

src/Infrastructure/LeanCode.ExternalIdentityProviders/Facebook/FacebookClient.cs
```diff
@@ -120,12 +124,18 @@ private FacebookUser ConvertResponse(JsonElement result)
 
         private string GenerateProof(string accessToken)
         {
-            logger.Debug("Signing Facebook request enabled - signing request");
-
-            var bytes = ParseKey(accessToken);
-            var hash = hmac.ComputeHash(bytes);
+            if (hmac is null)
+            {
+                logger.Error("Facebook's AppSecret is not configured, login with Facebook will fail");
```
Saancreed, line 129: I'm not sure if the best place for this message is here or in the constructor, but both seem fine to me.

Saancreed (approved):

src/Infrastructure/LeanCode.ExternalIdentityProviders/Facebook/FacebookClient.cs
jakubfijalkowski: I think here it will be easier to not miss the message.

PR #217, opened 2021-02-19
Log levels

jakubfijalkowski:
I lowered it based on our current projects (on v5 & v5.1). Probably missed something, so if you have something in mind - we can fix that in one go.

### DIFF
--- a/src/Domain/LeanCode.DomainModels.MassTransitRelay/Outbox/PeriodicEventsPublisher.cs
+++ b/src/Domain/LeanCode.DomainModels.MassTransitRelay/Outbox/PeriodicEventsPublisher.cs
@@ -40,7 +40,7 @@ namespace LeanCode.DomainModels.MassTransitRelay
 
             var events = await FetchUnpublishedEventsAsync();
 
-            logger.Information("There are {EventsCount} to be published", events.Count);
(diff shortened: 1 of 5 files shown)


Dragemil (approved):

marcin-dardzinski:
There was some log with missing argument, I think, but I don't remember it now :disappointed:

marcin-dardzinski (approved):

src/Domain/LeanCode.DomainModels.MassTransitRelay/Outbox/PeriodicEventsPublisher.cs
```diff
@@ -40,7 +40,7 @@ public async Task ExecuteAsync(CancellationToken stoppingToken)
 
             var events = await FetchUnpublishedEventsAsync();
 
-            logger.Information("There are {EventsCount} to be published", events.Count);
+            logger.Debug("There are {EventsCount} to be published", events.Count);
```
marcin-dardzinski: Yeah, that was the log that I missed - There are X **events** to be published.

jakubfijalkowski: Fixed

PR #227, opened 2021-04-01
Split build & unit test publish steps so that Dependabot PRs work

jakubfijalkowski:
See: https://github.com/EnricoMi/publish-unit-test-result-action/issues/95

### DIFF
--- a/.github/workflows/publish_unit_tests.yml
+++ b/.github/workflows/publish_unit_tests.yml
@@ -1,0 +1,62 @@
+name: Unit Test Results
+
+on:
+  workflow_run:
+    workflows: ["core_build_publish"]
+    types:
+      - completed
+
+jobs:
+  unit-test-results:
+    name: Unit Test Results
+    runs-on: ubuntu-latest
+    if: >
+      github.event.workflow_run.conclusion != 'skipped' && (
+         github.event.sender.login == 'dependabot[bot]' ||
+         github.event.workflow_run.head_repository.full_name != github.repository
+      )
+
+    steps:
+      - name: Download Artifacts
+        uses: actions/github-script@v3.1.0
+        with:
+          script: |
+            var fs = require('fs');
+            var path = require('path');
+            var artifacts_path = path.join('${{github.workspace}}', 'artifacts')
+            fs.mkdirSync(artifacts_path, { recursive: true })
+
+            var artifacts = await github.actions.listWorkflowRunArtifacts({
+               owner: context.repo.owner,
+               repo: context.repo.repo,
+               run_id: ${{ github.event.workflow_run.id }},
+            });
+
+            for (const artifact of artifacts.data.artifacts) {
+               var download = await github.actions.downloadArtifact({
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  artifact_id: artifact.id,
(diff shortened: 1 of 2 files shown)


.github/workflows/publish_unit_tests.yml
```diff
@@ -0,0 +1,62 @@
...
+               fs.writeFileSync(artifact_path, Buffer.from(download.data));
+               console.log(`Downloaded ${artifact_path}`);
+            }
+      - name: Extract Artifacts
+        run: |
+          for file in artifacts/*.zip
+          do
+            if [ -f "$file" ]
+            then
+              dir="${file/%.zip/}"
```
Saancreed, line 50: Assuming we're talking `bash` here, `${file%.zip}` already removes matching suffix pattern while being more obvious to the reader.

(And I'd prefer `[[` over `[` but I can live with the latter.)

.github/workflows/publish_unit_tests.yml
```diff
@@ -0,0 +1,62 @@
...
+
+    steps:
+      - name: Download Artifacts
+        uses: actions/github-script@v3.1.0
+        with:
+          script: |
+            var fs = require('fs');
+            var path = require('path');
+            var artifacts_path = path.join('${{github.workspace}}', 'artifacts')
+            fs.mkdirSync(artifacts_path, { recursive: true })
```
Saancreed: Inconsistent usage of semicolons (also in line 42).

.github/workflows/publish_unit_tests.yml
```diff
@@ -0,0 +1,62 @@
...
+      )
+
+    steps:
+      - name: Download Artifacts
+        uses: actions/github-script@v3.1.0
+        with:
+          script: |
+            var fs = require('fs');
+            var path = require('path');
+            var artifacts_path = path.join('${{github.workspace}}', 'artifacts')
```
Saancreed: `var` → `const`?

jakubfijalkowski: Since there is no frontend here, I hereby reject this request - this is nowhere being `const`ant (for the real meaning of constant). :P

.github/workflows/publish_unit_tests.yml
jakubfijalkowski: You mean like changing `${file/%.zip/}` to `${file%.zip}`? I would say that the latter is much less obvious - you "match", but "replace" is not evident IMO.

.github/workflows/publish_unit_tests.yml
Saancreed: So, `var` →  `let` maybe?

jakubfijalkowski: Doesn't change anything here IMO.

.github/workflows/publish_unit_tests.yml
Saancreed: Is it? I feel like it would be preferring `Regex.Replace(file, @"\.zip^", string.Empty)` to `file.TrimEnd(".zip")` (assuming C# had such a method), which just doesn't feel like the right tool for the job. Simple string chopping seems easier to comprehend for me.

Saancreed (approved):

PR #251, opened 2021-04-13
Fix sql syntax errors in published events cleaner

marcin-dardzinski:
I remember this was fixed before, must have been reintroduced when merging open telemetry :disappointed:


lukaszgarstecki (approved):

jakubfijalkowski (changes requested):
Tests are needed.

marcin-dardzinski:
Yeah, the problem is that we need sql server set up for this, you won't do this with sqlite...

jakubfijalkowski:
Why not? If you change the SQL to something like
```sql
DELETE FROM {tableName} WHERE
[DateOcurred] < @time AND [WasPublished] = 1
```
it will be valid in SQLite. :P

jakubfijalkowski (approved):

PR #279, opened 2021-05-31
.NET 6.0

Saancreed:
Not much to be done here really, just some framework and package bumps plus fixes for new warnings. `System.Time` a.k.a. `LeanCode.Time` has been removed because we now have `System.{Date,Time}Only` in `mscorelib`. The only issue was missing `ContentRoot` for entry point classes in test projects themselves, which doesn't seem to be very well supported scenario by .NET 6.0 version of `Microsoft.AspNetCore.Mvc.Testing`.

TODO:
* ~Retarget to `v6.0` branch~
* ~Remove `IdentityProvider`~
* ~Use new Azure libraries~
* Use new config variable naming scheme
* Update changelog
* Perhaps apply all the remaining IDE suggestions
* Perhaps find a better solution for missing `ContentRoot` issue


lukaszgarstecki (approved):

test/LeanCode.Tests.csproj
```diff
@@ -1,4 +1,26 @@
 <Project Sdk="Microsoft.NET.Sdk">
+
```
marcin-dardzinski, line 2: What is the reason for this change?

Saancreed: [Since .NET 6.0 preview1](https://github.com/dotnet/aspnetcore/commit/f7e1712db88db4c01fcc49e5995d9528dbf608df), `Microsoft.AspNetCore.Mvc.Testing` reads content roots from a JSON file first before consulting assembly attributes or solution directory. However, when that JSON is written, only projects referenced by the test projects are considered _without the test project itself_. So, because our `TEntryPoint` (Startup) class is in the test project itself, `WebApplicationFactory` [cannot find](https://github.com/dotnet/aspnetcore/blob/v6.0.0-preview.4.21253.5/src/Mvc/Mvc.Testing/src/WebApplicationFactory.cs#L196) its related `ContentRoot` and causes tests in `LeanCode.IntegrationTestHelpers.Tests` to fail. I work around that by ensuring it never tries to read that path from file, which requires me to set it via environment variable beforehand.

Directory.Build.targets
```diff
@@ -84,9 +84,9 @@
     <PackageReference Update="Microsoft.Extensions.Logging.Debug" Version="$(ExtensionsVersion)" />
     <PackageReference Update="Microsoft.Extensions.Options" Version="$(ExtensionsVersion)" />
     <PackageReference Update="Microsoft.Extensions.Options.ConfigurationExtensions" Version="$(ExtensionsVersion)" />
-    <!-- Warning: as of 2021.01.19, the rest of `Extension` packages is at version 5.0.0 and these two are at 5.0.1 and 5.0.2 respectively :) -->
-    <PackageReference Update="Microsoft.Extensions.DependencyInjection" Version="5.0.1" />
```
jakubfijalkowski, line 88: This is now an irrelevant comment :P

src/Domain/LeanCode.DomainModels.MassTransitRelay/DbContextExtensions.cs
```diff
@@ -8,14 +8,17 @@ public static class DbContextExtensions
         public static string GetFullTableName(this DbContext dbContext, Type entity)
         {
             var type = dbContext.Model.FindEntityType(entity);
-            if (string.IsNullOrEmpty(type.GetSchema()))
```
jakubfijalkowski, line 11: That's a totally unnecessary change. Please revert.

src/Tools/LeanCode.ContractsGenerator/CodeGenerator.cs
```diff
@@ -183,7 +183,7 @@ private IEnumerable<InterfaceStatement> GenerateClassesAndInterfaces(SemanticMod
 
             var classes = classesDeclarations
                 .Select(c => model.GetDeclaredSymbol(c))
-                .Union(structsDeclarations.Select(s => model.GetDeclaredSymbol(s)))
```
jakubfijalkowski, line 186: Why? This (as of .NET 5) changes the semantics of the equality (`SymbolEqualityComparer.Default` is not a default equality comparer).

src/Infrastructure/LeanCode.Firebase.FCM/EntityFramework/PushNotificationTokenStore.cs
```diff
@@ -129,14 +129,17 @@ public async Task RemoveTokensAsync(IEnumerable<string> tokens, CancellationToke
         private string GetTokensTableName()
         {
             var entity = dbContext.Model.FindEntityType(typeof(PushNotificationTokenEntity));
-            if (string.IsNullOrEmpty(entity.GetSchema()))
```
jakubfijalkowski, line 132: ditto

jakubfijalkowski (changes requested):

src/Domain/LeanCode.DomainModels.MassTransitRelay/DbContextExtensions.cs
Saancreed: It is not, otherwise a dereference of possibly–`null` variable `type` is reported as warning.

src/Tools/LeanCode.ContractsGenerator/CodeGenerator.cs
Saancreed: I thought we are looking at declarations here, which should not declare nullable symbols, but I will change this to `SymbolEqualityComparer.IncludeNullability` if you believe it to be more correct in this context.

Directory.Build.targets
Saancreed: Which is why it's gone now?

src/Tools/LeanCode.ContractsGenerator/CodeGenerator.cs
Saancreed: The only [documentation I've found on the subject](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.symbolequalitycomparer.default?view=roslyn-dotnet-3.9.0) claims that `SymbolEqualityComparer.Default` is the one equivalent to calling `IEquatable<T>.Equals(T)` and not the other one.

Directory.Build.targets
jakubfijalkowski: Yeah, but the new one is also irrelevant :p

Saancreed: I'd say that if attempting to use a `Microsoft.Extensions.*` package with version `$(ExtensionsVersion)` results in an error, the comment is just as justified as the previous one, otherwise this small discrepancy would be too easy to overlook.

src/Tools/LeanCode.ContractsGenerator/CodeGenerator.cs
jakubfijalkowski: Ugh, so they changed it once again. There was a time where official analyzers suggested using it explicitly. Thanks for the explanation.

Nevertheless, it doesn't really matter here - this generator will be hopefully removed soon.

jakubfijalkowski: Ugh, so they changed it once again. There was a time where official analyzers suggested using it explicitly. Thanks for the explanation.

Nevertheless, it doesn't really matter here - this generator will be hopefully removed soon.

Directory.Build.targets
jakubfijalkowski: Ah, dammit, it's different variable. Missed that.

I would say that we shouldn't mix the variables. It makes the intention really unclear and the benefit is rather small, especially that it probably will change differently really soon.

src/Domain/LeanCode.DomainModels.MassTransitRelay/DbContextExtensions.cs
jakubfijalkowski: Yeah, then add the `type is null` guard but don't change the rest - it's unnecessary in this PR.

Nevertheless, let's leave it changed for now.

jakubfijalkowski (approved):

Directory.Build.targets
Saancreed: I'll try to move it around so it doesn't blend with other `Microsoft.Extension.*` packages, but I'd let the comment stay anyway, at least until .NET 6 becomes GA and these `-preview…` suffixes go away.

src/Domain/LeanCode.DomainModels.MassTransitRelay/DbContextExtensions.cs
Saancreed: Fair enough, I just couldn't help myself when I saw an opportunity for a refactor that would let us avoid unnecessary method call :see_no_evil:

But the ternary operator… yeah, I shouldn't have done that, not in this PR at least.

src/Tools/LeanCode.ContractsGenerator/CodeGenerator.cs
Saancreed: > There was a time where official analyzers suggested using it explicitly.

The analyzer suggesting using a comparator explicitly was exactly the reason why I've changed that, it just doesn't force you to use any particular one, only that you shouldn't rely on implicit default that comes from implementation of `IEquatable<T>`.

> Nevertheless, it doesn't really matter here - this generator will be hopefully removed soon.

:+1:

Dragemil (approved):

PR #289, opened 2021-07-09
Enable overriding keys of connection strings and Azure Key Vault configuration

### LINKED ISSUE #288: Cannot override database connection string environment key
The key value in override configuration is hardcoded: https://github.com/leancodepl/corelibrary/blob/84d3929b0ee5d5709d9b891019b061c4a20e6e71/src/Testing/LeanCode.IntegrationTestHelpers/ConfigurationOverrides.cs#L65
It should be configurable :see_no_evil:

Dragemil:
Closes #288


Saancreed (approved):
LGTM, but you could avoid some code duplication by reimplementing existing overloads of `AddAppConfigurationFromAzureKeyVault…` using the new ones by passing current defaults there as arguments.

Also, removing `ConfigurationOverrides` constructors is technically an ABI break, but we probably don't have to worry about it.

That said, let's see if the rest of the team is fine with these changes before merging.

lukaszgarstecki (approved):

src/Core/LeanCode.Components.Startup/IHostBuilderExtensions.cs
```diff
@@ -32,6 +32,24 @@ public static IHostBuilder AddAppConfigurationFromAzureKeyVault(this IHostBuilde
             });
         }
 
+        public static IHostBuilder AddAppConfigurationFromAzureKeyVault(
+            this IHostBuilder builder,
+            string vaultKeyOverride,
```
jakubfijalkowski, line 37: If this is mandatory, it is no longer an `override` IMO.

src/Testing/LeanCode.IntegrationTestHelpers/ConfigurationOverrides.cs
```diff
@@ -62,7 +62,7 @@ public override void Load()
 
                 Data = new Dictionary<string, string>
                 {
-                    ["ConnectionStrings:Database"] = dbConnStr,
+                    [connectionStringKey] = dbConnStr,
                     ["InternalBase"] = "http://localhost",
```
jakubfijalkowski: Let's also make these two configurable.

jakubfijalkowski (changes requested):

src/Core/LeanCode.Components.Startup/IHostBuilderExtensions.cs
Dragemil: It's not mandatory if constructor with default values exists.

jakubfijalkowski:
Some tests would be beneficial :P

jakubfijalkowski (approved):

PR #301, opened 2021-07-30
Operations

### LINKED ISSUE #188: New way of making combined query/command
So, we currently limit ourselves to have only commands and queries (CQRS style) as a public API. There are cases where it is not possible to correctly and ergonomically express the intention this way. Sometimes one needs to do "command with a result" (e.g. payments are much easier to model/handle if you do it in a one server call).

For now, we tend to build such endpoints using "normal" ASP.Net Core APIs but we do lose some of the functionality (e.g. authorization). We also rather freely define such endpoints - ignoring all the CQRS-y, service-level APIs. I think it is high time to introduce something new that doesn't force us to escape from the CQRS limitations.

I don't have anything particular in mind, rather - I want this issue to be discussion about the possibilities.

My initial idea is to build on the ASP.NET Core endpoint routing w/ MVC and just introduce some soft limitations/conventions on how we use that. For example:
1. We still define input/output DTOs in the `Contracts`,
2. We don't call the "domain" directly - we use `IQueryExecutor`/`ICommandExecutor`.
3. We don't do direct validation/authorization there, we rely on CQRS pipeline to do what is necessary.
4. We only map exceptions/validation errors to some predefined types (and this needs to be developed).

This way we would end up with a small "controller"-like classes that really behave like facades, but provide much better ergonomics for the client apps. Because of (1), we would still have the same "discoverability" as with our current approach. We can even put paths as `const`s in contracts.

What do you think?

marcin-dardzinski:
Adresses #188

I'd like to discuss one thing. Currently all of the operation executing components are exactly the same the ones for queries and almost the same as for commands. Are we okay with that or do we see the room for some unification?

### DIFF
--- a/docs/basics/02_cqrs.md
+++ b/docs/basics/02_cqrs.md
@@ -49,6 +49,26 @@ public class FindDishesMatchingName : IQuery<List<DishInfoDTO>>
 
 It finds all the dishes that match the name filter (however we define the filter). It may be called anonymously and returns a list of `DishInfoDTO`s (we use a `List` instead of a `IList` or `IReadOnlyList` because of the DTO constraint; `List` is more DTO-ish than any interface).
 
+### Operation
+
+The strict separation of command and query has sometimes led to awkwardness for clients consuming the API. So, the type `IOperation` was introduced to add more flexibility. Operations change the state of the system, but also allow to return some result. Operations itself are not validated, if executing operation results in executing a command, it's the developer responsibility to decide how to handle potential validation errors.
+
+```csharp
+public class PayForOrder : IOperation<PaymentTokenDTO>
+{
+    public Guid OrderId { get; set; }
+}
+
+public class PaymentTokenDTO { }
+```
+
+Potential use cases:
+
+1. Acting as a proxy for multiple commands (i.e. working on different domains) which are supposed to run at the same time and returning result, potentially making clients life's easier (one HTTP request, clean indication, that these things are connected)
+2. Running a command creating an object and immediately returning result. Note that in general our approach is command + immediate query with client generated id, but there might be reasons to violate this rule.
+3. Integration with external services that do not conform to the CQRS pattern. E.g. creating payment in a third-party api and immediately returning some result which should not be stored in our system's database.
+   1. Integration with services which combine object validation and creation steps, making it impossible to validate command separately in command validator
+
 ## Implementation
 
 Commands and queries define the API surface but they aren't the actual code that will be executed. That's the reason why there are command and query handlers.
@@ -79,12 +99,12 @@ public class CreateDishCH : ICommandHandler<AppContext, CreateDish>
 
 As you can see, the command handler is really simple - it just converts the command into new aggregate, tracking who owns the dish (`UserId` - they are the ones that have `CreateDish` permission). That does not mean this is the only responsibility of the handlers (it's just an example), but there are some guidelines related to them:
 
- 1. Keep them simple and testable, do not try to model whole flows with a single command,
- 2. Commands should rely on aggregates to gather the data (try not to use queries inside command handlers),
- 3. Commands should modify just a single aggregate (try to `AddAsync`/`UpdateAsync`/`DeleteAsync` at most once),
- 4. If the business process requires to modify multiple aggregates, try to use events and event handlers (but don't over-engineer),
- 5. If that does not help, modify/add/delete multiple aggregates,
- 6. Do not throw exceptions from inside commands. The client will receive generic error (`500 Internal Server Error`). Do it only as a last resort.
+1.  Keep them simple and testable, do not try to model whole flows with a single command,
+2.  Commands should rely on aggregates to gather the data (try not to use queries inside command handlers),
+3.  Commands should modify just a single aggregate (try to `AddAsync`/`UpdateAsync`/`DeleteAsync` at most once),
+4.  If the business process requires to modify multiple aggregates, try to use events and event handlers (but don't over-engineer),
+5.  If that does not help, modify/add/delete multiple aggregates,
+6.  Do not throw exceptions from inside commands. The client will receive generic error (`500 Internal Server Error`). Do it only as a last resort.
 
 #### Validation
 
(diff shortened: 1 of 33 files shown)


src/Domain/LeanCode.CQRS.Default/Wrappers/OperationHandlerWrapper.cs
```diff
@@ -0,0 +1,23 @@
+using System.Threading.Tasks;
+using LeanCode.CQRS;
+using LeanCode.CQRS.Execution;
+
+namespace LeanCode.CQRS.Default.Wrappers
+{
+    internal class OperationHandlerWrapper<TAppContext, TOperation, TResult> : IOperationHandlerWrapper
+    where TOperation : IOperation<TResult>
```
jakubfijalkowski: Indent +4

src/Tools/LeanCode.CodeAnalysis/DiagnosticsIds.cs
```diff
@@ -5,5 +5,6 @@ public static class DiagnosticsIds
         public const string CommandsShouldHaveAuthorizers = "LNCD0001";
         public const string QueriesShouldHaveAuthorizers = "LNCD0002";
         public const string CommandsShouldHaveValidators = "LNCD0003";
+        public const string OperationsShouldHaveAuthorizers = "LNCD0004";
```
Dragemil, line 8: What about validators?

src/Domain/LeanCode.CQRS/IRemoteOperation.cs
```diff
@@ -0,0 +1,4 @@
+namespace LeanCode.CQRS
+{
+    public interface IRemoteOperation<out TResult> : IOperation<TResult> { }
```
jakubfijalkowski: Let's get rid of it. We don't need the `IRemoteT`/`IT` distinction. No one uses it anymore, really.

jakubfijalkowski (changes requested):

jakubfijalkowski:
Docs, we need docs for this one. Badly.

src/Domain/LeanCode.CQRS/IRemoteOperation.cs
marcin-dardzinski: Do you mean, for all types?

jakubfijalkowski: In this PR only for this one. For v6 - I would remove all of them.

docs/basics/02_cqrs.md
```diff
@@ -49,6 +49,26 @@ public class FindDishesMatchingName : IQuery<List<DishInfoDTO>>
 
 It finds all the dishes that match the name filter (however we define the filter). It may be called anonymously and returns a list of `DishInfoDTO`s (we use a `List` instead of a `IList` or `IReadOnlyList` because of the DTO constraint; `List` is more DTO-ish than any interface).
 
+### Operation
+
+The strict separation of command and query has sometimes led to awkwardness for clients consuming the API. So, the type `IOperation` was introduced to add more flexibility. Operations change the state of the system, but also allow to return some result. Operations itself are not validated, if executing operation results in executing a command, it's the developer responsibility to decide how to handle potential validation errors.
```
jakubfijalkowski, line 54:
```suggestion
The strict separation of command and query has had sometimes led to awkwardness for clients consuming the API. So, the type `IOperation` was introduced to add more flexibility. Operations change the state of the system, but also allow to return some result. Operations itself are not validated, if executing operation results in executing a command, it's the developer's responsibility to decide how to handle and report potential validation errors.
```

docs/basics/02_cqrs.md
```diff
@@ -49,6 +49,26 @@ public class FindDishesMatchingName : IQuery<List<DishInfoDTO>>
...
+{
+    public Guid OrderId { get; set; }
+}
+
+public class PaymentTokenDTO { }
+```
+
+Potential use cases:
+
+1. Acting as a proxy for multiple commands (i.e. working on different domains) which are supposed to run at the same time and returning result, potentially making clients life's easier (one HTTP request, clean indication, that these things are connected)
```
jakubfijalkowski, line 67:
```suggestion
1. Acting as a proxy for multiple commands (i.e. working on different domains) which are supposed to run at the same time and returning result, potentially making clients life's easier (one HTTP request, clean indication, that these things are connected).
```

docs/basics/02_cqrs.md
```diff
@@ -49,6 +49,26 @@ public class FindDishesMatchingName : IQuery<List<DishInfoDTO>>
...
+2. Running a command creating an object and immediately returning result. Note that in general our approach is command + immediate query with client generated id, but there might be reasons to violate this rule.
+3. Integration with external services that do not conform to the CQRS pattern. E.g. creating payment in a third-party api and immediately returning some result which should not be stored in our system's database.
+   1. Integration with services which combine object validation and creation steps, making it impossible to validate command separately in command validator
```
jakubfijalkowski, line 70:
```suggestion
   a. Integration with services which combine object validation and creation steps, making it impossible to validate command separately in command validator
```

docs/basics/02_cqrs.md
```diff
@@ -49,6 +49,26 @@ public class FindDishesMatchingName : IQuery<List<DishInfoDTO>>
...
+}
+
+public class PaymentTokenDTO { }
+```
+
+Potential use cases:
+
+1. Acting as a proxy for multiple commands (i.e. working on different domains) which are supposed to run at the same time and returning result, potentially making clients life's easier (one HTTP request, clean indication, that these things are connected)
+2. Running a command creating an object and immediately returning result. Note that in general our approach is command + immediate query with client generated id, but there might be reasons to violate this rule.
+3. Integration with external services that do not conform to the CQRS pattern. E.g. creating payment in a third-party api and immediately returning some result which should not be stored in our system's database.
```
jakubfijalkowski, line 69:
```suggestion
3. Integration with external services that do not conform to the CQRS pattern. E.g. creating payment in a third-party API and immediately returning the payment which might not be stored in our system's database.
```

docs/basics/02_cqrs.md
```diff
@@ -156,6 +176,34 @@ public class FindDishesMatchingNameQH : IQueryHandler<AppContext, FindDishesMatc
 }
 ```
 
+### Operation handlers
+
+Operation handlers execute complex operations. They should not contain logic themselves, instead they should orchestrate
+commands, queries (and potentially other services) via `ICommandExecutor`, `IQueryExecutor` interfaces.
```
jakubfijalkowski, line 182:
```suggestion
commands, queries (and potentially other services) via `ICommandExecutor` and `IQueryExecutor` interfaces.
```

jakubfijalkowski (approved):
Some nits, feel free to ignore.

jakubfijalkowski (approved):

PR #306, opened 2021-08-13
Introduce leancode azure identity

marcin-dardzinski:
I'd like to standardize the way we authenticate to Azure services and a single source of truth when it comes to configuration. I opted out of `DefaultAzureCredential` as we have little control of how it behaves. So, `DefaultLeanCodeCredential` it is with all options used by us set explicitly. I also didn't use `ChainedTokenCredential`, we usually do not use a chain and explicitly use one method. 

The second KV and IdentityServre related change is that key vault clients are no longer created by `SigningService` instead they're expected to be registered in the DI container. It plays nicely with this pattern https://docs.microsoft.com/en-us/dotnet/azure/sdk/dependency-injection#register-client. The only think it bothers me is that it'd mean that we can have only one `CryptographyClient` registered (and it's created for a concrete key), but we can use named clients to overcome this limitation.

I also removed recently added key vault configuration key overrides, I think that if we need such thing, it's better to simply pass `TokenCredential` secrets need to use.

I'll prepare a sample of how it'd look like.


marcin-dardzinski:
Ok, I decided to switch to named clients, here's the sample of how it'd get registered in the app https://bitbucket.org/leancode-team/tools/commits/3daf0cfe5a77d1dcc16389fd0e03126749ecca79?at=feature/update-template#Ltemplates/content/backend/src/Apps/LNCDApp.MainApp/MainAppModule.csT79

Directory.Build.targets
```diff
@@ -68,6 +69,7 @@
     <PackageReference Update="Microsoft.EntityFrameworkCore.Sqlite" Version="$(EntityFrameworkCoreVersion)" />
     <PackageReference Update="Microsoft.EntityFrameworkCore.SqlServer" Version="$(EntityFrameworkCoreVersion)" />
 
+    <PackageReference Include="Microsoft.Extensions.Azure" Version="1.1.0" />
```
Saancreed: This probably shouldn't be included repo–wide.

src/Infrastructure/LeanCode.AzureIdentity/DefaultLeanCodeCredential.cs
```diff
@@ -0,0 +1,63 @@
...
+        static string? GetEnv(string rawKey) => Environment.GetEnvironmentVariable(rawKey.Replace(":", "__"));
+    }
+
+    public static TokenCredential Create(AzureCredentialConfiguration config)
+    {
+        if (config.UseManagedIdentity)
+        {
+            return new ManagedIdentityCredential(config.ClientId);
+        }
+        else if (config.UseAzureCLI)
```
Saancreed, line 49: I find it not very obvious that this flag is completely ignored when `UseManagedIdentity` is set. Can make it a tri–state enum or at least throw an exception when attempting to create such `AzureCredentialConfiguration`?

src/Infrastructure/LeanCode.AzureIdentity/AzureCredentialConfiguration.cs
```diff
@@ -0,0 +1,16 @@
+namespace LeanCode.AzureIdentity;
+
+public record AzureCredentialConfiguration
+{
+    public string? ClientId { get; init; }
+    public string? TenantId { get; init; }
```
Saancreed, line 5: I'd reorder this to match the order of initializers in `DefaultLeanCodeCredential.Create`, so `TenantId` before `ClientId`.

Same with keys below.

src/Infrastructure/LeanCode.AzureIdentity/DefaultLeanCodeCredential.cs
```diff
@@ -0,0 +1,63 @@
...
+
+public static class DefaultLeanCodeCredential
+{
+    public static TokenCredential Create(IConfiguration configuration)
+    {
+        var config = new AzureCredentialConfiguration
+        {
+            TenantId = Get<string>(AzureCredentialConfiguration.TenantIdKey),
+            ClientId = Get<string>(AzureCredentialConfiguration.TenantIdKey),
+            ClientSecret = Get<string>(AzureCredentialConfiguration.TenantIdKey),
```
Saancreed: `TenantIdKey`? Looks like a happy copy paste accident.

src/Infrastructure/LeanCode.AzureIdentity/DefaultLeanCodeCredential.cs
```diff
@@ -0,0 +1,63 @@
...
+        T Get<T>(string key) => configuration.GetValue<T>(key);
+    }
+
+    public static TokenCredential CreateFromEnvironment()
+    {
+        var config = new AzureCredentialConfiguration
+        {
+            TenantId = GetEnv(AzureCredentialConfiguration.TenantIdKey),
+            ClientId = GetEnv(AzureCredentialConfiguration.TenantIdKey),
+            ClientSecret = GetEnv(AzureCredentialConfiguration.TenantIdKey),
```
Saancreed: Another happy copy paste accident here.

src/Core/LeanCode.Components.Startup/IHostBuilderExtensions.cs
```diff
@@ -15,10 +14,7 @@ public static class IHostBuilderExtensions
     {
         public const string SystemLoggersEntryName = LeanProgram.SystemLoggersEntryName;
 
-        public const string VaultKey = "Secrets:KeyVault:VaultUrl";
-        public const string ClientIdKey = "Secrets:KeyVault:ClientId";
-        public const string ClientSecretKey = "Secrets:KeyVault:ClientSecret";
-        public const string TenantIdKey = "Secrets:KeyVault:TenantId";
+        public const string VaultKey = "KeyVault:VaultUrl";
```
Saancreed: We can rename it to `VaultUrlKey` to be more precise. Also I'd keep overrides that allow us to use another key.

Saancreed (changes requested):
Comments, plus `DisposeAsync` in `LeanCodeTestFactory.cs` should be fixed as well.

Directory.Build.targets
```diff
@@ -68,6 +69,7 @@
     <PackageReference Update="Microsoft.EntityFrameworkCore.Sqlite" Version="$(EntityFrameworkCoreVersion)" />
     <PackageReference Update="Microsoft.EntityFrameworkCore.SqlServer" Version="$(EntityFrameworkCoreVersion)" />
 
+    <PackageReference Update="Microsoft.Extensions.Azure" Version="1.1.0" />
```
Saancreed: Now it's a duplicate of line 25 :upside_down_face:

src/Infrastructure/LeanCode.AzureIdentity/AzureCredentialConfiguration.cs
Saancreed: Only half done.

Saancreed (approved):

PR #311, opened 2021-08-20
SId

marcin-dardzinski:


src/Domain/LeanCode.DomainModels/Model/SId.cs
```diff
@@ -0,0 +1,162 @@
+using System.Reflection;
+using LeanCode.DomainModels.Serialization;
+
+namespace LeanCode.DomainModels.Model
+{
+    /// <summary>A stronly typed id consisting of a guid prefixed with a lower-cased name of an entity.</summary>
```
Dragemil, line 7: ...consisting of a guid with its dashes skipped...

Dragemil: Or do we want to stop skipping those dashes?

src/Domain/LeanCode.DomainModels/Model/SId.cs
```diff
@@ -0,0 +1,162 @@
...
+        public bool IsEmpty => value is null || value == Empty;
+
+        private SId(string v)
+        {
+            value = v ?? throw new ArgumentNullException(nameof(v));
+        }
+
+        public SId(Guid v)
+        {
+            value = TypeName + Separator + v.ToString("N");
```
Saancreed: Use string interpolation instead to avoid `ToString` call on the `Guid`.

Or, better yet, `value = string.Create(null, stackalloc char[RawLength], $"{TypeName}{Separator}{v:N}");`

src/Domain/LeanCode.DomainModels/Model/SId.cs
marcin-dardzinski: I'd skip them but I see no point in being explicit about in comments, you should never rely on that...

src/Domain/LeanCode.DomainModels/Model/SId.cs
```diff
@@ -0,0 +1,162 @@
...
+                return true;
+            }
+            else
+            {
+                id = default;
+                return false;
+            }
+        }
+
+        public static bool IsValid(string? v)
```
Saancreed: `[NotNullWhen(true)] string? v`

Saancreed (changes requested):

src/Domain/LeanCode.DomainModels/Model/SId.cs
```diff
@@ -0,0 +1,162 @@
...
+        public static readonly SId<TEntity> Empty = new(Guid.Empty);
+
+        private readonly string? value;
+
+        public string Value => value ?? Empty.Value;
+        public bool IsEmpty => value is null || value == Empty;
+
+        private SId(string v)
+        {
+            value = v ?? throw new ArgumentNullException(nameof(v));
```
Saancreed: By the way, do we need this check here? `v` is already marked as non–null and the constructor is `private` so we shouldn't have to verify twice.

Saancreed (approved):

PR #324, opened 2021-11-03
Add UserIdLogsCorrelationMiddleware

lukaszgarstecki:
Feature lost during purge of `LeanCode.Correlation`

### DIFF
--- a/src/Infrastructure/LeanCode.Logging/UserIdLogsCorrelationMiddleware.cs
+++ b/src/Infrastructure/LeanCode.Logging/UserIdLogsCorrelationMiddleware.cs
@@ -24,7 +24,7 @@ namespace LeanCode.Logging
         {
             if (httpCtx?.User.Identity?.IsAuthenticated ?? false)
             {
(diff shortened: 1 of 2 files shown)


src/Infrastructure/LeanCode.Logging/UserIdLogsCorrelationMiddleware.cs
```diff
@@ -0,0 +1,38 @@
...
+
+                await next();
+            });
+        }
+
+        private static IDisposable? UserId(HttpContext httpCtx, string? userIdClaim)
+        {
+            if (httpCtx?.User.Identity?.IsAuthenticated ?? false)
+            {
+                var userId = userIdClaim is not null ? httpCtx.User.FindFirstValue(userIdClaim) : httpCtx.User.Identity.Name;
```
jakubfijalkowski: It would be better to assume `sub` as the claim name if `userIdClaim` is null IMO. As far as I know, `User.Identity.Name` might contain strange data.

Saancreed (approved):

jakubfijalkowski (approved):

PR #343, opened 2021-11-08
Mass transit config revamp

### LINKED ISSUE #326: Rework MT configuration
The current configuration is deprecated, we should revise it.

marcin-dardzinski:
Resolves #326

Few things addressed here, some open for discussion:
1. `MassTransitRelayModule` is now an abstract class with a method left to implement to configure MT. This method should be responsible for registering consumers (initially it was done by the corelib code) and configuring the bus. IMO, this approach is better than what we are currently doing in our projects (static `MassTransitConfiguration` class with a bit awkward injection of a bus config delegate)
2. Our `MassTransitRelayHostedService` is ditched, MassTransit has it's owned hosted service for a long time - it's better and it supports multi bus.
3. The new way of configuring MT expects clients to use `Microsoft.Extensions.DependencyInjection` container. In general this container is better supported and some features are not available. On the other hand, there is one problem with it - we have to manually inject the lifetime scope for the application. There is one awkward test which ensures it works as expected. If we stay with Autofac registration, there's no such problem. 
4. Given the problems with 3. we may not enforce usage of `Microsoft.Extensions.DependencyInjection` and simply allow users to register MT as they wish.


src/Domain/LeanCode.DomainModels.MassTransitRelay/InjectLifetimeScopeFilter.cs
```diff
@@ -0,0 +1,38 @@
+using GreenPipes;
+using MassTransit;
+using MassTransit.Registration;
+
+namespace LeanCode.DomainModels.MassTransitRelay
+{
+    // When combining Autofac and Microsoft.Extensions.Dependency injection
+    // some DI components are not injected correctly into the message payload,
+    // so we have to fix it by ourselves.
+    public class InjectLifetimeScopeFilter<T> : IFilter<ConsumeContext<T>>
```
jakubfijalkowski, line 10: Tests, maybe?

src/Domain/LeanCode.DomainModels.MassTransitRelay/MassTransitRelayModule.cs
```diff
@@ -1,37 +1,27 @@
...
 using LeanCode.DomainModels.MassTransitRelay.Simple;
 using LeanCode.PeriodicService;
 using MassTransit;
-using MassTransit.AutofacIntegration;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace LeanCode.DomainModels.MassTransitRelay
 {
-    public class MassTransitRelayModule : AppModule
+    public abstract class MassTransitRelayModule : AppModule
```
jakubfijalkowski, line 13: I think the module should _enforce_ calling `AddMassTransit` and just allow to change whats specified inside the closure (or after the call).

src/Domain/LeanCode.DomainModels.MassTransitRelay/InjectLifetimeScopeFilter.cs
marcin-dardzinski: Do you mean a unit test for this particular class? I can add it although the only think it'll test it's whether or not I understood MT code correctly. There is a larger test [here](https://github.com/leancodepl/corelibrary/pull/343/files#diff-396609625e6c0a2ebaa2d90da2cc9ffdbb02128b67f5a3496b93242acc6b2c87R54), checking if everything works as expected.

src/Domain/LeanCode.DomainModels.MassTransitRelay/MassTransitRelayModule.cs
marcin-dardzinski: I don't think it'll be practical. Also, I think we may give client flexibility which container to use (say, have to abstract methods methods, so that they implement one and leave the second one empty). Also, sometimes we may not want to call  plain `AddMassTransit`, e.g. when using MultiBus (and we do use it in one project).

src/Domain/LeanCode.DomainModels.MassTransitRelay/InjectLifetimeScopeFilter.cs
jakubfijalkowski: I mean unit-test style tests. They will make it easier for future devs to understand your assumptions about MT. :)

lukaszgarstecki (approved):

PR #347, opened 2021-11-09
Update .NET to 6.0 RTM

jakubfijalkowski:
Instead of multiple Dependabot PRs ;)


Saancreed:
I'd suggest also updating SDK version in `global.json` and workflows.

Saancreed (approved):

PR #349, opened 2021-11-10
Fix warnings introduced with .NET 6

jakubfijalkowski:
This are proposed changes - some rules are disabled, some are fixed.
Some fixes introduced breaking changes (although small), so these might be candidates to exclude
from the PR.


LeanCode.CodeAnalysis.ruleset
```diff
@@ -11,6 +11,32 @@
...
+        <Rule Id="CA1720" Action="None"/>
+        <!-- IdentifiersShouldNotContainTypeNames - we don't really care about it -->
+        <Rule Id="CA1019" Action="None"/>
+        <!-- DefineAccessorsForAttributeArguments - not really needed -->
+        <Rule Id="CA2234" Action="None"/>
+        <!-- PassSystemUriObjectsInsteadOfStrings - most of the time it will result in an exception -->
+        <Rule Id="CA1303" Action="None"/>
+        <!-- DoNotPassLiteralsAsLocalizedParameters - we don't support localization in places where this is enabled -->
+        <Rule Id="CA1014" Action="None"/>
+        <!-- MarkAssembliesWithClsCompliant - we are not CLS compliant deliberately -->
```
Saancreed: Can we keep them ordered by Id?

src/Domain/LeanCode.CQRS/CanBeNullAttribute.cs
```diff
@@ -4,5 +4,5 @@ namespace LeanCode.CQRS
 {
     [Obsolete("Use C# 8.0's Nullable Reference Types language feature instead.")]
     [AttributeUsage(AttributeTargets.Property, AllowMultiple = false, Inherited = false)]
-    public class CanBeNullAttribute : Attribute { }
+    public sealed class CanBeNullAttribute : Attribute { }
```
Saancreed: We should probably dump it already, new generator doesn't even support it and the old one is already gone.

src/Domain/LeanCode.CQRS/Security/AuthorizeWhenHasAnyOfAttribute.cs
```diff
@@ -2,15 +2,13 @@
...
+    [SuppressMessage("?", "CA1715", Justification = "Authorizer interfaces are exempt.")]
+    [SuppressMessage("?", "CA1040", Justification = "Marker interface.")]
     public interface HasPermissions { }
 
-    public class AuthorizeWhenHasAnyOfAttribute : AuthorizeWhenAttribute
+    public sealed class AuthorizeWhenHasAnyOfAttribute : AuthorizeWhenAttribute
     {
         public AuthorizeWhenHasAnyOfAttribute(params string[] permissions)
-            : base(typeof(HasPermissions), permissions) { }
+               : base(typeof(HasPermissions), permissions) { }
```
Saancreed: Unnecessary change?

src/Domain/LeanCode.CQRS/Security/AuthorizeWhenHasAnyOfAttribute.cs
```diff
@@ -2,15 +2,13 @@
 
 namespace LeanCode.CQRS.Security
 {
-    [SuppressMessage(
-        "StyleCop.CSharp.NamingRules",
-        "SA1302:InterfaceNamesMustBeginWithI",
-        Justification = "Authorizer interfaces are exempt.")]
+    [SuppressMessage("?", "CA1715", Justification = "Authorizer interfaces are exempt.")]
+    [SuppressMessage("?", "CA1040", Justification = "Marker interface.")]
     public interface HasPermissions { }
```
Saancreed: Actually, I'd just rename it to `IHasPermissions` / `IHavePermissions`, afaik this type is not used directly by any project :peepo_hide:

src/Domain/LeanCode.CQRS/Security/AuthorizeWhenAttribute.cs
```diff
@@ -6,12 +6,12 @@
...
 {
     [AttributeUsage(AttributeTargets.Class | AttributeTargets.Interface, AllowMultiple = true, Inherited = true)]
-    public class AuthorizeWhenAttribute : Attribute
+    public abstract class AuthorizeWhenAttribute : Attribute
     {
         private readonly Type authorizerType;
         private readonly object? customData;
 
-        public AuthorizeWhenAttribute(Type authorizerType, object? customData = null)
+        protected AuthorizeWhenAttribute(Type authorizerType, object? customData = null)
```
Saancreed, line 14: Afaik we already have `<LangVersion>preview</LangVersion>`, so maybe let's take advantage of that and make this attribute generic `AuthorizeWhenAttribute<TAuthorizer>`?

Saancreed (commented):
I've got a bunch of suggestions, but nothing too severe that should be applied asap. Most of them are quite unrelated to changes introduces in this PR, so we can leave them for later.

Not approving yet as it's still marked as draft; I'll re-review once ready.

LeanCode.CodeAnalysis.ruleset
jakubfijalkowski: Not yet - I want to discuss them first, and then apply all the cleaning.

src/Domain/LeanCode.CQRS/Security/AuthorizeWhenHasAnyOfAttribute.cs
jakubfijalkowski: Ugh, yep - OmniSharp broke formatting here. :(

src/Domain/LeanCode.CQRS/Security/AuthorizeWhenHasAnyOfAttribute.cs
jakubfijalkowski: No, let's keep the convention everywhere.

src/Domain/LeanCode.CQRS/Security/AuthorizeWhenAttribute.cs
jakubfijalkowski: Don't want to - we want to force new, correctly named attributes for each and every new authorization-like attribute.

Saancreed: Of course, and we can still do that if we keep the constructor `protected` to disallow using the generic type directly. Still, I think having something like `public class AuthorizeWhenOwnsOrderAttribute : AuthorizeWhenAttribute<(I)OwnOrder> { … }` would be _some_ improvement over `: base(typeof(OwnOrder), …)`.

src/Domain/LeanCode.CQRS/Security/AuthorizeWhenHasAnyOfAttribute.cs
jakubfijalkowski: Or... let's get rid of the convention altogether, but let's discuss it first on a forum.

src/Domain/LeanCode.CQRS/Security/AuthorizeWhenAttribute.cs
jakubfijalkowski: That's a really good point. I don't really want to add another breaking change (esp. that we would still need some form of not-generic class for proper discovery), but this won't be a breaking change it occured to me. I'll add the new generic helper.

LeanCode.CodeAnalysis.ruleset
```diff
@@ -11,6 +11,32 @@
...
+        <Rule Id="CA1030" Action="None"/>
+        <!-- UseEventsWhereAppropriate - we won't use events anyway -->
+        <Rule Id="CA1308" Action="None"/>
+        <!-- NormalizeStringsToUppercase - we know what we're doing -->
+        <Rule Id="CA1720" Action="None"/>
+        <!-- IdentifiersShouldNotContainTypeNames - we don't really care about it -->
+        <Rule Id="CA1019" Action="None"/>
+        <!-- DefineAccessorsForAttributeArguments - not really needed -->
+        <Rule Id="CA2234" Action="None"/>
+        <!-- PassSystemUriObjectsInsteadOfStrings - most of the time it will result in an exception -->
```
marcin-dardzinski, line 38: Why 'most of the time it will result in an exception'? Anyway, I think it's a sane rule.

jakubfijalkowski: When you use `HttpClient` with `BasePath`, and pass relative path to the calls, you can't use simple `new Uri("/mypath")` as it expects absolute paths. So simple changes there just result in wrong code.

Dragemil (approved):

src/Infrastructure/LeanCode.ExternalIdentityProviders/Facebook/FacebookClient.cs
```diff
@@ -157,13 +164,28 @@ private string GenerateProof(string accessToken)
...
                 : $"{ApiVersion}/{uri}?access_token={accessToken}{proof}";
         }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!disposedValue)
+            {
+                if (disposing)
+                {
+                    hmac?.Dispose();
```
Saancreed, line 184: Actually, I would prefer if we dumped the `HMACSHA256` instance and used one of the [new and shiny static overloads](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.hmacsha256.hashdata?view=net-6.0#System_Security_Cryptography_HMACSHA256_HashData_System_ReadOnlySpan_System_Byte__System_ReadOnlySpan_System_Byte__System_Span_System_Byte__) instead of making the whole class `IDisposable`.

Saancreed (approved):
One more suggestion with a better (I think) solution to disposability of `FacebookClient`, otherwise LGTM.

src/Infrastructure/LeanCode.ExternalIdentityProviders/Facebook/FacebookClient.cs
jakubfijalkowski: Let's create an issue and track it there.

Saancreed: Done, we have #352 to track this now.

marcin-dardzinski (approved):

ZoltePudeleczko (approved):

lukaszgarstecki (approved):

PR #354, opened 2021-11-22
Pass database name using Database keyword instead of Initial Catalog

Saancreed:
Npgsql does not understand the latter.


marcin-dardzinski (approved):
I wonder if we should come up with a more future proof solution, but I guess we'll worry the next time.

Saancreed:
> I wonder if we should come up with a more future proof solution, but I guess we'll worry the next time.

I was initially considering another `ConfigurationOverrides` constructor parameter, but that _would_ be a breaking change. Maybe it's something to revisit when .NET 7.0 comes.

jakubfijalkowski (approved):

PR #361, opened 2021-11-29
Remove `RowVersion` property from `IOptimisticConcurrency` interface

Saancreed:
It's now an optional shadow property.

As discussed last week. The other overload is for when the database used something other than `byte[]` as the type for row version column.

When using Npgsql we can now configure aggregates like so:

```csharp
builder.Entity<Foo>().UseXminAsConcurrencyToken().IsOptimisticConcurrent(addRowVersion: false);
```

Fwiw I believe these methods should return the builder so multiple calls can be chained, but I don't care that much.

Btw, do we need an ADR here?

### DIFF
--- a/docs/adrs/2021-11-25_optimistic_concurrency.md
+++ b/docs/adrs/2021-11-25_optimistic_concurrency.md
@@ -1,0 +1,42 @@
+# How to handle concurrency token columns in non–MSSQL databases
+
+The `IOptimisticConcurrency` interface requires `DateTime DateModified` and `byte[] RowVersion` properties to be implemented by aggregate roots. However, the `byte[]` representation of `RowVersion` column is specific to Microsoft SQL Server and is not compatible with other databases, such as PostgreSQL and its convention to use `uint xmin` hidden column as a concurrency token.
+
+## Status
+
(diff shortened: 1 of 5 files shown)


marcin-dardzinski:
> Btw, do we need an ADR here?

Since you've suggested :stuck_out_tongue:

Dragemil (approved):

Saancreed:
> > Btw, do we need an ADR here?
> 
> Since you've suggested :stuck_out_tongue:

I had not! I merely _asked_ whether we need it, that was very much not intended to suggest that we do :sweat_smile:

marcin-dardzinski (approved):

lukaszgarstecki (approved):

jakubfijalkowski:
> > > Btw, do we need an ADR here?
> > 
> > 
> > Since you've suggested stuck_out_tongue
> 
> I had not! I merely _asked_ whether we need it, that was very much not intended to suggest that we do sweat_smile

The ADR would be really, really nice. :)

jakubfijalkowski (approved):

lukaszgarstecki (approved):

PR #372, opened 2022-02-01
Fix MT Serializer constructor handling

lukaszgarstecki:
Fixes issue that MT used incorrect (public constructor with default values instead of private parameterless consturctor) constructor when deserializing messages.

### DIFF
--- a/src/Domain/LeanCode.DomainModels.MassTransitRelay/Outbox/NewtonsoftJsonEventsSerializer.cs
+++ b/src/Domain/LeanCode.DomainModels.MassTransitRelay/Outbox/NewtonsoftJsonEventsSerializer.cs
@@ -23,6 +23,7 @@ namespace LeanCode.DomainModels.MassTransitRelay.Outbox
             this.typesCatalog = typesCatalog;
             this.settings = settings ?? new()
             {
+                ConstructorHandling = ConstructorHandling.AllowNonPublicDefaultConstructor,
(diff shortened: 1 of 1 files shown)


src/Domain/LeanCode.DomainModels.MassTransitRelay/Outbox/NewtonsoftJsonEventsSerializer.cs
```diff
@@ -23,6 +23,7 @@ public NewtonsoftJsonEventsSerializer(TypesCatalog typesCatalog, JsonSerializerS
             this.typesCatalog = typesCatalog;
             this.settings = settings ?? new()
             {
+                ConstructorHandling = ConstructorHandling.AllowNonPublicDefaultConstructor,
                 ContractResolver = new ContractResolver(),
```
Saancreed, line 27: While we are already here, can we make this class public so the next consumer that comes here with a need to pass custom `settings` won't have to copy it again?

Saancreed (approved):

src/Domain/LeanCode.DomainModels.MassTransitRelay/Outbox/NewtonsoftJsonEventsSerializer.cs
lukaszgarstecki: Out of scope of this PR but we should do that I guess

PR #374, opened 2022-02-01
Make contracts resolver public

### LINKED ISSUE #373: Make `ContractResolver` class public
While we are already here, can we make this class public so the next consumer that comes here with a need to pass custom `settings` won't have to copy it again?

_Originally posted by @Saancreed in https://github.com/leancodepl/corelibrary/pull/372#discussion_r796574326_

lukaszgarstecki:
Resolves #373

### DIFF
--- a/src/Domain/LeanCode.DomainModels.MassTransitRelay/Outbox/NewtonsoftJsonEventsSerializer.cs
+++ b/src/Domain/LeanCode.DomainModels.MassTransitRelay/Outbox/NewtonsoftJsonEventsSerializer.cs
@@ -47,23 +47,23 @@ namespace LeanCode.DomainModels.MassTransitRelay.Outbox
                 .FirstOrDefault()
                 ?? throw new InvalidOperationException($"Type {type} is not defined in any of the provided assemblies.");
         }
+    }
 
-        private class ContractResolver : DefaultContractResolver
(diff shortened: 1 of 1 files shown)


src/Domain/LeanCode.DomainModels.MassTransitRelay/Outbox/NewtonsoftJsonEventsSerializer.cs
```diff
@@ -47,23 +47,23 @@ private Type GetEventType(string type)
                 .FirstOrDefault()
                 ?? throw new InvalidOperationException($"Type {type} is not defined in any of the provided assemblies.");
         }
+    }
 
-        private class ContractResolver : DefaultContractResolver
-        {
-            protected override JsonProperty CreateProperty(MemberInfo member, MemberSerialization memberSerialization)
-                => MakeWriteable(base.CreateProperty(member, memberSerialization), member);
+    public class ContractResolver : DefaultContractResolver
```
jakubfijalkowski: If we're making it public, let's rename it to something more descriptive, e.g. `NewtonsoftJsonContractResolver`.

jakubfijalkowski (changes requested):

jakubfijalkowski (approved):

PR #389, opened 2022-03-29
Downgrade Microsoft.CodeAnalysis to latest stable

jakubfijalkowski:


Directory.Build.targets
```diff
@@ -11,7 +11,7 @@
     <SystemIdentityModelVersion>6.16.0</SystemIdentityModelVersion>
 
     <!-- Do not bump these dependencies if you don't want to force users to use newer .NET Core SDK -->
-    <CodeAnalysisVersion>4.2.0-2.final</CodeAnalysisVersion>
+    <CodeAnalysisVersion>4.1.0</CodeAnalysisVersion>
```
Saancreed, line 14: What was wrong with `4.2.0-2.final`? Does it require .NET 7.0 preview SDK or…?

jakubfijalkowski: The `2.final` is a prerelease version in terms of SemVer, which I overlooked - this means it is not distributed with the stable SDK yet.

Saancreed (approved):

Issue #382, opened 2022-03-17
Firebase.FCM does not work with PostgreSQL


wojtek2288:
Using postgres, FCMClient, SendToUsersAsync throws an exception when passing `HashSet<Guid>` and expecting `IReadOnlySet<Guid>`

```
[12:31:21 ERR] R-FAULT loopback://localhost/***-api-events be987b02-09f5-4ae5-be93-382c7357a9c1 ***.Core.Domain.Events.Events.EventLocationOrTimingChanged ***.Core.Services.Processes.Firebase.NotifyOnEventLocationOrTimingChanged(00:00:00.0153645)
***-integration_tests-1  | System.NotSupportedException: IEnumerable parameters are not supported, pass an array or List instead
***-integration_tests-1  |    at Npgsql.TypeMapping.ConnectorTypeMapper.ResolveByClrType(Type type)
***-integration_tests-1  |    at Npgsql.NpgsqlParameter.ResolveHandler(ConnectorTypeMapper typeMapper)
***-integration_tests-1  |    at Npgsql.NpgsqlParameterCollection.ValidateAndBind(ConnectorTypeMapper typeMapper)
***-integration_tests-1  |    at Npgsql.NpgsqlCommand.ExecuteReader(CommandBehavior behavior, Boolean async, CancellationToken cancellationToken)
***-integration_tests-1  |    at Npgsql.NpgsqlCommand.ExecuteReader(CommandBehavior behavior, Boolean async, CancellationToken cancellationToken)
***-integration_tests-1  |    at Npgsql.NpgsqlCommand.ExecuteDbDataReaderAsync(CommandBehavior behavior, CancellationToken cancellationToken)
***-integration_tests-1  |    at Dapper.SqlMapper.QueryAsync[T](IDbConnection cnn, Type effectiveType, CommandDefinition command) in /_/Dapper/SqlMapper.Async.cs:line 418
***-integration_tests-1  |    at LeanCode.Dapper.DbContextDapperExtensions.WithConnectionAsync[TResult](DbContext context, Func`2 call)
***-integration_tests-1  |    at LeanCode.Firebase.FCM.EntityFramework.PushNotificationTokenStore`1.GetTokensAsync(IReadOnlySet`1 userIds, CancellationToken cancellationToken)
[...]

PR #391, opened 2022-03-30
Change IReadOnlySet, IEnumerable in notification queries, executes to List

wojtek2288:
#382 Firebase.FCM does not work with PostgreSQL


jakubfijalkowski (approved):

Saancreed (approved):
Would be nice to make sure this actually fixes the thing before merging.

lukaszgarstecki (approved):

ZoltePudeleczko (approved):

wojtek2288:
Still doesn't fix it for Postgre-SQL due to [https://github.com/DapperLib/Dapper/issues/1324](url). In our project we modify it to be Postgre-SQL specific, so probably won't be merging.

PR #392, opened 2022-03-31
Upgrade to MassTransit V8

Dragemil:
Biggest breaking change was probably end of 3rd party DI containers support so now it relays mostly on `Microsoft.Extensions.DependencyInjection`. Tests are passing so it seems it is fine.

Be prepared for the `Newtonsoft.Json` purge, because it's high time to stop supporting both serializers.

I leave it in draft state, because it is a breaking change that needs to target v6.1 CoreLib branch which is not prepared yet.


jakubfijalkowski:
image
Codecov, go home, you're drunk...

jakubfijalkowski (approved):

src/Domain/LeanCode.DomainModels.MassTransitRelay/IPipeContextServiceResolver.cs
```diff
@@ -15,14 +15,14 @@ TService GetService<TService>(PipeContext pipe)
 
     public class AutofacPipeContextServiceResolver : IPipeContextServiceResolver
```
marcin-dardzinski: Do we need this class anymore? It was used when we had to use service locator pattern in filters, but now we rely on MT.

src/Domain/LeanCode.DomainModels.MassTransitRelay/InjectLifetimeScopeFilter.cs
```diff
@@ -15,19 +13,19 @@ public void Probe(ProbeContext context)
 
         public Task Send(ConsumeContext<T> context, IPipe<ConsumeContext<T>> next)
         {
-            if (!context.TryGetPayload(out Autofac.Extensions.DependencyInjection.AutofacServiceProvider scopeProvider))
+            if (!context.TryGetPayload(out Autofac.Extensions.DependencyInjection.AutofacServiceProvider? scopeProvider))
```
marcin-dardzinski: Now I wonder if this class is necessary too. We needed it to patch MT Autofac integration, but right now it's all Microsoft.Extensions

jakubfijalkowski (changes requested):
@marcin-dardzinski makes a great point - since MT does not use Autofac anymore, we should try to remove all of our Autofac or confirm that we won't be able to.

Dragemil:
I attempted to remove everything related to MT Autofac integration and I hope it is not needed anymore.

jakubfijalkowski:
Let's not put faith in it and add tests for that (if we don't have already) :)

Dragemil:
Alright, bad wording - `ScopedFiltersTests` are testing that and they're passing.

jakubfijalkowski (approved):

marcin-dardzinski (approved):

PR #393, opened 2022-04-01
Split Firebase.FCM to db specific implementations

wojtek2288:


jakubfijalkowski:
Although I'm OK with the overall implementation, I'm not so sure about the naming. .NET uses variation of "SqlServer" (or `SqlClient`) and going with something like that would be "more consistent" with naming the other package with "PostgreSQL" suffix.

Also, we probably should have tests for this (but that might be out of scope of this PR).

src/Infrastructure/LeanCode.Firebase.FCM.PostgreSQL/PostgreSQLPushNotificationEntity.cs
```diff
@@ -0,0 +1,25 @@
...
+        public static void Configure(ModelBuilder builder)
+        {
+            builder.Entity<PostgreSQLPushNotificationEntity>(c =>
+            {
+                c.HasKey(e => e.Id);
+                c.HasIndex(e => e.UserId);
+                c.HasIndex(e => e.Token).IsUnique(true);
+
+                c.Property(e => e.Id).ValueGeneratedNever();
+                c.Property(e => e.Token).IsRequired(true).HasMaxLength(512);
```
Saancreed: We can drop `HasMaxLength` for Postgres.

test/Infrastructure/LeanCode.Firebase.FCM.Tests/MSSQLTokenStore/MSSQLTestDbContext.cs
```diff
@@ -1,30 +1,28 @@
...
             connection = Database.GetDbConnection();
         }
 
-        public static async Task<TestDbContext> CreateInMemory()
+        public static async Task<MSSQLTestDbContext> CreateInMemory()
         {
             SqlMapper.AddTypeHandler(new GuidTypeHandler());
 
-            var context = new TestDbContext(new DbContextOptionsBuilder<TestDbContext>()
+            var context = new MSSQLTestDbContext(new DbContextOptionsBuilder<MSSQLTestDbContext>()
```
Saancreed: It's not exactly MSSQL in this case :see_no_evil:

src/Infrastructure/LeanCode.Firebase.FCM.MSSQL/MSSQLPushNotificationTokenEntity.cs
```diff
@@ -1,9 +1,8 @@
-using System;
 using Microsoft.EntityFrameworkCore;
 
-namespace LeanCode.Firebase.FCM.EntityFramework
+namespace LeanCode.Firebase.FCM.MSSQL
 {
-    public class PushNotificationTokenEntity
+    public class MSSQLPushNotificationTokenEntity
```
Saancreed: What @jakubfijalkowski said, but I feel like we can rename packages to `.SqlServer` and `.PostgreSQL` while keeping the actual class names shortened to `MsSqlPush…` and `PgSqlPush…`. Just not all uppercase letters please.

src/Infrastructure/LeanCode.Firebase.FCM.PostgreSQL/PostgreSQLPushNotificationTokenStore.cs
```diff
@@ -0,0 +1,150 @@
...
+        {
+            if (userIds.Count > MaxTokenBatchSize)
+            {
+                throw new ArgumentException($"You can only pass at most {MaxTokenBatchSize} users in one call.", nameof(userIds));
+            }
+
+            var userIdsList = userIds.ToList();
+
+            var res = await dbContext.QueryAsync<UserToken>(
+                $"SELECT {GetTokensColumnName(nameof(PostgreSQLPushNotificationEntity.UserId))}, {GetTokensColumnName(nameof(PostgreSQLPushNotificationEntity.Token))} FROM {GetTokensTableName()} WHERE {GetTokensColumnName(nameof(PostgreSQLPushNotificationEntity.UserId))} = ANY(@userIdsList)",
```
Saancreed: Btw we should ensure that Dapper doesn't do anything unexpected with `ANY` when the list is empty or has only one element.

test/Infrastructure/LeanCode.Firebase.FCM.Tests/MSSQLTokenStore/MSSQLTestDbContext.cs
wojtek2288: Rename it to SqliteTestDbContext ? Because for testing PostgreSQL token store we will have to use something different than Sqlite

jakubfijalkowski: I would keep it as-is for now. We need to use proper database for testing after this split, but it is better to have _something_ than to have _nothing_ at all.

src/Infrastructure/LeanCode.Firebase.FCM.PostgreSQL/PostgreSQLPushNotificationTokenStore.cs
wojtek2288: Still works fine

Saancreed (approved):
It still feels kinda weird to test `MsSqlPushNotificationStore` using SQLite but oh well.

LGTM but please rebase on top of and retarget to `v6.1`, we probably don't want any breaking changes on `v6.0` anymore.

jakubfijalkowski (approved):

PR #394, opened 2022-04-04
Feature: Purge `Newtonsoft.Json` where possible

Dragemil:
What we achieve here is that systems using our library won't need to use `Newtonsoft.Json` at all, since it is inconvenient to support both serializers and `System.Text.Json` is still required. A breaking change for sure.

However, 'where possible' is not much places - `LeanCode.DomainModels.MassTransitRelay` mostly and some unit tests. Other packages relay on Newtonsoft indirectly, so we have to leave it.


src/Domain/LeanCode.DomainModels.MassTransitRelay/Outbox/SystemTextJsonEventsSerializer.cs
```diff
@@ -1,18 +1,17 @@
...
-using System.Linq;
 using System.Text.Json;
 using LeanCode.Components;
 
 namespace LeanCode.DomainModels.MassTransitRelay.Outbox
 {
     /// <remarks>
-    /// This is not usable yet. `System.Text.Json` does not support private constructors and requires
-    /// users to add `JsonIncludeAttribute` to properties with private setters.
+    /// This is hardly usable. `System.Text.Json` does not support private constructors and requires
```
jakubfijalkowski: Let's remove the `This is hardly usable` comment, since from now on this will be the only one ;P

lukaszgarstecki (approved):

jakubfijalkowski (changes requested):
Let's also check if the comments regarding "newtonsoft.json" are still valid, and if not - let's get rid of the direct dependencies.

src/Domain/LeanCode.CQRS/CommandResult.cs
```diff
@@ -17,7 +18,13 @@ public CommandResult(IEnumerable<ValidationError>? validationErrors)
                 : validationErrors.ToImmutableList();
         }
 
-        public static CommandResult Success { get; } = new CommandResult(null);
+        [JsonConstructor]
```
jakubfijalkowski: Do we need that? Won't a single ctor work?

Dragemil: We may have a single ctor, but it needs to map all properties directly. I'll leave the recently provided one and remove the old one.

Dragemil:
Apparently managed to remove `Newtonsoft` references in everywhere but Sendgrid, since it uses explicit `[JsonIgnore]`s.

jakubfijalkowski (approved):

PR #404, opened 2022-06-20
`LeanCode.Contracts`

Saancreed:
As side–victims of this rework, `IRemote{Command,Query}` are now gone.


src/Tools/LeanCode.CodeAnalysis/Analyzers/EnsureCommandsAndQueriesHaveAuthorizers.cs
```diff
@@ -10,9 +10,9 @@ public class EnsureCommandsQueriesAndOperationsHaveAuthorizers : DiagnosticAnaly
     {
```
jakubfijalkowski, line 10: We need to rename the file also (probably a separate issue/PR).

jakubfijalkowski (approved):

src/Tools/LeanCode.CodeAnalysis/Analyzers/EnsureCommandsAndQueriesHaveAuthorizers.cs
Saancreed: Actually I'd prefer to rename it now so I don't forget about it later.

PR #413, opened 2022-10-26
Bump all dependencies

jakubfijalkowski:
A couple of questions:
1. Should I bump the required version of Roslyn (4.2 will work correctly with .400 SDK)?
2. Should I bump the `Microsoft.Data.SqlClient` to v5 or leave it at v4?

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -1,17 +1,17 @@
 <Project>
   <PropertyGroup>
-    <EntityFrameworkCoreVersion>6.0.3</EntityFrameworkCoreVersion>
-    <AspNetCoreVersion>6.0.3</AspNetCoreVersion>
+    <EntityFrameworkCoreVersion>6.0.10</EntityFrameworkCoreVersion>
+    <AspNetCoreVersion>6.0.10</AspNetCoreVersion>
     <ExtensionsVersion>6.0.0</ExtensionsVersion>
 
     <GoogleApisVersion>1.57.0</GoogleApisVersion>
-    <HangfireVersion>1.7.28</HangfireVersion>
+    <HangfireVersion>1.7.31</HangfireVersion>
     <IdentityServerVersion>4.1.2</IdentityServerVersion>
-    <SystemIdentityModelVersion>6.16.0</SystemIdentityModelVersion>
+    <SystemIdentityModelVersion>6.24.0</SystemIdentityModelVersion>
 
     <!-- Do not bump these dependencies if you don't want to force users to use newer .NET Core SDK -->
     <!-- Keep the major.minor values at exactly the one listed here: https://github.com/dotnet/roslyn/blob/main/docs/wiki/NuGet-packages.md -->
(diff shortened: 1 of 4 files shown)


Directory.Build.targets
```diff
@@ -24,44 +24,44 @@
...
-    <PackageReference Update="Sendgrid" Version="9.27.0" />
+    <PackageReference Update="Sendgrid" Version="9.28.1" />
   </ItemGroup>
 
   <ItemGroup>
     <PackageReference Update="Microsoft.AspNetCore.Authentication.JwtBearer" Version="$(AspNetCoreVersion)" />
-    <PackageReference Update="Microsoft.AspNetCore.Identity.EntityFrameworkCore" Version="$(AspNetCoreVersion)" />
     <PackageReference Update="Microsoft.AspNetCore.Mvc.Testing" Version="$(AspNetCoreVersion)" />
     <PackageReference Update="Microsoft.AspNetCore.Razor.Language" Version="$(AspNetCoreVersion)" />
+    <PackageReference Update="Microsoft.AspNetCore.Identity.EntityFrameworkCore" Version="$(AspNetCoreVersion)" />
```
Saancreed: Why is this one our of order now?

Saancreed (approved):
> Should I bump the required version of Roslyn (4.2 will work correctly with .400 SDK)?

I'd say yes, but let's make sure our SDK images are up to date first.

> Should I bump the `Microsoft.Data.SqlClient` to v5 or leave it at v4?

If it doesn't break anything for us (especially `Microsoft.EntityFrameworkCore.SqlServer`) then sure, why not.

Directory.Build.targets
jakubfijalkowski: Gah, moved it accidentally. Will rollback.

jakubfijalkowski:
> If it doesn't break anything for us (especially Microsoft.EntityFrameworkCore.SqlServer) then sure, why not.

That's the question that is hard to answer - EF, in theory, supports it. The minimal integration tests that we have here also work correctly.

test/LeanCode.IntegrationTests/docker/docker-compose.yml
```diff
@@ -29,14 +29,9 @@ services:
 
   #### Infrastructure
   db:
-    image: microsoft/mssql-server-linux:latest
+    image: mcr.microsoft.com/mssql/server:2022-latest
     environment:
       - ACCEPT_EULA=Y
       - SA_PASSWORD=Passw12#
```
Saancreed, line 35: Nit, but…

> The `SA_PASSWORD` environment variable is deprecated. Please use `MSSQL_SA_PASSWORD` instead.

https://learn.microsoft.com/en-us/sql/linux/sql-server-linux-configure-environment-variables?view=sql-server-linux-ver16#environment-variables

Saancreed (approved):
> That's the question that is hard to answer - EF, in theory, supports it. The minimal integration tests that we have here also work correctly.

:shipit:

PR #425, opened 2022-11-10
Remove `AsyncTasks`

### LINKED ISSUE #416: Get rid of `LeanCode.AsyncTasks`?
It was mainly for two things
- scheduling some work for the future 
- long running background tasks

For both cases we tend to use other means now, MassTransit scheduling and (more recently) MassTransit job consumers that better integrate with the rest of our ecosystem.

jakubfijalkowski:
Closes #416


marcin-dardzinski (approved):
Pls also remove leftovers from `Directory.Build.targets`

jakubfijalkowski:
@marcin-dardzinski which one? Hangfire is used by the integration tests also.

marcin-dardzinski:
Hmm, so you kept hangfire initializer for tests? :thinking:
Don't see a reason why we should keep it there, I'd say wipe it there as well.

jakubfijalkowski:
I kept it as an example on how to do it - in the end, there are still some projects that use Hangfire.

But maybe you're right. We haven't used it for a long time.

PR #430, opened 2022-11-17
`FindAndEnsureExistsAsync` exception

jakubfijalkowski:


src/Domain/LeanCode.DomainModels/DataAccess/IRepositoryExtensions.cs
```diff
@@ -15,7 +12,20 @@ public static class IRepositoryExtensions
             where TIdentity : notnull
         {
             var entity = await repository.FindAsync(id, cancellationToken);
-            return entity ?? throw new ArgumentException($"Aggregate of type: {typeof(TEntity).Name} with id: {id} does not exist");
+            return entity ?? throw new EntityDoesNotExistException(typeof(TEntity), id.ToString());
+        }
+    }
+
+    public class EntityDoesNotExistException : ArgumentNullException
```
Saancreed: I'd say it should derive from either `ArgumentException` or `InvalidOperationException`. Deriving from `ArgumentNullException` doesn't make much sense when `TIdentity` is not `null` and our entity it not some method argument.

jakubfijalkowski: Dammit. I wanted to use the `ArgumentException` but I must have selected wrong item from the list...

Saancreed (approved):

PR #448, opened 2023-01-28
Remove async methods in `IRepository` and `EFRepository`

### LINKED ISSUE #446: Cleanup `IRepository`
`IRepository` currently has two set of "changing" methods - sync and async - that work differently:
1. Sync methods just trigger `DateModified` update,
2. Async methods commit DB transaction.

This is very unintuitive and allows you to shoot oneself in the foot very easily. Since most of the time transactions are done using pipeline elements, then we should remove the async methods altogether.

lukaszgarstecki:
Closes #446

### DIFF
--- a/src/Domain/LeanCode.DomainModels.EF/EFRepository.cs
+++ b/src/Domain/LeanCode.DomainModels.EF/EFRepository.cs
@@ -60,9 +60,6 @@ namespace LeanCode.DomainModels.EF
         }
 
         public abstract Task<TEntity?> FindAsync(TIdentity id, CancellationToken cancellationToken = default);
-
-        protected Task SaveChangesAsync(CancellationToken cancellationToken = default) =>
-            DbContext.SaveChangesAsync(cancellationToken);
     }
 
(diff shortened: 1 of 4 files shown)


src/Domain/LeanCode.DomainModels.EF/EFRepository.cs
```diff
@@ -63,32 +59,6 @@ public virtual void Update(TEntity entity)
...
-
-        public virtual Task UpdateAsync(TEntity entity, CancellationToken cancellationToken = default)
-        {
-            Update(entity);
-            return SaveChangesAsync(cancellationToken);
-        }
-
         public abstract Task<TEntity?> FindAsync(TIdentity id, CancellationToken cancellationToken = default);
 
         protected Task SaveChangesAsync(CancellationToken cancellationToken = default) =>
```
jakubfijalkowski: This is unused now.

jakubfijalkowski (changes requested):

jakubfijalkowski (approved):

PR #449, opened 2023-02-02
Remove IRepository<T>

### LINKED ISSUE #264: Autofac fails to resolve generic interface
Since we use contravariant registration, Autofac fails to resolve IRepository<Aggregate, CustomType>, because there are two types registered here https://github.com/leancodepl/corelibrary/blob/v5.1/src/Domain/LeanCode.DomainModels/DataAccess/IRepository.cs

### LINKED ISSUE #375: Remove `IRepository<T>` so that client can decide which id type is the default one for the project
https://github.com/leancodepl/corelibrary/blob/3ad8ead897f01894842c939f5c801e2ec7f7dda3/src/Domain/LeanCode.DomainModels/DataAccess/IRepository.cs

mateuszkulak:


src/Domain/LeanCode.DomainModels.EF/EFRepository.cs
```diff
@@ -63,7 +63,7 @@ public virtual void Update(TEntity entity)
     }
 
     public abstract class EFRepository<TEntity, TContext>
```
jakubfijalkowski: We also should get rid of this one.

jakubfijalkowski (approved):

PR #450, opened 2023-02-02
Change the EFRepository update call to not force an update of the entire entity

### LINKED ISSUE #409: Don't force `dbContext.Update` in `EFRepository`
Currently [`EFRepository.Update`](https://github.com/leancodepl/corelibrary/blob/d4e10dce69d1c327b75770488b71b88c4c99943a/src/Domain/LeanCode.DomainModels.EF/EFRepository.cs#L63) method calls a `dbContext.Update` method as a part of it's procedure. 
IMO, this is an overkill, it forces all the properties of an entity to be written to the database, while it could send only those that really changed. 
I suggest we make this method only bump `DateModified` for `IOptimisiticConcurrency` instances and let EF do it's job.

mateuszkulak:


jakubfijalkowski:
Can we rename the PR to describe what will be achieved by the change (so no forced update of whole entity), and not what the code change is?

jakubfijalkowski (approved):

PR #451, opened 2023-02-03
Prevent race conditions in FixedTimeProvider

### LINKED ISSUE #421: Race conditions in `FixedTimeProvider`
I have a case when I need to enforce that some time is monotonically increasing. In my unit tests, some tests use `FixedTimeProvider` but some don't care for the time as long as it's increasing. Some tests sometimes fail because the time goes backwards. My suspicion in that because the test run in parallel, some other tests use `FixedTimeProvider.SetTo` and messes with time (the default async local date is a zero value)

If `FixedTimeProvider` doesn't find a date in async local storage it should return current now and not zero date.
Plus, I'd consider adding `Func<DateTime>` overload for more flexibility, e.g. I had a case when I wanted the date to be bumped per call.

mateuszkulak:


jakubfijalkowski (changes requested):
Let's fix formatting the way GH Actions says and let's add tests.

test/Domain/LeanCode.TimeProvider.Tests/FixedTimeProviderTests.cs
```diff
@@ -83,5 +83,46 @@ public void Attempt_to_assign_local_date_time_throws()
...
+            Assert.True(fixedTime == expectedTime || differenceInSeconds < 1);
+        }
+
+        [Fact]
+        public void Check_if_date_with_offset_is_correct_if_you_set_value()
+        {
+            var fixedTime = new DateTimeOffset(2023, 1, 14, 12, 1, 15, TimeSpan.FromHours(2));
+            FixedTimeProvider.SetTo(fixedTime);
+
+            Assert.Equal(fixedTime, TimeProvider.Now);
```
jakubfijalkowski:
```suggestion
            Assert.Equal(fixedTime, TimeProvider.NowWithOffset);
```

test/Domain/LeanCode.TimeProvider.Tests/FixedTimeProviderTests.cs
```diff
@@ -83,5 +83,46 @@ public void Attempt_to_assign_local_date_time_throws()
...
+        }
+
+        [Fact]
+        public void Check_if_date_is_correct_if_you_have_not_set_value()
+        {
+            var fixedTime = TimeProvider.Now;
+            var expectedTime = DateTime.UtcNow;
+
+            int differenceInSeconds = (int)(fixedTime - expectedTime).TotalSeconds;
```
jakubfijalkowski: The cast is unnecessary.

jakubfijalkowski (changes requested):

test/Domain/LeanCode.TimeProvider.Tests/FixedTimeProviderTests.cs
```diff
@@ -83,5 +83,46 @@ public void Attempt_to_assign_local_date_time_throws()
...
+        public void Check_if_date_with_offset_is_correct_if_you_have_not_set_value()
+        {
+            var fixedTime = TimeProvider.NowWithOffset;
+            var expectedTime = DateTimeOffset.UtcNow;
+
+            var differenceInSeconds = (fixedTime - expectedTime).TotalSeconds;
+
+            Assert.True(fixedTime == expectedTime || differenceInSeconds < 1);
+        }
+
```
jakubfijalkowski: Unnecessary newline.

jakubfijalkowski (approved):
Last nitpick :)

PR #453, opened 2023-02-03
Add query responses in logs

mateuszkulak:


src/Domain/LeanCode.CQRS.Execution/QueryLoggerElement.cs
```diff
@@ -0,0 +1,22 @@
...
+        where AppContext : notnull, IPipelineContext
+    {
+        public readonly Serilog.ILogger logger = Serilog.Log.ForContext<QueryLoggerElement<AppContext, TObj, TRes>>();
+
+        public Task<TRes> ExecuteAsync(
+            AppContext ctx,
+            TObj input,
+            Func<AppContext, TObj, Task<TRes>> next)
+        {
+            logger.Information("Query executed with response {@Response}", input);
```
jakubfijalkowski: If you don't limit the `TObj` to be a `IQuery`, then this can be used for objects different than queries (e.g. operations, or even commands).

This is a very good approach, but let's not use `query` in names in that case.

src/Domain/LeanCode.CQRS.Execution/LeanCode.CQRS.Execution.csproj
```diff
@@ -7,6 +7,7 @@
   <ItemGroup>
     <PackageReference Include="LeanCode.Contracts" />
     <PackageReference Include="Serilog" />
+    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="7.0.0" />
```
jakubfijalkowski: The version is already pinned [here](https://github.com/leancodepl/corelibrary/blob/v7.0/Directory.Build.targets#L84-L85).
```suggestion
    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" />
```

src/Domain/LeanCode.CQRS.Execution/DevResponseLoggerElement.cs
```diff
@@ -0,0 +1,34 @@
...
+
+namespace LeanCode.CQRS.Execution
+{
+    public class DevResponseLoggerElement<AppContext, TObj, TRes> : IPipelineElement<AppContext, TObj, TRes>
+        where AppContext : notnull, IPipelineContext
+    {
+        public readonly Serilog.ILogger logger = Serilog.Log.ForContext<DevResponseLoggerElement<AppContext, TObj, TRes>>();
+        public readonly IHostEnvironment environment;
+
+        DevResponseLoggerElement(IHostEnvironment env)
```
jakubfijalkowski:
```suggestion
        public DevResponseLoggerElement(IHostEnvironment env)
```

src/Domain/LeanCode.CQRS.Execution/DevResponseLoggerElement.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            logger = Serilog.Log.ForContext<DevResponseLoggerElement<AppContext, TObj, TRes>>();
```
jakubfijalkowski:
```suggestion
```

You already initializing it in a field initializer.

You should have a separate ctor, that takes both `IHostEnvironment` and `ILogger` for tests only - but that one should be separate from the one that should be used most of the time.

src/Domain/LeanCode.CQRS.Execution/DevResponseLoggerElement.cs
```diff
@@ -0,0 +1,34 @@
...
+            environment = env;
+        }
+
+        public Task<TRes> ExecuteAsync(
+            AppContext ctx,
+            TObj input,
+            Func<AppContext, TObj, Task<TRes>> next)
+        {
+            if (environment.IsDevelopment())
```
jakubfijalkowski:
```suggestion
            if (!environment.IsProduction())
```
Maybe? I'm not sure - we probably should ask on our internal Slack.

src/Domain/LeanCode.CQRS.Execution/DevResponseLoggerElement.cs
```diff
@@ -1,21 +1,25 @@
...
 using Microsoft.Extensions.Hosting;
+using Serilog;
 
 namespace LeanCode.CQRS.Execution
 {
     public class DevResponseLoggerElement<AppContext, TObj, TRes> : IPipelineElement<AppContext, TObj, TRes>
         where AppContext : notnull, IPipelineContext
     {
-        public readonly Serilog.ILogger logger = Serilog.Log.ForContext<DevResponseLoggerElement<AppContext, TObj, TRes>>();
+        public readonly ILogger logger;
```
jakubfijalkowski: `private`

jakubfijalkowski: The one below should also be private.

src/Domain/LeanCode.CQRS.Execution/ResponseLoggerElement.cs
```diff
@@ -2,13 +2,24 @@
...
 using LeanCode.Pipelines;
+using Serilog;
 
 namespace LeanCode.CQRS.Execution
 {
     public class ResponseLoggerElement<AppContext, TObj, TRes> : IPipelineElement<AppContext, TObj, TRes>
         where AppContext : notnull, IPipelineContext
     {
-        public readonly Serilog.ILogger logger = Serilog.Log.ForContext<ResponseLoggerElement<AppContext, TObj, TRes>>();
+        public readonly ILogger logger;
```
jakubfijalkowski: `private`

jakubfijalkowski (changes requested):

test/Domain/LeanCode.CQRS.Tests/Execution/QueryExecutionTests.cs
```diff
@@ -42,6 +46,30 @@ public async Task Correctly_routes_data_through_pipeline_elements()
             Assert.Equal(query, element.Data);
         }
 
+        [Fact]
+        public async Task Correctly_log_query_response_in_pipeline_element()
```
jakubfijalkowski: This is not the correct place to test this - you should have a separate test class for `ResponseLoggerElement` and `DevResponseLoggerElement` that tests just that. This class tests `QueryExecutionTests` and it has nothing to do with these two elements. Let's move that there. And also - we don't need whole pipeline here, we're just testing the single element so we can test it directly.

jakubfijalkowski: Same applies to the `OperationExecutionTests`.

jakubfijalkowski (changes requested):

test/Domain/LeanCode.CQRS.Tests/Execution/OperationExecutionTests.cs
```diff
@@ -1,4 +1,3 @@
-using System.Threading.Tasks;
```
jakubfijalkowski: Let's revert this change.

test/Domain/LeanCode.CQRS.Tests/Execution/QueryExecutionTests.cs
```diff
@@ -1,4 +1,3 @@
-using System.Threading.Tasks;
```
jakubfijalkowski: And revert this one.

test/Domain/LeanCode.CQRS.Tests/BaseCQRSTests.cs
```diff
@@ -42,6 +42,8 @@ public class BaseCQRSTests
             builder.RegisterType<SamplePipelineElement<ICommand, CommandResult>>().AsSelf().SingleInstance();
             builder.RegisterType<SamplePipelineElement<IQuery, object>>().AsSelf().SingleInstance();
             builder.RegisterType<SamplePipelineElement<IOperation, object>>().AsSelf().SingleInstance();
+            builder.RegisterType<ResponseLoggerElement<AppContext, object, object>>().AsSelf().SingleInstance();
```
jakubfijalkowski: This is not needed.

test/Domain/LeanCode.CQRS.Tests/LoggingResponses/LoggingOperationResponsesTests.cs
```diff
@@ -0,0 +1,36 @@
...
+
+            var operationResponse = "Success!";
+            var nextFunctionResult = "Completed!";
+
+            var logger = new LoggerConfiguration().MinimumLevel
+                .Is(LogEventLevel.Verbose)
+                .WriteTo.Sink(fakeSink, LogEventLevel.Verbose)
+                .CreateLogger();
+
+            var element = Container.Resolve<ResponseLoggerElement<AppContext, object, object>>(new NamedParameter("logger", logger));
```
jakubfijalkowski: You don't need the container. Just `new` the `ResponseLoggerElement`.

test/Domain/LeanCode.CQRS.Tests/LoggingResponses/LoggingOperationResponsesTests.cs
```diff
@@ -0,0 +1,36 @@
...
+        [Fact]
+        public async Task Correctly_log_operation_response_and_run_next()
+        {
+            Prepare();
+
+            var appCtx = new AppContext();
+            var fakeSink = new FakeSink();
+
+            var operationResponse = "Success!";
+            var nextFunctionResult = "Completed!";
```
jakubfijalkowski:
```suggestion
            const OperationResponse = "Success!";
            const NextFunctionResult = "Completed!";
```

plus let's move it to the class level.

src/Domain/LeanCode.CQRS.Execution/ResponseLoggerElement.cs
```diff
@@ -0,0 +1,31 @@
+
```
jakubfijalkowski: Remember to fix the warnings in this file and in the `DevResponseLoggerElement`.

src/Domain/LeanCode.CQRS.Execution/DevResponseLoggerElement.cs
```diff
@@ -0,0 +1,38 @@
...
+        }
+
+        public Task<TRes> ExecuteAsync(
+            AppContext ctx,
+            TObj input,
+            Func<AppContext, TObj, Task<TRes>> next)
+        {
+            if (!environment.IsProduction())
+            {
+                logger.Information("Request executed with response {@Response}", input);
```
jakubfijalkowski: You're logging the wrong thing - you should not log the input, but the response.

test/Domain/LeanCode.CQRS.Tests/LoggingResponses/LoggingOperationResponsesTests.cs
```diff
@@ -0,0 +1,36 @@
...
+
+            var logger = new LoggerConfiguration().MinimumLevel
+                .Is(LogEventLevel.Verbose)
+                .WriteTo.Sink(fakeSink, LogEventLevel.Verbose)
+                .CreateLogger();
+
+            var element = Container.Resolve<ResponseLoggerElement<AppContext, object, object>>(new NamedParameter("logger", logger));
+
+            var result = await element.ExecuteAsync(appCtx, operationResponse, new Func<AppContext, object, Task<object>>((_, _) => Task.FromResult<object>(nextFunctionResult)));
```
jakubfijalkowski:
```suggestion
            var result = await element.ExecuteAsync(appCtx, operationResponse, (ctx, obj) => Task.FromResult<object>(nextFunctionResult));
```

test/Domain/LeanCode.CQRS.Tests/LoggingResponses/LoggingOperationResponsesTests.cs
```diff
@@ -0,0 +1,36 @@
...
+
+            Assert.Equal(1, fakeSink.Messages.Count());
```
jakubfijalkowski:
```suggestion
            Assert.Single(fakeSink.Messages);
```

test/Domain/LeanCode.CQRS.Tests/LoggingResponses/LoggingOperationResponsesTests.cs
```diff
@@ -0,0 +1,36 @@
...
+            Assert.Equal("Request executed with response \"Success!\"", fakeSink.Messages[0]);
```
jakubfijalkowski: Don't check for the whole message, it does not matter. Check if the message contains what you expect (so the result of executing `next`).

Also, the test checks not what should be checked :P

jakubfijalkowski (changes requested):

test/Domain/LeanCode.CQRS.Tests/Execution/OperationExecutionTests.cs
mateuszkulak: VS Code says that this is unnecessary using directive and all tests are working without that so I dont understand why we need that

jakubfijalkowski: Because that is not the topic of this PR.

src/Domain/LeanCode.CQRS.Execution/DevResponseLoggerElement.cs
mateuszkulak: As I understand it, in this situation input is the response

jakubfijalkowski: The query handler will be executed when called `next`, not _before_. Elements are run _before_ or _after_ execution, where `before` is before calling `next`, and `after` is after calling `next`. `next` has the result, `input` is _the input_, so a query object.

test/Domain/LeanCode.CQRS.Tests/LoggingResponses/LoggingOperationResponsesTests.cs
```diff
@@ -0,0 +1,83 @@
...
+namespace LeanCode.CQRS.Tests
+{
+    public class LoggingOperationResponsesTests : BaseCQRSTests
+    {
+        const string OperationResponse = "Success!";
+
+        [Fact]
+        public async Task Correctly_log_operation_response_and_run_next()
+        {
+            Prepare();
```
jakubfijalkowski: Not needed.

test/Domain/LeanCode.CQRS.Tests/LoggingResponses/LoggingOperationResponsesTests.cs
```diff
@@ -0,0 +1,83 @@
+using LeanCode.CQRS.Execution;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Hosting.Internal;
+using Serilog;
+using Serilog.Events;
+using Xunit;
+
+namespace LeanCode.CQRS.Tests
+{
+    public class LoggingOperationResponsesTests : BaseCQRSTests
```
jakubfijalkowski:
```suggestion
    public class LoggingOperationResponsesTests
```

test/Domain/LeanCode.CQRS.Tests/LoggingResponses/LoggingOperationResponsesTests.cs
```diff
@@ -0,0 +1,83 @@
...
+
+            var result = await element.ExecuteAsync(appCtx, OperationResponse, (ctx, obj) => Task.FromResult<object>(obj));
+
+            Assert.Single(fakeSink.Messages);
+            Assert.Equal(OperationResponse, result);
+            Assert.Contains(result.ToString(), fakeSink.Messages[0]);
+        }
+
+        [Fact]
+        public async Task Does_not_log_operation_response_and_run_next_on_prod_env()
```
jakubfijalkowski: These are two tests - one that checks if the message is logged, one that checks if `next` is called.

src/Domain/LeanCode.CQRS.Execution/DevResponseLoggerElement.cs
```diff
@@ -0,0 +1,38 @@
+using LeanCode.Pipelines;
```
jakubfijalkowski: There are still some warnings left (esp. regarding the generic parameter names).

test/Domain/LeanCode.CQRS.Tests/BaseCQRSTests.cs
jakubfijalkowski: It's not resolved yet.

jakubfijalkowski (changes requested):

jakubfijalkowski:
Okay, very close now. TODO:
1. Helpers to register the elements, like this: https://github.com/leancodepl/corelibrary/blob/v7.0/src/Domain/LeanCode.CQRS.Security/CQRSSecurityElement.cs#L61-L82
2. Squash the commits, so that we the git history looks less cluttered.

src/Domain/LeanCode.CQRS.Execution/NonProductionResponseLoggerElement.cs
```diff
@@ -0,0 +1,50 @@
...
+                logger.Information("Request executed with response {@Response}", result);
+            }
+
+            return result;
+        }
+    }
+
+    public static class LogResponsesnOnNonProductionPipelineBuilderExtensions
+    {
+        public static PipelineBuilder<TContext, TObj, TRes> LogRespones<TContext, TObj, TRes>(
```
jakubfijalkowski: Let's rename this to sth like `LogResponesOnNonProduction` otherwise it clashes with the other `LogResponses`.

jakubfijalkowski (approved):

PR #454, opened 2023-02-06
Make dateonly and timeonly json converters lax by deafault

mateuszkulak:


test/Infrastructure/LeanCode.Serialization.Tests/JsonConvertersTests.cs
```diff
@@ -35,6 +27,31 @@ public void Throws_when_attempting_to_deserialize_DateOnly_with_time_part(string
...
+        string serialized, int hour, int minute, int second, int millisecond, string reserialized)
+    {
+        var time = new TimeOnly(hour, minute, second, millisecond);
+
+        Assert.Equal(time, JsonSerializer.Deserialize<TimeOnly>(serialized, LaxSerializerOptions));
+        Assert.Equal(reserialized, JsonSerializer.Serialize(time, LaxSerializerOptions));
+    }
+
+    [Theory]
+    [InlineData("\"2021-12-15T07:23:10.115Z\"", 2021, 12, 15, 7, 23, 10, 115,1, "\"2021-12-15T07:23:10.115Z\"")]
```
jakubfijalkowski: Let's add more cases that handle the "lax" part.

Also
```suggestion
    [InlineData("\"2021-12-15T07:23:10.115Z\"", 2021, 12, 15, 7, 23, 10, 115, 1, "\"2021-12-15T07:23:10.115Z\"")]
```

test/Infrastructure/LeanCode.Serialization.Tests/JsonConvertersTests.cs
```diff
@@ -35,6 +27,31 @@ public void Throws_when_attempting_to_deserialize_DateOnly_with_time_part(string
...
+    [Theory]
+    [InlineData("\"2021-12-15T07:23:10.115\"", 7, 23, 10, 115, "\"07:23:10.115\"")]
+    public void Correctly_deserializes_TimeOnly_with_date_part_when_using_lax_converter(
+        string serialized, int hour, int minute, int second, int millisecond, string reserialized)
+    {
+        var time = new TimeOnly(hour, minute, second, millisecond);
+
+        Assert.Equal(time, JsonSerializer.Deserialize<TimeOnly>(serialized, LaxSerializerOptions));
+        Assert.Equal(reserialized, JsonSerializer.Serialize(time, LaxSerializerOptions));
+    }
```
jakubfijalkowski: Small hint,
```suggestion
    [InlineData("2021-12-15T07:23:10.115", 7, 23, 10, 115, "07:23:10.115")]
    public void Correctly_deserializes_TimeOnly_with_date_part_when_using_lax_converter(
        string serialized, int hour, int minute, int second, int millisecond, string reserialized)
    {
        var time = new TimeOnly(hour, minute, second, millisecond);

        Assert.Equal(time, JsonSerializer.Deserialize<TimeOnly>(@"\"{serialized}\"", LaxSerializerOptions));
        Assert.Equal(@"\"reserialized\"", JsonSerializer.Serialize(time, LaxSerializerOptions));
    }
```

Reads better :)

test/Infrastructure/LeanCode.Serialization.Tests/JsonConvertersTests.cs
```diff
@@ -35,6 +27,31 @@ public void Throws_when_attempting_to_deserialize_DateOnly_with_time_part(string
         Assert.Equal(reserialized, JsonSerializer.Serialize(date, LaxSerializerOptions));
     }
 
+    [Theory]
```
jakubfijalkowski: What about failure case? We need tests for that also.

jakubfijalkowski (changes requested):
A couple of comments, plus let's fix all the styling errors reported in the code.

src/Infrastructure/LeanCode.Serialization/JsonConverters/JsonLaxDateOnlyConverter.cs
```diff
@@ -1,16 +1,26 @@
...
 public class JsonLaxDateOnlyConverter : JsonConverter<DateOnly>
 {
+    private static readonly UTF8Encoding encoder = new(false);
+
     public override DateOnly Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
     {
-        return reader.GetString() is string s
-            ? DateOnly.FromDateTime(DateTime.Parse(s, CultureInfo.InvariantCulture))
-            : default;
+        var value = reader.ValueSpan;
```
Saancreed: What if the reader was constructed from `ReadOnlySequence<byte>` instead? You must check if `HasValueSequence` is false before accessing `ValueSpan` property (and if it's not, use `ValueSequence` property instead).

But actually, you probably want to use `CopyString` instead, see the comment about (un)escaping below.

src/Infrastructure/LeanCode.Serialization/JsonConverters/JsonLaxDateOnlyConverter.cs
```diff
@@ -1,16 +1,26 @@
...
+        Span<char> chars = stackalloc char[value.Length];
```
Saancreed: Ensure that input is no longer than some reasonable number of bytes. If it is, allocate your span on the heap using `new` or just throw an exception outright.

src/Infrastructure/LeanCode.Serialization/JsonConverters/JsonLaxDateOnlyConverter.cs
```diff
@@ -1,16 +1,26 @@
...
+        encoder.GetChars(value, chars);
+        if (DateTime.TryParse(chars ,out DateTime date))
```
Saancreed: Won't this attempt to parse a potentially escaped raw JSON value? Please make sure that the following succeeds:

```csharp
JsonSerializer.Deserialize<DateOnly>(@"""\u0032\u0030\u0032\u0032-\u0030\u0032-\u0030\u0032""")
```

.NET's builtin converter properly recognizes this as `2022-02-02`. please make sure ours does as well.

src/Infrastructure/LeanCode.Serialization/JsonConverters/JsonLaxDateOnlyConverter.cs
```diff
@@ -1,16 +1,26 @@
...
+        {
+            return new DateOnly(date.Year, date.Month, date.Day);
```
Saancreed:
```suggestion
            return DateOnly.FromDateTime(date);
```

src/Infrastructure/LeanCode.Serialization/JsonConverters/JsonLaxDateTimeOffsetConverter.cs
```diff
@@ -0,0 +1,29 @@
...
+        {
+            var offset = new DateTimeOffset(date).Offset;
+            return new DateTimeOffset(date.Ticks -  offset.Ticks, offset);
+        }
+
+        throw new JsonException("Invalid format");
+    }
+
+    public override void Write(Utf8JsonWriter writer, DateTimeOffset value, JsonSerializerOptions options) =>
+        writer.WriteStringValue(value.ToString(format, CultureInfo.InvariantCulture));
```
Saancreed: This will allocate a string, write to `stackalloc`'d `Span<char>` instead.

src/Infrastructure/LeanCode.Serialization/JsonConverters/JsonLaxDateTimeOffsetConverter.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Text;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace LeanCode.Serialization;
+
+public class JsonLaxDateTimeOffsetConverter : JsonConverter<DateTimeOffset>
+{
+    private static readonly UTF8Encoding encoder = new(false);
+    private readonly string format = "yyyy'-'MM'-'dd'T'HH':'mm':'ss.fffZ";
```
Saancreed: Is that `Z` a custom format specifier or a literal? Because neither seems correct for me here (but I don't exactly remember our requirements on how the offset can be specified).

Please make sure that a string like `2023-02-06T21:37:42.666+01:00` can be both written and read correctly.

src/Infrastructure/LeanCode.Serialization/JsonConverters/JsonLaxDateTimeOffsetConverter.cs
```diff
@@ -0,0 +1,29 @@
...
+    public override DateTimeOffset Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+    {
+        var value = reader.ValueSpan;
+        Span<char> chars = stackalloc char[value.Length];
+        encoder.GetChars(value, chars);
+        if (DateTime.TryParse(chars, out DateTime date))
```
Saancreed: Aren't you losing the offset data by parsing with `DateTime` instead of `DateTimeOffset` directly?

src/Infrastructure/LeanCode.Serialization/JsonConverters/JsonLaxDateOnlyConverter.cs
```diff
@@ -1,16 +1,26 @@
...
+        }
+
+        throw new JsonException("Invalid date format");
     }
 
     public override void Write(Utf8JsonWriter writer, DateOnly value, JsonSerializerOptions options) =>
```
Saancreed: While you're at it, let's update this method to write to a span instead (see my comment in the other converter).

src/Infrastructure/LeanCode.Serialization/JsonConverters/JsonLaxTimeOnlyConverter.cs
```diff
@@ -0,0 +1,28 @@
...
+public class JsonLaxTimeOnlyConverter : JsonConverter<TimeOnly>
+{
+    private static readonly UTF8Encoding encoder = new(false);
+    private readonly string format = "HH:mm:ss.fff";
+    public override TimeOnly Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+    {
+        var value = reader.ValueSpan;
+        Span<char> chars = stackalloc char[value.Length];
+        encoder.GetChars(value, chars);
+        if (DateTime.TryParse(chars, out DateTime time))
```
Saancreed: I think that's far too lax. Is `TimeOnly.TryParse` not good enough?

src/Infrastructure/LeanCode.Serialization/JsonConverters/JsonLaxTimeOnlyConverter.cs
```diff
@@ -0,0 +1,28 @@
...
+        {
+            return new TimeOnly(time.Hour, time.Minute, time.Second, time.Millisecond);
+        }
+
+        throw new JsonException("Invalid format");
+    }
+
+    public override void Write(Utf8JsonWriter writer, TimeOnly value, JsonSerializerOptions options) =>
```
Saancreed: Also here, write to a span first.

Saancreed (commented):
Note that comments related to reading the value in `JsonLaxDateOnlyConverter` also apply to other converters.

src/Infrastructure/LeanCode.Serialization/JsonConverters/JsonLaxDateOnlyConverter.cs
```diff
@@ -1,18 +1,47 @@
...
     {
-        return reader.GetString() is string s
-            ? DateOnly.FromDateTime(DateTime.Parse(s, CultureInfo.InvariantCulture))
-            : default;
+        Span<char> source = stackalloc char[0];
+
+        if(reader.HasValueSequence)
+        {
+            var valueLength = (int)reader.ValueSequence.Length;
+            Span<char> buffer = ArrayPool<char>.Shared.Rent(valueLength);
```
Saancreed: You must return memory you rent from the array pool.

It would be easier if you just ensured the length does not exceed some reasonable value and always stackalloc'd the buffer. Try the following instead (pick one `#if` path):

```csharp
var length = reader.HasValueSequence ? (int)reader.ValueSeqence.Length : reader.ValueSpan.Length;

#if THROW_EXCEPTION
if (length > MaxBufferSize)
{
    throw new JsonException($"Input too long ({length})");
}

Span<char> buffer = stackalloc char[length];
#else
Span<char> buffer = length > MaxBufferSize ? new char[length] : stackalloc char[length];
#endif
var count = reader.CopyString(buffer);

if (DateTime.TryParse(buffer[..count], out var dt))
{
    return DateOnly.FromDateTime(dt);
}
else
{
    throw new JsonException("Failed to parse input, invalid format.");
}
```

src/Infrastructure/LeanCode.Serialization/JsonConverters/JsonLaxDateOnlyConverter.cs
```diff
@@ -1,18 +1,47 @@
...
+            int charsRead = reader.CopyString(buffer);
+            source = buffer.Slice(0, charsRead);
+        }
+        else
+        {
+            var value = reader.ValueSpan;
+            source  = stackalloc char[value.Length];
+            encoder.GetChars(value, source);
```
Saancreed: This still doesn't do the unescaping part, does it? You should probably use `CopyString` in both scenarios.

src/Infrastructure/LeanCode.Serialization/JsonConverters/JsonLaxDateTimeOffsetConverter.cs
```diff
@@ -0,0 +1,47 @@
...
+        else
+        {
+            var value = reader.ValueSpan;
+            source  = stackalloc char[value.Length];
+            encoder.GetChars(value, source);
+        }
+
+        if (DateTimeOffset.TryParse(source, CultureInfo.InvariantCulture, out DateTimeOffset date))
+        {
+            return new DateTimeOffset(date.Ticks, date.Offset);
```
Saancreed: Won't this recreate the value equal to `date`? If you want to round to milliseconds, `ticks` are too precise for that (they are 100ns each).

src/Infrastructure/LeanCode.Serialization/JsonConverters/JsonLaxDateOnlyConverter.cs
```diff
@@ -1,18 +1,47 @@
...
 using System.Text.Json;
 using System.Text.Json.Serialization;
 
 namespace LeanCode.Serialization;
 
 public class JsonLaxDateOnlyConverter : JsonConverter<DateOnly>
 {
+    private static readonly UTF8Encoding encoder = new(false);
+    private const int maxBufferSize = 32;
+    private readonly string format = "yyyy-MM-dd";
```
Saancreed: Isn't this the same thing as `o` standard format? It _might_ be a little faster than using a custom one.

src/Infrastructure/LeanCode.Serialization/JsonConverters/JsonLaxDateOnlyConverter.cs
```diff
@@ -1,18 +1,47 @@
...
+        }
+
+        throw new JsonException("Invalid date format");
+
     }
 
-    public override void Write(Utf8JsonWriter writer, DateOnly value, JsonSerializerOptions options) =>
-        writer.WriteStringValue(value.ToString("o", CultureInfo.InvariantCulture));
+    public override void Write(Utf8JsonWriter writer, DateOnly value, JsonSerializerOptions options) {
+        Span<char> buffer = stackalloc char[maxBufferSize];
```
Saancreed: In case of `DateOnly`, you could get away with just 10 chars when writing (but feel free to round to 16).

test/Infrastructure/LeanCode.Serialization.Tests/JsonConvertersTests.cs
```diff
@@ -35,6 +27,39 @@ public void Throws_when_attempting_to_deserialize_DateOnly_with_time_part(string
...
+        string serialized, int year, int month, int day, int hour, int minute, int second, int millisecond, int timeSpanOffset, string reserialized)
+    {
+        var date = new DateTimeOffset(year, month, day, hour, minute, second, millisecond, new TimeSpan(timeSpanOffset,0,0));
+
+        Assert.Equal(date, JsonSerializer.Deserialize<DateTimeOffset>(serialized, LaxSerializerOptions));
+        Assert.Equal(reserialized, JsonSerializer.Serialize(date, LaxSerializerOptions));
+    }
+
+    [Theory]
+    [InlineData("\"2021-12-15T07:23:10.115+03:00\"", 2021, 12, 15, 7, 23, 10, 115, 3, "\"2021-12-15T07:23:10.115Z\"")]
```
Saancreed: I believe this is wrong, you are losing the offset when serializing again. The reserialized value should be exactly the same as the serialized input.

src/Infrastructure/LeanCode.Serialization/JsonConverters/JsonLaxDateOnlyConverter.cs
```diff
@@ -1,18 +1,57 @@
...
+using System.Text;
+using System.Text.Encodings.Web;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 
 namespace LeanCode.Serialization;
 
 public class JsonLaxDateOnlyConverter : JsonConverter<DateOnly>
 {
+    private static readonly UTF8Encoding encoder = new(false);
```
Saancreed: Unused?

src/Infrastructure/LeanCode.Serialization/JsonConverters/JsonLaxDateOnlyConverter.cs
```diff
@@ -1,18 +1,57 @@
...
+    private const int maxDateOnlyBufferSize = 16;
+
     public override DateOnly Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
     {
-        return reader.GetString() is string s
-            ? DateOnly.FromDateTime(DateTime.Parse(s, CultureInfo.InvariantCulture))
-            : default;
+        Span<char> source = stackalloc char[0];
+
+        if(reader.HasValueSequence)
```
Saancreed: Everything except extracting the input length is shared between both branches of this `if`, please avoid repeating your code.

This should also let you avoid weird early declaration of `source` above.

src/Infrastructure/LeanCode.Serialization/JsonConverters/JsonLaxDateTimeOffsetConverter.cs
```diff
@@ -0,0 +1,67 @@
...
+            }
+            Span<char> buffer = stackalloc char[value.Length];
+            int charsRead = reader.CopyString(buffer);
+            source = buffer.Slice(0, charsRead);
+        }
+
+        if (DateTimeOffset.TryParse(source, CultureInfo.InvariantCulture, out DateTimeOffset date))
+        {
+            var roundedTicks = (date.Ticks + TimeSpan.TicksPerMillisecond / 2) / TimeSpan.TicksPerMillisecond * TimeSpan.TicksPerMillisecond;
+            return new DateTimeOffset(roundedTicks, date.Offset);
```
Saancreed:
```suggestion
            return date.AddTicks(-(date.Ticks % TimeSpan.TicksPerMillisecond));
```

I think we can afford a simpler solution because it's not defined how we are supposed to handle values that are too precise (and just truncating is by far the easiest way out).

src/Infrastructure/LeanCode.Serialization/JsonConverters/JsonLaxDateTimeOffsetConverter.cs
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        throw new JsonException("Invalid format");
+    }
+
+    public override void Write(Utf8JsonWriter writer, DateTimeOffset value, JsonSerializerOptions options) {
+        Span<char> buffer = stackalloc char[maxDateTimeOffsetBufferSize];
+        var success = value.TryFormat(buffer, out var written, format, CultureInfo.InvariantCulture);
+
+        if(value.Offset == TimeSpan.Zero)
```
Saancreed: I think it's okay to leave `+00:00` for UTC DTOs? I'll need to check that one though.
Also you don't need to null-terminate as long as you slice properly.

src/Infrastructure/LeanCode.Serialization/JsonConverters/JsonLaxDateOnlyConverter.cs
```diff
@@ -1,18 +1,57 @@
...
+using System.Text.Encodings.Web;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 
 namespace LeanCode.Serialization;
 
 public class JsonLaxDateOnlyConverter : JsonConverter<DateOnly>
 {
+    private static readonly UTF8Encoding encoder = new(false);
+    private const int maxBufferSize = 64;
```
Saancreed: I think this should be six times the length in chars of the longest valid value because every character could be escaped as `\u0123` sequence.

test/Infrastructure/LeanCode.Serialization.Tests/JsonConvertersTests.cs
```diff
@@ -51,7 +61,81 @@ public void Correctly_serializes_and_deserializes_TimeOnly(string serialized, in
...
+
         Assert.Equal(serialized, JsonSerializer.Serialize(time, DefaultSerializerOptions));
         Assert.Equal(time, JsonSerializer.Deserialize<TimeOnly>(serialized, DefaultSerializerOptions));
     }
+
+    [Theory]
+    [InlineData("\"213:37:42.123\"")]
+    public void Throws_trying_deserialize_incorrect_TimeOnly(string serialized)
+    {
+        Assert.Throws<JsonException>(() => JsonSerializer.Deserialize<DateOnly>(serialized, DefaultSerializerOptions));
```
Saancreed:
```suggestion
        Assert.Throws<JsonException>(() => JsonSerializer.Deserialize<TimeOnly>(serialized, DefaultSerializerOptions));
```

Saancreed (commented):
Top 3 comments in `JsonLaxDateOnlyConverter` apply to others as well.

Saancreed (approved):
One last thing: address the formatting issues as reported in tests (multiple empty lines, spaces after commas, etc.), otherwise LGTM.

jakubfijalkowski (approved):

jakubfijalkowski (approved):

PR #456, opened 2023-02-13
RFC: Ulid<T> and SUlid<T>

Saancreed:
That's pretty much how I imagine this to work. I'm not yet sure about the public API surface for `SUlid<T>` though (and it's _not_ consistent with `SId<T>` precisely so we can discuss a new iteration) so feel free to suggest changes.

Notable changes from existing types:

1. New structs are `record` types so I don't have to write some of equality boilerplate, at least for `Ulid<T>`.
2. EF converters have public constructors so they can be used in `HaveConversion` methods.
3. EF extensions are for DB-wide conventions instead of per-column configuration.

We could also consider allowing passing custom collation / column type because we don't really need anything more than a ordinal comparison but we can't set any reasonable defaults anyway. At best we could do something like `.UseSqlServerUtf8Column(string? collation)` which could expand to `.HaveColumnType("nchar(26)").UseCollation(collation ?? "Latin1_General_100_BIN2_UTF8").AreUnicode()` but that's gross.

I'll add some tests once we're sure we want this.

### DIFF
--- a/src/Domain/LeanCode.DomainModels.EF/PropertiesConfigurationBuilderExtensions.cs
+++ b/src/Domain/LeanCode.DomainModels.EF/PropertiesConfigurationBuilderExtensions.cs
@@ -1,0 +1,44 @@
+using LeanCode.DomainModels.Model;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
+
+namespace LeanCode.DomainModels.EF;
+
+public static class PropertiesConfigurationBuilderExtensions
+{
+    public static PropertiesConfigurationBuilder<SUlid<T>> AreTypedId<T>(this PropertiesConfigurationBuilder<SUlid<T>> builder)
+        where T : class, IIdentifiable<SUlid<T>>
+    {
+        return builder
+            .HaveConversion<SUlidConverter<T>>()
+            .HaveMaxLength(SUlid<T>.RawLength)
+            .AreFixedLength();
+    }
+
+    public static PropertiesConfigurationBuilder<SUlid<T>?> AreTypedId<T>(this PropertiesConfigurationBuilder<SUlid<T>?> builder)
+        where T : class, IIdentifiable<SUlid<T>>
+    {
+        return builder
+            .HaveConversion<SUlidConverter<T>>()
+            .HaveMaxLength(SUlid<T>.RawLength)
+            .AreFixedLength();
+    }
+
+    public static PropertiesConfigurationBuilder<Ulid<T>> AreTypedId<T>(this PropertiesConfigurationBuilder<Ulid<T>> builder)
+        where T : class, IIdentifiable<Ulid<T>>
+    {
+        return builder
+            .HaveConversion<UlidConverter<T>>()
+            .HaveMaxLength(26)
+            .AreFixedLength();
+    }
+
+    public static PropertiesConfigurationBuilder<Ulid<T>?> AreTypedId<T>(this PropertiesConfigurationBuilder<Ulid<T>?> builder)
+        where T : class, IIdentifiable<Ulid<T>>
+    {
+        return builder
+            .HaveConversion<UlidConverter<T>>()
+            .HaveMaxLength(26)
+            .AreFixedLength();
+    }
+}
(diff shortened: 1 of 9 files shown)


marcin-dardzinski:
Not sure what to think about pulling this lib straight into domain models, but otherwise it's okay to me.

src/Domain/LeanCode.DomainModels/Model/SUlid.cs
```diff
@@ -0,0 +1,135 @@
...
+[Serialization.TypedIdConverter]
+[System.Diagnostics.DebuggerDisplay("{ToString(),nq}")]
+[SuppressMessage("?", "CA1000", Justification = "The methods are expected.")]
+[SuppressMessage("?", "CA1036", Justification = "We don't want to have easy comparison as it might be abused.")]
+public readonly record struct SUlid<TEntity> : IEquatable<SUlid<TEntity>>, IComparable<SUlid<TEntity>>
+    where TEntity : class, IIdentifiable<SUlid<TEntity>>
+{
+    private const int UlidLength = 26;
+    private const char Separator = '_';
+    private static readonly string TypeName = SIdExtensions.GetPrefix<TEntity>();
```
lukaszgarstecki, line 16:
```suggestion
    private static readonly string TypePrefix = SIdExtensions.GetPrefix<TEntity>();
```

src/Domain/LeanCode.DomainModels/Model/SUlid.cs
```diff
@@ -0,0 +1,135 @@
...
+
+    public static SUlid<TEntity> FromString(string v)
+    {
+        if (IsValid(v))
+        {
+            return new SUlid<TEntity>(v);
+        }
+        else
+        {
+            throw new FormatException($"The id has invalid format. It should look like {TypeName}{Separator}(random id).");
```
lukaszgarstecki, line 70:
```suggestion
            throw new FormatException($"The id has invalid format. It should look like {TypeName}{Separator}(random ulid).");
```

lukaszgarstecki:
Overall seems good to test, let's see this in battle

Saancreed:
> Not sure what to think about pulling this lib straight into domain models, but otherwise it's okay to me.

@marcin-dardzinski Fiołek mentioned the same thing to me yesterday, I'll probably move all this stuff to a separate `LeanCode.DomainModels.Ulid` package. But I'm wondering if we'd like a separate `LeanCode.DomainModels.EF.Ulid` package as well or are we okay with `LeanCode.DomainModels.EF` unconditionally pulling `Ulid`.

src/Domain/LeanCode.DomainModels.EF/PropertiesConfigurationBuilderExtensions.cs
```diff
@@ -0,0 +1,44 @@
+using LeanCode.DomainModels.Model;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
+
+namespace LeanCode.DomainModels.EF;
+
+public static class PropertiesConfigurationBuilderExtensions
+{
+    public static PropertiesConfigurationBuilder<SUlid<T>> AreTypedId<T>(this PropertiesConfigurationBuilder<SUlid<T>> builder)
```
mishioo, line 9: Why `AreTypedId` rather than `IsTypedId`?

lukaszgarstecki: This method is used to configure convention, so the meaning here is `configurationBuilder.Properties<SUlid<T>>().AreTypedId()` is configured context wise, not just for a property, thus the plural here

jakubfijalkowski:
The more I think about it, the more I think we should either write our own `Ulid` implementation or vendor the `Ulid` from Cysharp. I would really like to keep the single package for our whole domain model bases, & keep it as dependency-free as possible. This might be beneficial when we finally move to source generated ids.

Other than that, I would consider removing the non-S version of ID.

Dragemil (approved):

Dragemil:
> I would really like to keep the single package for our whole domain model bases

But why? E.g. having ULIDs in one's project seems pretty opt-in to me.

jakubfijalkowski:
> E.g. having ULIDs in one's project seems pretty opt-in to me.

That's exactly what I don't want to have - being able to pick what you need. You get the whole package (with all our requirements) or you don't.

marcin-dardzinski:
Closing in favour of #505

PR #458, opened 2023-02-17
Extract StyleCop out of CodeAnalysis

mateuszkulak:


Directory.Build.targets
```diff
@@ -121,7 +121,7 @@
   </PropertyGroup>
 
   <ItemGroup>
```
jakubfijalkowski: This whole item group should also be removed (forgot about that earlier).

LeanCode.CodeAnalysis.ruleset
```diff
@@ -37,83 +37,4 @@
         <Rule Id="CA2234" Action="None"/>
         <!-- PassSystemUriObjectsInsteadOfStrings - most of the time it will result in an exception -->
     </Rules>
-
-    <Rules AnalyzerId="StyleCop.Analyzers" RuleNamespace="StyleCop.Analyzers">
```
jakubfijalkowski: Let's revert this change. Totally forgot that this is required for StyleCop to work. We can't split this because you can't have multiple `ruleset`s in a project, so until we migrate to [AnalyzerConfig](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-files#global-analyzerconfig)s, it needs to stay like that.

jakubfijalkowski (changes requested):

jakubfijalkowski (approved):

PR #461, opened 2023-02-21
Use file-scoped namespaces across all projects

mateuszkulak:


jakubfijalkowski (approved):
Summoning @marcin-dardzinski to see what has happened in the analyzers tests.

marcin-dardzinski:
> what has happened in the analyzers tests.

My wild guess is that they were indentation sensitive, but I really don't have time for it now.

PR #462, opened 2023-02-22
Issue 457 Use CSharpier

mateuszkulak:


.vscode/settings.json
```diff
@@ -11,5 +11,8 @@
     "paket-files": true,
     "**/obj/**/en": true,
     "**/obj/**/pl": true
-  }
+  },
+  "[csharp]": {
```
jakubfijalkowski: Not a fan of forcing a formatter (someone might want to use no formatter at all and format only before commit).

jakubfijalkowski (changes requested):

jakubfijalkowski (approved):

PR #465, opened 2023-02-27
Add Scoped Fluent Validation

mateuszkulak:


src/Domain/LeanCode.CQRS.Validation.Fluent.Scoped/ValidationContextExtensions.cs
```diff
@@ -0,0 +1,41 @@
+using System;
```
jakubfijalkowski: We should get rid of this file altogether.

src/Domain/LeanCode.CQRS.Validation.Fluent.Scoped/AdapterLoader.cs
```diff
@@ -0,0 +1,29 @@
...
+    where TAppContext : notnull
+    where TCommand : ICommand
+{
+    private static readonly Task<ValidationResult> NoError = Task.FromResult(new ValidationResult(null));
+
+    private readonly FluentValidationCommandValidatorAdapter<TAppContext, TCommand>? adapter;
+
+    public AdapterLoader(IComponentContext ctx)
+    {
+        if (ctx.TryResolve<IValidator<TCommand>>(out var val))
```
jakubfijalkowski: We can't resolve the validator up to a point where we know the `appContext` value - it can't be here.

src/Domain/LeanCode.CQRS.Validation.Fluent.Scoped/AdapterLoader.cs
```diff
@@ -0,0 +1,29 @@
...
+        {
+            adapter = new FluentValidationCommandValidatorAdapter<TAppContext, TCommand>(val, ctx);
+        }
+    }
+
+    public Task<ValidationResult> ValidateAsync(TAppContext appContext, TCommand command)
+    {
+        return adapter is null ? NoError : adapter.ValidateAsync(appContext, command);
```
jakubfijalkowski: We need to instantiate the `adapter` here, passing the `appContext` as a typed parameter to the [resolution](https://docs.autofac.org/en/stable/resolve/parameters.html).

src/Domain/LeanCode.CQRS.Validation.Fluent.Scoped/FluentValidationCommandValidatorAdapter.cs
```diff
@@ -0,0 +1,56 @@
...
+    private ValidationContext<TCommand> PrepareContext(TAppContext appContext, TCommand command)
+    {
+        var ctx = new ValidationContext<TCommand>(
+            command,
+            new PropertyChain(),
+            ValidatorOptions.Global.ValidatorSelectors.DefaultValidatorSelectorFactory()
+        );
+
+        ctx.RootContextData[ValidationContextExtensions.AppContextKey] = appContext;
+        ctx.RootContextData[ValidationContextExtensions.ComponentContextKey] = componentContext;
```
jakubfijalkowski: And since we can
a) take the dependencies in ctor,
b) we get the context in ctor,
we don't need to abuse `RootContextData`.

jakubfijalkowski (changes requested):
Also, remember to add the project to the solution.

src/Domain/LeanCode.CQRS.Validation.Fluent.Scoped/AdapterLoader.cs
mateuszkulak: I may be wrong, but I don't think that the registered validator takes an appContext as a parameter. So I don't understand why we need appContext to resolve it?

jakubfijalkowski: It related to [this comment](https://github.com/leancodepl/corelibrary/pull/465/files#r1118740477) - it _has to_ takee it if you want to use it, because you won't have other way to get to it.

src/Domain/LeanCode.CQRS.Validation.Fluent.Scoped/AdapterLoader.cs
```diff
@@ -0,0 +1,39 @@
...
+        this.ctx = ctx;
+    }
+
+    public Task<ValidationResult> ValidateAsync(TAppContext appContext, TCommand command)
+    {
+        IValidator<TCommand> val;
+
+        try
+        {
+            val = ctx.Resolve<IValidator<TCommand>>(new TypedParameter(typeof(TAppContext), appContext));
```
jakubfijalkowski: Let's use either `TryResolve` (I'm not sure if there is a correct overload) or `ResolveOptional` (there is a necessary overload). Using exceptions for control flow is an antipattern.

jakubfijalkowski:
I can't see tests for this feature :)

src/Domain/LeanCode.CQRS.Validation.Fluent.Scoped/FluentValidationCommandValidatorAdapter.cs
```diff
@@ -0,0 +1,51 @@
...
+
+    private static ValidationContext<TCommand> PrepareContext(TAppContext appContext, TCommand command)
+    {
+        var ctx = new ValidationContext<TCommand>(
+            command,
+            new PropertyChain(),
+            ValidatorOptions.Global.ValidatorSelectors.DefaultValidatorSelectorFactory()
+        );
+
+        return ctx;
```
jakubfijalkowski:
```suggestion
    {
        return new ValidationContext<TCommand>(
            command,
            new PropertyChain(),
            ValidatorOptions.Global.ValidatorSelectors.DefaultValidatorSelectorFactory()
        );
```

src/Domain/LeanCode.CQRS.Validation.Fluent.Scoped/FluentValidationCommandValidatorAdapter.cs
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    private ValidationError MapFluentError(ValidationFailure failure)
+    {
+        var state = failure.CustomState as FluentValidatorErrorState;
+
+        return new ValidationError(failure.PropertyName, failure.ErrorMessage, state?.ErrorCode ?? 0);
+    }
+
+    private static ValidationContext<TCommand> PrepareContext(TAppContext appContext, TCommand command)
```
jakubfijalkowski:
```suggestion
    private static ValidationContext<TCommand> PrepareContext(TCommand command)
```

`appContext` is not used

src/Domain/LeanCode.CQRS.Validation.Fluent.Scoped/AdapterLoader.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    private static readonly Task<ValidationResult> NoError = Task.FromResult(new ValidationResult(null));
+    private IComponentContext ctx;
+
+    public AdapterLoader(IComponentContext ctx)
+    {
+        this.ctx = ctx;
+    }
+
+    public Task<ValidationResult> ValidateAsync(TAppContext appContext, TCommand command)
```
jakubfijalkowski, line 22: These class, especially the resolution, needs to be checked. To do that, extend `ModuleTests` with:
1. A test that resolve a command validator where there is no validator,
2. A test that resolves a command validator where the validator takes `TAppContext` as a ctor parameter (and we need to ensure that the value there is correct!),
3. A test that resolves a command validator where the validator takes some external dependency as a ctor parameter,
4. And last, but not least, we need to ensure that, when we resolve the same validator in two different scopes, we get two different instances.

jakubfijalkowski (changes requested):

test/Domain/LeanCode.CQRS.Validation.Fluent.Scoped.Tests/ModuleTests.cs
```diff
@@ -0,0 +1,134 @@
...
+{
+    public CustomValidatorTakingExternalDependency(ExternalDependency dependency)
+    {
+        RuleFor(c => c.Field).Empty().WithCode(dependency.Number);
+    }
+}
+
+public class ExternalDependency
+{
+    public int Number = 13;
```
jakubfijalkowski:
```suggestion
    public int Number { get; set; } = 13;
```

test/Domain/LeanCode.CQRS.Validation.Fluent.Scoped.Tests/ModuleTests.cs
```diff
@@ -0,0 +1,134 @@
...
+{
+    public CustomValidatorTakingTAppContext(AppCtx appContext)
+    {
+        RuleFor(c => c.Field).GreaterThan(5).WithMessage(appContext.Message);
+    }
+}
+
+public class AppCtx
+{
+    public string Message = "test";
```
jakubfijalkowski:
```suggestion
    public string Message { get; set; } = "test";
```

jakubfijalkowski (approved):
Last two nitpicks, otherwise looks good!

PR #466, opened 2023-02-27
Source Generated IDs

### LINKED ISSUE #364: Use source generators for typed ids
Currently, our typed id feature is implemented using generics. It works, but since [source generators](https://docs.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/source-generators-overview) appeared we could reimplement ids with them. Potential usage
```
[TypedId(prefix: "user")]
public record struct UserId {}
```
The generator would add the actual implementation, equality, comparison, casting operators etc.

Things to consider:
- json serialization support (might be harder without stable interface)
- ef core conversion support (as above)

jakubfijalkowski:
A initial design of source-generated IDs. They support the same feature set as our `Id`/`IId`/`LId`/`SId` plus some (i.e. you can override how the initial value is generated).

Things left to do:
- [x] Tests of EFCore integration,
- [x] Fixed docs,
- [ ] Decide if we want to deprecate good old `XId<T>`s.

Closes #364


lukaszgarstecki (approved):
Overall LGTM. Do we plan to support configuring the db context using `ConfigureConventions`?

src/Domain/LeanCode.DomainModels.EF/TypedIdConverter.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace LeanCode.DomainModels.EF;
+
+[System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]
+public class PrefixedTypedIdConverter<TId> : ValueConverter<TId, string>
+    where TId : struct, IPrefixedTypedId<TId>
+{
+    public static readonly PrefixedTypedIdConverter<TId> Instance = new();
+
+    private PrefixedTypedIdConverter()
```
Dragemil: Making the ctors `public` would allow the converters to be used when configuring DB conventions.

docs/domain/ids.md
```diff
@@ -0,0 +1,86 @@
...
+
+The format of the ID can be configured using:
+
+1. `TypedIdFormat` - this configures the underlying type and the format of the ID. You need to specify it as a first parameter to the `TypedIdAttribute`. Possibilities:
+   a) `RawInt` - uses `int` as the underlying type; works as a wrapper over `int`; does not support generating new IDs at runtime by default;
+   b) `RawLong` - uses `long` as the underlying type; works as a wrapper over `long`; does not support generating new IDs at runtime by default;
+   c) `RawGuid` - uses `Guid` as the underlying type; works as a wrapper over `Guid`; can generate new ID at runtime using `Guid.NewGuid`;
+   d) `PrefixedGuid` - uses `string` as the underlying type; it is represented as a `(prefix)_(guid)` string that can be generated at runtime; by default `(prefix)` is a lowercase class name with `id` at the end removed;
+2. `CustomPrefix` - for `Prefixed*` formats, you can configure what prefix it uses (if you e.g. want to use a shorter one),
+3. `CustomGenerator` - allows specifying different code to generate the random value for ID; you can use it to add `New` for `Raw` formats.
```
marcin-dardzinski: Do we need this? 
You can always create your own factory method in your struct definition, seems to be more readable then passing some stringed fragments of code.

For defaults, we could always have `CreateRandomGenerator = true` flag.

src/Domain/LeanCode.DomainModels.EF/PropertyBuilderExtensions.cs
```diff
@@ -61,4 +62,38 @@ public static PropertyBuilder<SId<T>> IsTypedId<T>(this PropertyBuilder<SId<T>>
             .IsFixedLength()
             .ValueGeneratedNever();
     }
+
+    public static PropertyBuilder<TId> IsTypedId<TBacking, TId>(this PropertyBuilder<TId> builder)
```
Dragemil: We could expose methods for configuring conventions as well.

src/Domain/LeanCode.DomainModels.Generators/TypedIdFormat.cs
```diff
@@ -0,0 +1,9 @@
+namespace LeanCode.DomainModels.Generators;
+
+public enum TypedIdFormat : int
```
Dragemil, line 3: Would be nice to coordinate this PR with ULIDs, so that a new project can embrace them both as a single breaking change.

src/Tools/LeanCode.CodeAnalysis/DiagnosticsIds.cs
```diff
@@ -6,4 +6,5 @@ public static class DiagnosticsIds
     public const string QueriesShouldHaveAuthorizers = "LNCD0002";
     public const string CommandsShouldHaveValidators = "LNCD0003";
     public const string OperationsShouldHaveAuthorizers = "LNCD0004";
+    public const string TypedIdShouldBeStruct = "LNCD0005";
```
marcin-dardzinski: Nit: should or must? :upside_down_face:

src/Domain/LeanCode.DomainModels.EF/TypedIdConverter.cs
jakubfijalkowski: I fear it won't allow that - there is too much magic going on. If conventions _will_ be available, they should be supported by the CoreLib.

src/Domain/LeanCode.DomainModels.Generators/TypedIdFormat.cs
jakubfijalkowski: ULIDs must come later (and, IMO, I wouldn't allow ULIDs in the current form).

src/Domain/LeanCode.DomainModels/Ids/TypedIdAttribute.cs
```diff
@@ -0,0 +1,67 @@
...
+
+/// <summary>
+/// Attribute that mark <c>readonly partial record struct</c> for being <c>TypedId</c>, i.e. a stronly typed identifier
+/// that will be generated during compilation and that can be used as a ID for
+/// <see cref="Model.IIdentifiable{TIdentity}" /> and <see cref="Model.IAggregateRoot{TIdentity}" />.
+/// </summary>
+/// <remarks>
+/// See <see href="../../../../docs/domain/ids.md">documentation</see> for more details.
+/// </remarks>
+[AttributeUsage(AttributeTargets.Struct, Inherited = false, AllowMultiple = true)]
```
marcin-dardzinski: Does it makes sense to allow multiple usage here?

marcin-dardzinski (commented):
Looks awesome, just a few nitpicks.

src/Domain/LeanCode.DomainModels.EF/TypedIdConverter.cs
jakubfijalkowski: W8, does conversions are configured _per type_ either way?

src/Domain/LeanCode.DomainModels/Ids/TypedIdAttribute.cs
jakubfijalkowski: Nooope, forgot to change that.

docs/domain/ids.md
jakubfijalkowski: That's a great remark. Will change that.

src/Domain/LeanCode.DomainModels.EF/TypedIdConverter.cs
Dragemil: > If conventions will be available, they should be supported by the CoreLib.

I don't get what you mean here. In order to use `builder.Properties<Type>().HaveConversion<TypeConverter>();` the `TypeConverter` must expose public argless ctor.

> W8, does conversions are configured per type either way?

Yes, `DbContext` conventions in `ConfigureConventions` work per type.

src/Domain/LeanCode.DomainModels.EF/PropertyBuilderExtensions.cs
jakubfijalkowski: DOne

Dragemil (approved):

marcin-dardzinski (approved):

src/Domain/LeanCode.DomainModels/Ids/TypedIdAttribute.cs
```diff
@@ -0,0 +1,67 @@
...
+/// </remarks>
+public enum TypedIdFormat : int
+{
+    /// <summary>
+    /// Raw <see cref="int" />, without prefix. It's backing type is <see cref="int" />.
+    /// </summary>
+    RawInt = 0,
+
+    /// <summary>
+    /// Raw <see cref="long" />, without prefix. It's backing type is <see cref="int" />.
```
marcin-dardzinski: It's backing type is **long**
Same for guid below.

src/Domain/LeanCode.DomainModels.Generators/IdSource.cs
```diff
@@ -0,0 +1,243 @@
...
+    )
+    {
+        var prefix = data.CustomPrefix?.ToLowerInvariant() ?? GetDefaultPrefix(data.TypeName);
+        var fromCtor = generateFromCtor
+            ? $@"public {data.TypeName}(Guid v) => value = string.Create(null, stackalloc char[RawLength], $""{{TypePrefix}}{{Separator}}{{v:{rawFormat}}}"");"
+            : "";
+        var randomFactory = !data.SkipRandomGenerator
+            ? $@"public static {data.TypeName} New() => new({randomValueGenerator});"
+            : "";
+        return $@"// <auto-generated />
```
Saancreed: Wouldn't all this be more readable using `$$"""` multiline strings?

src/Domain/LeanCode.DomainModels/Ids/ITypedId.cs
```diff
@@ -0,0 +1,36 @@
+using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
+
+namespace LeanCode.DomainModels.Ids;
+
+[SuppressMessage("?", "CA1000", Justification = "Roslyn bug.")]
+[EditorBrowsable(EditorBrowsableState.Never)]
+public interface IPrefixedTypedId<TSelf> : IEquatable<TSelf>, IComparable<TSelf>
```
Saancreed, line 9: Can we also add and implement `ISpanFormattable` and `ISpanParsable`? They are very useful when performing string interpolation/formatting (and they would be even more useful if we used any stringlike type (Ulid? :stuck_out_tongue:) that isn't directly supported by STJ).

src/Domain/LeanCode.DomainModels/Ids/ITypedId.cs
```diff
@@ -0,0 +1,36 @@
...
+    where TSelf : struct, IPrefixedTypedId<TSelf>
+{
+    string Value { get; }
+    public static abstract int RawLength { get; }
+    public static abstract TSelf Parse(string? v);
+
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public static abstract System.Linq.Expressions.Expression<Func<string, TSelf>> FromDatabase { get; }
```
Saancreed: Nit, but `using System.Linq.Expressions;` at the top maybe?

test/Domain/LeanCode.DomainModels.Tests/LeanCode.DomainModels.Tests.csproj
```diff
@@ -4,8 +4,40 @@
...
+
   <ItemGroup>
     <ProjectReference Include="../../../src/Domain/LeanCode.DomainModels/LeanCode.DomainModels.csproj" />
+    <ProjectReference
+      Include="../../../src/Domain/LeanCode.DomainModels.Generators/LeanCode.DomainModels.Generators.csproj"
+      OutputItemType="Analyzer" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NETCore.App.Ref" Version="7.0.0" IncludeAssets="none" PrivateAssets="all" GeneratePathProperty="true" />
```
Saancreed, line 19: `GeneratePathProperty` allows you to find the path using properties named like `Pkg${PACKAGE//./_}` instead of accessing `$(NuGetPackageRoot)`, we should probably do that.

src/Domain/LeanCode.DomainModels/Ids/ITypedId.cs
jakubfijalkowski: Next version, this one is already big. Will add that to backlog.

src/Domain/LeanCode.DomainModels/Ids/ITypedId.cs
jakubfijalkowski: +1

src/Domain/LeanCode.DomainModels.Generators/IdSource.cs
jakubfijalkowski: I don't think it would change that much, but will try

Saancreed (approved):
All good, but I'd personally prefer if prefixed string identifiers always required the prefix to be specified explicitly. But it's no big deal if they don't.

Issue #468, opened 2023-03-01
Implement `ISpanFormattable` and `ISpanParsable` in source-generated IDs


jakubfijalkowski:
Can we also add and implement `ISpanFormattable` and `ISpanParsable`? They are very useful when performing string interpolation/formatting (and they would be even more useful if we used any stringlike type (Ulid? :stuck_out_tongue:) that isn't directly supported by STJ).

_Originally posted by @Saancreed in https://github.com/leancodepl/corelibrary/pull/466#discussion_r1120013920_

PR #470, opened 2023-03-06
Allow serializing all IDs as dictionary keys

Dragemil:


src/Domain/LeanCode.DomainModels/Ids/TypedIdConverter.cs
```diff
@@ -34,6 +55,18 @@ public class LongTypedIdConverter<TId> : JsonConverter<TId>
 
     public override void Write(Utf8JsonWriter writer, TId value, JsonSerializerOptions options) =>
         writer.WriteNumberValue(value.Value);
+
+    public override TId ReadAsPropertyName(
+        ref Utf8JsonReader reader,
+        Type typeToConvert,
+        JsonSerializerOptions options
+    ) =>
+        reader.GetString() is string s && long.TryParse(s, out var id)
```
jakubfijalkowski: Why not [GetInt64](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonreader.getint64?view=net-7.0)?

src/Domain/LeanCode.DomainModels/Ids/TypedIdConverter.cs
```diff
@@ -45,4 +78,16 @@ public class GuidTypedIdConverter<TId> : JsonConverter<TId>
 
     public override void Write(Utf8JsonWriter writer, TId value, JsonSerializerOptions options) =>
         writer.WriteStringValue(value.Value);
+
+    public override TId ReadAsPropertyName(
+        ref Utf8JsonReader reader,
+        Type typeToConvert,
+        JsonSerializerOptions options
+    ) =>
+        reader.GetString() is string s && Guid.TryParse(s, out var id)
```
jakubfijalkowski: Why not [GetGuid](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonreader.getguid?view=net-7.0)?

src/Domain/LeanCode.DomainModels/Ids/TypedIdConverter.cs
```diff
@@ -23,6 +32,18 @@ public class IntTypedIdConverter<TId> : JsonConverter<TId>
...
         writer.WriteNumberValue(value.Value);
+
+    public override TId ReadAsPropertyName(
+        ref Utf8JsonReader reader,
+        Type typeToConvert,
+        JsonSerializerOptions options
+    ) =>
+        reader.GetString() is string s && int.TryParse(s, out var id)
+            ? TId.Parse(id)
+            : throw new JsonException($"Could not deserialize {typeToConvert.Name}");
```
Saancreed: Read to span, parse from span. Also let's use `{ return …; }` blocks for multiline expressions.

src/Domain/LeanCode.DomainModels/Model/TypedIdConverterAttribute.cs
```diff
@@ -49,6 +49,25 @@ public override void Write(Utf8JsonWriter writer, Id<T> value, JsonSerializerOpt
...
             writer.WriteStringValue(value.Value);
         }
+
+        public override Id<T> ReadAsPropertyName(
+            ref Utf8JsonReader reader,
+            Type typeToConvert,
+            JsonSerializerOptions options
+        )
+        {
+            if (reader.GetString() is string s && Guid.TryParse(s, out var id))
```
jakubfijalkowski: Why not [GetGuid](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonreader.getguid?view=net-7.0)?

src/Domain/LeanCode.DomainModels/Model/TypedIdConverterAttribute.cs
```diff
@@ -68,6 +87,25 @@ public override void Write(Utf8JsonWriter writer, IId<T> value, JsonSerializerOp
...
             writer.WriteNumberValue(value.Value);
         }
+
+        public override IId<T> ReadAsPropertyName(
+            ref Utf8JsonReader reader,
+            Type typeToConvert,
+            JsonSerializerOptions options
+        )
+        {
+            if (reader.GetString() is string s && int.TryParse(s, out var id))
```
jakubfijalkowski: ditto

src/Domain/LeanCode.DomainModels/Model/TypedIdConverterAttribute.cs
```diff
@@ -87,6 +125,25 @@ public override void Write(Utf8JsonWriter writer, LId<T> value, JsonSerializerOp
...
             writer.WriteNumberValue(value.Value);
         }
+
+        public override LId<T> ReadAsPropertyName(
+            ref Utf8JsonReader reader,
+            Type typeToConvert,
+            JsonSerializerOptions options
+        )
+        {
+            if (reader.GetString() is string s && long.TryParse(s, out var id))
```
jakubfijalkowski: ditto

src/Domain/LeanCode.DomainModels/Ids/TypedIdConverter.cs
```diff
@@ -23,6 +32,18 @@ public class IntTypedIdConverter<TId> : JsonConverter<TId>
 
     public override void Write(Utf8JsonWriter writer, TId value, JsonSerializerOptions options) =>
         writer.WriteNumberValue(value.Value);
+
+    public override TId ReadAsPropertyName(
+        ref Utf8JsonReader reader,
+        Type typeToConvert,
+        JsonSerializerOptions options
+    ) =>
+        reader.GetString() is string s && int.TryParse(s, out var id)
```
jakubfijalkowski: ditto

src/Domain/LeanCode.DomainModels/Ids/TypedIdConverter.cs
```diff
@@ -23,6 +32,18 @@ public class IntTypedIdConverter<TId> : JsonConverter<TId>
...
+            ? TId.Parse(id)
+            : throw new JsonException($"Could not deserialize {typeToConvert.Name}");
+
+    public override void WriteAsPropertyName(Utf8JsonWriter writer, TId value, JsonSerializerOptions options) =>
+        writer.WritePropertyName(value.ToString()!);
```
Saancreed: Format to span, write from span.

src/Domain/LeanCode.DomainModels/Ids/TypedIdConverter.cs
Saancreed:
```suggestion
        reader.TryGetGuid(out var id)
```

Saancreed (commented):
Some allocation-related nits. I'll take a closer look later today.

src/Domain/LeanCode.DomainModels/Ids/TypedIdConverter.cs
Saancreed: The token value won't be a `Number` here.

jakubfijalkowski: Good point, sorry

src/Domain/LeanCode.DomainModels/Ids/TypedIdConverter.cs
Dragemil: Unfortunately it throws:
```
System.InvalidOperationException
Cannot get the value of a token type 'PropertyName' as a string.
   at System.Text.Json.ThrowHelper.ThrowInvalidOperationException_ExpectedString(JsonTokenType tokenType)
   at System.Text.Json.Utf8JsonReader.TryGetGuid(Guid& value)
   at LeanCode.DomainModels.Model.TypedIdConverterAttribute.IdConverter`1.Read(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options)
```

It checks whether the JSON token is `string` token 😕

jakubfijalkowski: D'oh :(

Dragemil: As mentioned above, it throws on `PropertyName` JSON token.

Saancreed (approved):
Looks reasonable, let's see if @jakubfijalkowski agrees :see_no_evil:

src/Domain/LeanCode.DomainModels/Ids/TypedIdConverter.cs
```diff
@@ -12,37 +16,143 @@ public class StringTypedIdConverter<TId> : JsonConverter<TId>
...
+
+    public override void WriteAsPropertyName(Utf8JsonWriter writer, TId value, JsonSerializerOptions options) =>
+        writer.WritePropertyName(value.Value);
 }
 
 [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]
 public class IntTypedIdConverter<TId> : JsonConverter<TId>
     where TId : struct, IRawTypedId<int, TId>
 {
+    private static readonly int SpanSize = int.MinValue.ToString(CultureInfo.InvariantCulture).Length;
```
jakubfijalkowski: I'm not sure if we should deduce the span size like that. IMO this is no better than `const int SpanSize = 11` because it assumes that `int.MinValue.ToString(CultureInfo.InvariantCulture).Length > int.MaxValue.ToString(CultureInfo.InvariantCulture).Length`

jakubfijalkowski: I would, but that is only an opinion, go with `SpanSize = 16` (or 32, doesn't really matter) and would add a comment explaining that.

jakubfijalkowski (approved):
I wonder if it is high time to solve issue #468 first, since most of the code is already done :P

Dragemil:
> I wonder if it is high time to solve issue first #468, since most of the code is already done :P

Some other time if I'm to do it, I just wanted to serialize some dicts for now.

src/Domain/LeanCode.DomainModels/Ids/TypedIdConverter.cs
Dragemil: Doesn't matter to me, hardcoded this as you've suggested for now.

PR #472, opened 2023-03-14
MassTransit messages redelivers

mateuszkulak:


test/Domain/LeanCode.DomainModels.MassTransitRelay.Tests/LeanCode.DomainModels.MassTransitRelay.Tests.csproj
```diff
@@ -8,6 +8,7 @@
     <ProjectReference Include="../../../src/Testing/LeanCode.IntegrationTestHelpers/LeanCode.IntegrationTestHelpers.csproj" />
 
     <ProjectReference Include="../../LeanCode.Test.Helpers/LeanCode.Test.Helpers.csproj" />
+    <PackageReference Include="Quartz" Version="3.6.2" />
```
jakubfijalkowski, line 11: Version should be specified in the central targets file, just like others.

src/Domain/LeanCode.DomainModels.MassTransitRelay/Middleware/StoreAndPublishEventsFilter.cs
```diff
@@ -24,12 +24,15 @@ public StoreAndPublishEventsFilter(AsyncEventsInterceptor interceptor, EventsSto
...
 
-        await store.StoreAndPublishEventsAsync(
-            events,
-            context.ConversationId,
-            new EventPublisher(context),
-            context.CancellationToken
-        );
+        if (context.GetRedeliveryCount() == 0)
+        {
+            await store.StoreAndPublishEventsAsync(
```
marcin-dardzinski, line 29: This means that any event raised during redelivery wouldn't be published. Also if the consumer relies on this piece to save it's data - it'd also be lost.

src/Domain/LeanCode.DomainModels.MassTransitRelay/Middleware/ConsumeMessageOnceFilter.cs
```diff
@@ -30,6 +30,7 @@ public ConsumeMessageOnceFilter(IConsumedMessagesContext consumedMessages)
             await consumedMessages.ConsumedMessages
                 .Where(m => m.ConsumerType == msg.ConsumerType && m.MessageId == msg.MessageId)
                 .AnyAsync()
+            && context.GetRedeliveryCount() == 0
```
marcin-dardzinski, line 33: This means that we actually do not acknowledge consumption of any redelivered message, which might be unexpected for the user.

jakubfijalkowski, line 33: Why extending this condition is needed? I think it would introduce unnecessary retries in this case:
1) Consumer run, it decided to redeliver the message (`RedeliveryCount++`),
2) Message redelivered but MT failed to ACK it (so, the `ConsumedMessage` was added, `RedeliveryCount` stayed 1)
3) ASB redelivered the message once again, with `RedeliveryCount == 1`

In this case, the message should be excluded by the `ConsumedMessages` check.

marcin-dardzinski (commented):
I think if we want to handle it the correct way we need to track every redelivery separately in `ConsumedMessages` - otherwise the behaviour might be not as the consumers expect. 

I'd also take a look how native MT outbox handles such cases and 'get inspired' (btw I'm fan of migrating to the MT outbox anyway).

test/Domain/LeanCode.DomainModels.MassTransitRelay.Tests/RedeliverMessagesTests.cs
```diff
@@ -0,0 +1,169 @@
...
+        {
+            configuration.UseConcurrencyLimit(1);
+            configuration.UseDelayedMessageScheduler();
+            configuration.StoreAndPublishDomainEvents(container.Resolve<IServiceProvider>());
+            configuration.UseConsumedMessagesFiltering(container.Resolve<IServiceProvider>());
+        };
+    }
+
+    [Fact]
+    public async Task Redelivered_messages_are_not_filtered()
```
jakubfijalkowski, line 47: We should have a separate cases for when:
1) Message runs successfully w/o redelivery the first time,
2) The first consumer run requests redelivery,
3) The message is redelivered, but the second attempt also requests redelivery.

jakubfijalkowski: Also, TBH, I don't believe these tests - the redelivery is triggered in the test, and it should be in the handler. You never check if the consumer has ever run, and I'm not sure if "message consumed" means that the consumer ran, or if just the MT processed the message (and ignored), but I might be wrong here. You also never check if the message ends up in the consumed messages or if the events are published.

All these problems needs to be addressed.

jakubfijalkowski (changes requested):

PR #473, opened 2023-03-22
Issue/378 lean startup forces log.close and flush

mateuszkulak:


ZoltePudeleczko (approved):
We have example of setting up `LeanStartup` class in `01_app.md` doc - it would be nice to add there info about this option and change example code to directly pass flag value - `base(config)` -> `base(config, true)`

src/Core/LeanCode.Components.Startup/LeanStartup.cs
```diff
@@ -11,6 +11,7 @@ public abstract class LeanStartup
 {
     protected IConfiguration Configuration { get; }
     protected ILogger Logger { get; }
+    protected bool CloseAndFlushLogger { get; set; } = true;
```
jakubfijalkowski:
```suggestion
    protected virtual bool CloseAndFlushLogger { get; } = true;
```

src/Core/LeanCode.Components.Startup/LeanStartup.cs
```diff
@@ -49,7 +56,10 @@ public void Configure(IApplicationBuilder app)
 
         var appLifetime = app.ApplicationServices.GetRequiredService<IHostApplicationLifetime>();
 
-        appLifetime.ApplicationStopped.Register(Log.CloseAndFlush);
+        if (CloseAndFlushLogger == true)
```
jakubfijalkowski:
```suggestion
        if (CloseAndFlushLogger)
```

src/Core/LeanCode.Components.Startup/LeanStartup.cs
```diff
@@ -22,6 +23,12 @@ protected LeanStartup(IConfiguration config)
         Logger.Information("Configuration loaded, starting app");
     }
 
+    protected LeanStartup(IConfiguration config, bool closeAndFlushLogger)
```
jakubfijalkowski: Let's remove this ctor and use the ability to override the property in derived classes - will be easier to use (and will have proper defaults).

test/Core/LeanCode.Components.Startup.Tests/LeanStartupTests.cs
```diff
@@ -0,0 +1,91 @@
...
+        var logger = host.Services.GetService<ILogger<TestStartupWithTrueParameter>>();
+
+        var isSinkClosed =
+            !logger.IsEnabled(LogLevel.Debug)
+            && !logger.IsEnabled(LogLevel.Information)
+            && !logger.IsEnabled(LogLevel.Warning)
+            && !logger.IsEnabled(LogLevel.Error)
+            && !logger.IsEnabled(LogLevel.Critical);
+
+        Assert.True(isSinkClosed);
```
jakubfijalkowski: This can be done simpler - we know that we use Serilog and we know how it works, we can [use that knowledge](https://github.com/serilog/serilog/blob/f72a7a328d38f559632ae64aba08dfcf0b1dae44/src/Serilog/Log.cs#L51-L56) to make this assert:

```suggestion
        Assert.Same(Log.Logger, Serilog.Core.Logger.None);
```

test/Core/LeanCode.Components.Startup.Tests/LeanStartupTests.cs
```diff
@@ -0,0 +1,91 @@
...
+        var logger = host.Services.GetService<ILogger<TestStartupWithFalseParameter>>();
+
+        var isSinkClosed =
+            !logger.IsEnabled(LogLevel.Debug)
+            && !logger.IsEnabled(LogLevel.Information)
+            && !logger.IsEnabled(LogLevel.Warning)
+            && !logger.IsEnabled(LogLevel.Error)
+            && !logger.IsEnabled(LogLevel.Critical);
+
+        Assert.False(isSinkClosed);
```
jakubfijalkowski:
```suggestion
        Assert.NotSame(Log.Logger, Serilog.Core.Logger.None);
```

jakubfijalkowski (changes requested):

src/Core/LeanCode.Components.Startup/LeanStartup.cs
```diff
@@ -11,6 +11,7 @@ public abstract class LeanStartup
 {
     protected IConfiguration Configuration { get; }
     protected ILogger Logger { get; }
+    protected virtual bool CloseAndFlushLogger { get; set; } = true;
```
jakubfijalkowski:
```suggestion
    protected virtual bool CloseAndFlushLogger { get; } = true;
```

Otherwise, someone from the outside can "just" set it (just the way you do in tests, which is not the aim IMO).

jakubfijalkowski (changes requested):

jakubfijalkowski (approved):

PR #474, opened 2023-03-22
Add document describing how to create Query

mateuszkulak:


docs/basics/04_how_to_create_query.md
```diff
@@ -0,0 +1,52 @@
+## Adding new query
+In order to create query you have to follow those steps.
+
+### 1. Create Query
+Query creation requires creating class that implements `IQuery<TResult>` interface. `TResult` parameter specifies the type that query returns when executed. Query should define properties with will be later use in QueryHandler. Defining a Query, you can specify Attributes such as `[AllowUnauthorized]` or `[AuthorizeWhenHasAnyOf("Some_permission")]` to pass additional metadata.
```
ZoltePudeleczko, line 5: Can we add hyperlinks for things like `IQuery`? I think it would be neat if it could point to interface implementation

docs/basics/04_how_to_create_query.md
```diff
@@ -0,0 +1,52 @@
+## Adding new query
+In order to create query you have to follow those steps.
+
+### 1. Create Query
```
ZoltePudeleczko, line 4:
```suggestion
### 1. Create Query contract
```
maybe?

docs/basics/04_how_to_create_query.md
ZoltePudeleczko, line 5: Typo 'define properties with will' -> 'define properties which will'

ZoltePudeleczko, line 5: Also I think authorization attributes deserve they own part of this (or maybe even another one) document. They're more than just passing some metadata

docs/basics/04_how_to_create_query.md
```diff
@@ -0,0 +1,52 @@
...
+{
+    public string Name { get; set; }
+}
+
+public class QueryResultDTO
+{
+    public string Greeting { get; set; }
+}
+```
+Important things is that TResult should be DTO (but it can also be an enum) because most of the time it will be serialized.
```
ZoltePudeleczko, line 24: What does it mean that 'TResult should be DTO'?

PR #475, opened 2023-03-24
Change ConfigurationOverrides class

mateuszkulak:


src/Testing/LeanCode.IntegrationTestHelpers/ConfigurationOverrides.cs
```diff
@@ -6,89 +6,71 @@ namespace LeanCode.IntegrationTestHelpers;
...
 
-            Data = new Dictionary<string, string?>
-            {
-                [connectionStringKey] = dbConnStr,
-                [internalBaseKey] = "http://localhost",
-                [publicBaseKey] = "http://localhost",
-                [IHostBuilderExtensions.EnableDetailedInternalLogsKey] = enableInternalLogs.ToString(),
-                [IHostBuilderExtensions.MinimumLogLevelKey] = minimumLevel.ToString(),
-            };
+    public class OverridesBuilder
```
jakubfijalkowski: We can simplify this IMO - the `ConfigurationOverrides` and `OverridesBuilder` can be one and the same class. The overrides class can work as a builder (it is a builder - for the `Provider` class) and we don't need to make one more indirection.

Otherwise, looks good.

src/Testing/LeanCode.IntegrationTestHelpers/ConfigurationOverrides.cs
```diff
@@ -6,89 +6,71 @@ namespace LeanCode.IntegrationTestHelpers;
...
+    {
+        private readonly Dictionary<string, string?> data = new();
+
+        public OverridesBuilder AddValue(string key, string? value)
```
jakubfijalkowski:
```suggestion
        public OverridesBuilder Add(string key, string? value)
```

test/Testing/LeanCode.IntegrationTestHelpers.Tests/ConfigurationOverridesTests.cs
```diff
@@ -0,0 +1,53 @@
+using Microsoft.Extensions.Configuration;
+using Xunit;
+
+namespace LeanCode.IntegrationTestHelpers.Tests;
+
+public class ConfigurationOverridesTests
+{
+    [Fact]
+    public void TestBuild()
```
jakubfijalkowski: This should be split into a number of tests:
1) Test if empty overrides results in empty configuration (this is the most basic case),
2) Test if a single `AddValue` works,
3) Test if multiple (separate) `AddValue` works,
4) Test if `AddValue` called multiple times (2) with the same key work as expected,
5) The same set of 2-4 tests for connection strings,
6) And a mix of both `AddValue` and `AddConnectionString` (but a simple test with a single call to both).

docs/integrationTests/README.md
jakubfijalkowski, line 1: `docs/integration_tests`

docs/integrationTests/README.md
```diff
@@ -0,0 +1,52 @@
+# ConfigurationOverrides class
```
jakubfijalkowski: Too detailed - we want to describe the idea, not every method on the class.

jakubfijalkowski (changes requested):

docs/integration_tests/MultipleDbContexts/README.md
jakubfijalkowski, line 1: `docs/integration_tests/configuration.md`

docs/integration_tests/MultipleDbContexts/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+ConfigurationOverrides class provides a way to override configuration values at runtime using a dictionary of key-value pairs. The class has a dictionary field named `data` which stores the configuration key-value pairs that will be used to override the original configuration values.
+
+The class has two public methods for adding key-value pairs to the data dictionary:
+
+- `Add` : Adds a single key-value pair to the dictionary.
+- `AddConnectionString` : Adds a connection string to the dictionary.
+
+
+
```
jakubfijalkowski: Too many new lines.

docs/integration_tests/MultipleDbContexts/README.md
```diff
@@ -0,0 +1,11 @@
+# ConfigurationOverrides class
```
jakubfijalkowski: Let's structure the file differently:
1. First introduce `ConfigurationOverrides` as a general concept and what it is used for, without going into the detail on how it works (the description of `data` field is unnecessary),
2. Then describe how you can use that if you have a single `DbContext` in the app,
3. Then describe how you can use that if you have multiple `DbContext`s.

src/Testing/LeanCode.IntegrationTestHelpers/ConfigurationOverrides.cs
```diff
@@ -1,94 +1,62 @@
...
-        this.connectionStringKey = connectionStringKey ?? ConnectionStringKeyDefault;
-        this.internalBaseKey = internalBaseKey ?? InternalBaseKeyDefault;
-        this.publicBaseKey = publicBaseKey ?? PublicBaseKeyDefault;
+        var connectionString = OverrideDatabaseConnectionString(dbPrefix, sourceEnv, varName);
+        data[key] = connectionString;
+        return this;
     }
 
-    public IConfigurationProvider Build(IConfigurationBuilder builder)
+    public static string OverrideDatabaseConnectionString(
```
jakubfijalkowski:
```suggestion
    public static string GenerateConnectionString(
```

test/Testing/LeanCode.IntegrationTestHelpers.Tests/ConfigurationOverridesTests.cs
```diff
@@ -0,0 +1,183 @@
...
+
+        Assert.Empty(configuration.AsEnumerable());
+    }
+
+    [Fact]
+    public void ConfigurationOverrides_single_Add_method_test()
+    {
+        var overrides = new ConfigurationOverrides().Add("key", "value");
+
+        var configuration = new ConfigurationBuilder().Add(overrides).Build();
```
jakubfijalkowski: Let's wrap the `new ConfigurationBuilder().Add(overrides).Build()` in a method called `BuildConfiguration(ConfigurationOverrides overrides)` - will make things cleaner.

test/Testing/LeanCode.IntegrationTestHelpers.Tests/ConfigurationOverridesTests.cs
```diff
@@ -0,0 +1,183 @@
...
+    [Fact]
+    public void Empty_ConfigurationOverrides_Results_In_Empty_Configuration_test()
+    {
+        var configuration = new ConfigurationBuilder().Add(new ConfigurationOverrides()).Build();
+
+        Assert.Empty(configuration.AsEnumerable());
+    }
+
+    [Fact]
+    public void ConfigurationOverrides_single_Add_method_test()
```
jakubfijalkowski: This name (and the others below) follows a totally different convention to the first test. We need to be consistent and name it, for example, `Adding_single_value_results_in_single_value_in_configration`.

jakubfijalkowski: Also, you don't need to specify `ConfigurationOverrides` in each class name - that is given by the class name.

test/Testing/LeanCode.IntegrationTestHelpers.Tests/ConfigurationOverridesTests.cs
```diff
@@ -0,0 +1,183 @@
...
+
+        var configuration = new ConfigurationBuilder().Add(overrides).Build();
+
+        Assert.Equal("value3", configuration["key"]);
+    }
+
+    [Fact]
+    public void ConfigurationOverrides_single_AddConnectionString_method_test()
+    {
+        var connectionStringKey = "ConnectionStrings:DbContext";
```
jakubfijalkowski: Most of these `var`s should be `const`.

jakubfijalkowski (changes requested):

jakubfijalkowski (approved):

PR #478, opened 2023-05-09
Replace custom outbox with the MassTransit one

### LINKED ISSUE #441: Integrate MassTransit Transactional Outbox
I very often find myself needing to transactionally publish some messages that are not `IDomainEvent`s . Right now the only workaround is to artificially create one and proxy it through domain, which is not convenient. 
 [Transactional Outbox](https://masstransit-project.com/advanced/transactional-outbox.html) should be fairly easy to integrate, looks like it's just a matter of injecting the correct outbox scope in CQRS/MT pipelines + we could remove a lot of out infrastructure code.

marcin-dardzinski:
Closes #441 

Essentially, just removed everything related to our outbox and switched the test to use MT's one.
Also:
- added `CommitDatabaseTransactionElement`
- adjusted `EventsPublisherElement` to use `IPublishEndpoint` - if MT bus outbox is enabled 
- MassTransit hosted service will now wait for bus start by default
- A helper method to add consumers with default configuration   

This is a breaking change, so keeping it as draft, we'll need a next version of corelibrary


src/Domain/LeanCode.DomainModels.MassTransitRelay/MassTransitRegistrationConfigurationExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+using MassTransit.Metadata;
+using MassTransit.Util;
+
+namespace LeanCode.DomainModels.MassTransitRelay;
+
+public static class MassTransitRegistrationConfigurationExtensions
+{
+    // These methods are based on
+    // https://github.com/MassTransit/MassTransit/blob/develop/src/MassTransit/Configuration/RegistrationExtensions.cs
+    public static void AddConsumersWithDefaultConfiguration(
```
jakubfijalkowski, line 13: We need tests (unit, preferably), for both of the methods. With and without custom definitions.

src/Domain/LeanCode.DomainModels.MassTransitRelay/MassTransitRelayModule.cs
```diff
@@ -1,66 +1,33 @@
...
             .SingleInstance();
-
-        builder.RegisterType<SimpleEventsExecutor>().AsSelf().SingleInstance().WithParameter("useOutbox", useOutbox);
-
-        builder.RegisterType<SimpleFinalizer>().AsSelf();
     }
 
     public sealed override void ConfigureServices(IServiceCollection services)
     {
+        services
```
jakubfijalkowski: If we change that here, I would make it configurable - i.e. make the `Configure` closure a virtual method here.

src/Domain/LeanCode.DomainModels.MassTransitRelay/MassTransitRelayModule.cs
```diff
@@ -1,66 +1,33 @@
...
             .AsSelf()
             .OnActivated(a => a.Instance.Configure())
             .SingleInstance();
-
-        builder.RegisterType<SimpleEventsExecutor>().AsSelf().SingleInstance().WithParameter("useOutbox", useOutbox);
-
-        builder.RegisterType<SimpleFinalizer>().AsSelf();
     }
 
     public sealed override void ConfigureServices(IServiceCollection services)
```
jakubfijalkowski: And, since we're here, un-seal this?

jakubfijalkowski (commented):
More general question - how is the "inbox" (i.e. process once) part handled? I see that the `ConsumeMessageOnceFilter` has been deleted, but is it provided by MT out of the box?

src/Domain/LeanCode.DomainModels.MassTransitRelay/MassTransitRelayModule.cs
marcin-dardzinski: Good point, I was thinking to just call `AddOptions` again, but virtual method would be better.

marcin-dardzinski:
> More general question - how is the "inbox" (i.e. process once) part handled? I see that the `ConsumeMessageOnceFilter` has been deleted, but is it provided by MT out of the box?

According to the [docs](https://masstransit.io/documentation/patterns/transactional-outbox#transactional-outbox):
> The Consumer Outbox is a combination of an inbox and an outbox. The inbox is used to keep track of received messages to guarantee exactly-once consumer behavior.

marcin-dardzinski:
OK, this one is ready.
There will be changes related to switching to asp.net middlewares, but I'd rather have it separately.

lukaszgarstecki (approved):

src/Domain/LeanCode.DomainModels.MassTransitRelay/MassTransitRegistrationConfigurationExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+        foreach (var item in enumerable)
+        {
+            if (item.DefinitionType is null)
+            {
+                configurator.AddConsumer(item.ConsumerType, defaultDefinition.MakeGenericType(item.ConsumerType));
+            }
+            else
+            {
+                configurator.AddConsumer(item.ConsumerType, item.DefinitionType);
+            }
```
Dragemil:
```suggestion
            configurator.AddConsumer(item.ConsumerType, item.DefinitionType ?? defaultDefinition.MakeGenericType(item.ConsumerType));
```

src/Domain/LeanCode.DomainModels.MassTransitRelay/Middleware/EventsPublisherElement.cs
```diff
@@ -44,25 +41,18 @@ private Task PublishEventsAsync(TContext ctx, List<IDomainEvent> events)
...
-            await publisher.PublishAsync(evt, conversationId, ctx.CancellationToken);
-        }
-        catch (Exception e)
-        {
-            logger.Fatal(e, "Could not publish event {@DomainEvent}", evt);
-        }
-
-        logger.Information("Domain event {DomainEvent} published", evt);
+        return publishEndpoint.Publish(
+            (object)evt,
```
Dragemil: Why cast here?

test/Domain/LeanCode.DomainModels.MassTransitRelay.Tests/Integration/MassTransitIntegrationTest.cs
Dragemil, line 1: A lil bit of FluentAssertions lobbying here

test/Domain/LeanCode.DomainModels.MassTransitRelay.Tests/Integration/TestApp.cs
```diff
@@ -103,46 +93,42 @@ public async Task InitializeAsync()
...
-                            rcv.UseLogsCorrelation();
-                            rcv.UseRetry(retryConfig => retryConfig.Immediate(5));
-                            rcv.UseConsumedMessagesFiltering(ctx);
-                            rcv.StoreAndPublishDomainEvents(ctx);
-
-                            rcv.ConfigureConsumers(ctx);
-                            rcv.ConnectReceiveEndpointObservers(ctx);
-                        }
-                    );
-
```
Dragemil, line 145: ❤️

test/Domain/LeanCode.DomainModels.MassTransitRelay.Tests/Integration/TestApp.cs
```diff
@@ -103,46 +93,42 @@ public async Task InitializeAsync()
...
+                    busCfg.ConfigureEndpoints(ctx);
                     busCfg.ConnectBusObservers(ctx);
```
Dragemil, line 132: Sooo what do we observe here?

Dragemil (approved):

test/Domain/LeanCode.DomainModels.MassTransitRelay.Tests/Integration/TestApp.cs
marcin-dardzinski: Well, the bus.
This registers all the bus observers registered in the container - in case of this test, nothing.
Btw @jakubfijalkowski I remember this was introduced to facilitate MT health checks, but it was a long time ago, maybe we should dump it entirely?

src/Domain/LeanCode.DomainModels.MassTransitRelay/Middleware/EventsPublisherElement.cs
marcin-dardzinski: Without a cast you'll get `Publish<IDomainEvent>` overload which will publish the event as `IDomainEvent` - not the one you expect.

Object overload will use a runtime type - though a comment would be nice.

test/Domain/LeanCode.DomainModels.MassTransitRelay.Tests/Integration/MassTransitIntegrationTest.cs
marcin-dardzinski: If it gets the 'official' blessing I'm all for it.

test/Domain/LeanCode.DomainModels.MassTransitRelay.Tests/Integration/TestApp.cs
jakubfijalkowski: That's perhaps true. I have a faint memory that it was used for something else but I really can't remember what it was.

We should have a full test that runs real MT w/o any test harnesses. Preferably with ASB or at least RabbitMQ.

jakubfijalkowski (approved):

PR #479, opened 2023-05-16
Add `LeanCode.Kratos` package

Saancreed:
Some properties of Kratos sessions are mapped to OIDC–inspired claims in case the application would like to consume them. I found some use for `auth_time` at least which can be used to refuse processing a dangerous request if the session is older than X minutes, forcing the user to reauthenticate.

~~`KratosConfig` contains a property for webhook secret, even though webhooks are not part of this package. There isn't really any code I could put here to make writing wehbook handlers easier because their content is fully customizable, but it's very likely that we will have a generic after-registration handler soon that propagates new identities to another user store.~~

~~Minimal integration looks like this:~~ Not applicable anymore, see https://github.com/leancodepl/corelibrary/pull/479#issuecomment-1552931958

~~1. Register `KratosConfig` with appropriate endpoint URL(s)~~
~~2. Add `KratosModule` to Startup~~
~~3. Configure authentication in `ConfigureServices`~~

But perhaps we could add an option to always map user's schema ID into some (possibly equal to `RoleClaimType`) claim, making `ClaimsExtractor` optional in the process?


lukaszgarstecki (approved):
I believe that there will be value of moving the description of this PR to some docs or readme of the package you introduce

src/Infrastructure/LeanCode.Kratos/KratosAuthenticationHandler.cs
```diff
@@ -0,0 +1,168 @@
...
+                    Value: v,
+                    Name: (
+                        typeof(MethodEnum)
+                            .GetField(v.ToString(), BindingFlags.Public | BindingFlags.Static)
+                            ?.GetCustomAttribute(typeof(EnumMemberAttribute)) as EnumMemberAttribute
+                    )?.Value
+                )
+        )
+        .Where(t => t.Name is not null)
+        .ToImmutableDictionary(t => t.Value, t => t.Name!);
```
jakubfijalkowski: Let's wrap this in a static method, will be easier to read.

src/Infrastructure/LeanCode.Kratos/KratosConfig.cs
```diff
@@ -0,0 +1,10 @@
+namespace LeanCode.Kratos;
+
+public sealed record class KratosConfig(
+    string AdminEndpoint,
```
jakubfijalkowski: `Uri`s?

jakubfijalkowski (commented):
Tests, tests, tests :)

src/Infrastructure/LeanCode.Kratos/KratosModule.cs
```diff
@@ -0,0 +1,16 @@
...
+using LeanCode.Components;
+using Ory.Kratos.Client.Api;
+
+namespace LeanCode.Kratos;
+
+public class KratosModule : AppModule
+{
+    protected override void Load(ContainerBuilder builder)
+    {
+        builder.Register(c => new CourierApi(c.Resolve<KratosConfig>().AdminEndpoint)).AsSelf();
```
marcin-dardzinski: We're are going to ditch Autofac anyway, maybe just and `IServiceCollection.AddKratos` extension method?

src/Infrastructure/LeanCode.Kratos/KratosConfig.cs
Saancreed: Fair, I went with `string`s because this is what `*Api` clients accept but I can change that.

src/Infrastructure/LeanCode.Kratos/KratosModule.cs
Saancreed: Sure, I can try that. This should also allow me to drop the `LeanCode.Components` reference.

src/Infrastructure/LeanCode.Kratos/KratosConfig.cs
Saancreed: Actually we can have both if we dump the `Config` class entirely :upside_down_face:

Saancreed:
v2:

* API Clients are now registered and accessed using their interface types
* DI registration is now done using extension methods on `IServiceCollection`
* `Config` and `Module` classes are now gone
* No longer depends on `LeanCode.Components` / `Autofac`
* Enum and timestamp helpers have been refactored into methods
* Properties of `Options` class are no longer nullable
* Some tests were added

Minimal integration now looks like this:

```csharp
services
    .AddKratosClients(cfg =>
    {
        cfg.AddFrontendApiClient("http://kratos-svc.kratos.svc.cluster.local");
        // cfg.AddIdentityApiClient("http://kratos-svc.kratos.svc.cluster.local:4434"); if desired
    })
    .AddAuthentication()
    .AddKratos(cfg =>
    {
        cfg.RoleClaimType = "role";
        cfg.NameClaimType = "sub";

        cfg.ClaimsExtractor = (s, o, c) =>
        {
            switch (s.Identity.SchemaId)
            {
                case "admin":
                    c.Add(new(o.RoleClaimType, "admin"));
                    goto case "user";
                case "user":
                    c.Add(new(o.RoleClaimType, "user"));
                    break;
            }
        };
    });
```

lukaszgarstecki:
I'd say the tests readability could benefit from FluentAssertions

Saancreed:
> I'd say the tests readability could benefit from FluentAssertions

One of them at most, and I see no reason to bring FluentAssertions just for that. Besides, I consider its syntax to be far too awkward to use. But feel free to suggest another alternative if you have any ideas on _how_ to make it more readable, preferably without using yet another library.

Dragemil:
> > I'd say the tests readability could benefit from FluentAssertions
> 
> One of them at most, and I see no reason to bring FluentAssertions just for that. Besides, I consider its syntax to be far too awkward to use. But feel free to suggest another alternative if you have any ideas on _how_ to make it more readable, preferably without using yet another library.

Readability is a big thing, but you also get a very thorough error messages when you encounter them.

src/Infrastructure/LeanCode.Kratos/KratosAuthenticationHandler.cs
```diff
@@ -0,0 +1,155 @@
...
+
+namespace LeanCode.Kratos;
+
+public class KratosAuthenticationHandler<TOptions> : AuthenticationHandler<TOptions>
+    where TOptions : KratosAuthenticationOptions, new()
+{
+    private static string ToUnixTimeSecondsString(DateTime value) =>
+        new DateTimeOffset(value).ToUnixTimeSeconds().ToString(CultureInfo.InvariantCulture);
+
+    private static ImmutableDictionary<T, string> ExtractEnumNames<T>()
```
jakubfijalkowski: Leeet's move this down the file. It's not that important to be this high.

src/Infrastructure/LeanCode.Kratos/KratosAuthenticationHandler.cs
```diff
@@ -0,0 +1,155 @@
...
+            )
+            {
+                session = await api.ToSessionAsync(xSessionToken: value.Parameter);
+            }
+            else if (
+                Request.Cookies.TryGetValue(Options.SessionCookieName, out var cookie) && !string.IsNullOrEmpty(cookie)
+            )
+            {
+                session = await api.ToSessionAsync(cookie: $"{Options.SessionCookieName}={cookie}");
+            }
```
jakubfijalkowski: Let's do a quick refactoring of this method - extract this block into a separate method. Will be easier to read, plus you would be able to get rid of the dangling variable.

src/Infrastructure/LeanCode.Kratos/KratosAuthenticationHandler.cs
```diff
@@ -0,0 +1,155 @@
...
+
+            foreach (var am in session.AuthenticationMethods)
+            {
+                if (am.Method is { } method && AuthenticationMethods.TryGetValue(method, out var methodName))
+                {
+                    claims.Add(new("amr", methodName));
+                }
+            }
+
+            Options.ClaimsExtractor(session, Options, claims);
```
jakubfijalkowski: Let's extract this into sth like `ExtractClaims`?

jakubfijalkowski (approved):
Last ones, and then we're good to go.

Saancreed:
@jakubfijalkowski All done. I made an attempt at also extracting session validation but it's hard to convince Roslyn that some parameter to a method won't be `null` if the call returns `null` :weary:

Saancreed:
I also added missing `DynamicallyAccessedMembers`, forgot to enable analyzer for it :see_no_evil:

marcin-dardzinski (approved):

PR #480, opened 2023-05-17
Docs/creating an aggregate

FrackiewiczP:
So, the idea behind this (and PRs yet to come) is to create a really basic app while describing how to implement some common scenarios (see `How to: ...` tasks in Linear)  along the way, providing the user with a working example and explanations on how to use the CoreLibrary features to build an app.


lukaszgarstecki:
Shouldn't this PR target `v8.0` branch?

docs/guides/01_creating_an_aggregate.md
```diff
@@ -0,0 +1,190 @@
...
+        return new Project
+        {
+            Id = SId<Project>.New(),
+            Name = name,
+        };
+    }
+}
+```
+
+As you can see, the class implements `IAggregateRoot` interface - it marks the class as being the root of an aggregate. Moreover the `Id` field of the class is of type `SId<Project>` - it is a special type present in the core library which is basically a Guid prefixed with a class name. In this case, the Id of the Project will look somewhat like `project_45a8f39f-9df0-4a23-b781-2a46de22fac1`.
```
Dragemil: We should have a unified reference to the package library.

```suggestion
As you can see, the class implements `IAggregateRoot` interface - it marks the class as being the root of an aggregate. Moreover the `Id` field of the class is of type `SId<Project>` - it is a special type present in the CoreLibrary which is basically a Guid prefixed with a class name. In this case, the Id of the Project will look somewhat like `project_45a8f39f-9df0-4a23-b781-2a46de22fac1`.
```

docs/guides/01_creating_an_aggregate.md
```diff
@@ -0,0 +1,190 @@
...
+The `Project` also has a list of `Tasks`. Notice that there are two lists containing tasks of a project - the `Tasks` one is a readonly interface for the `tasks` which contents can be changed by the class.
```
lukaszgarstecki: I'd stress that we design the API in a way that we cannot manipulate the aggregate from the outside world (only object can modify it's state based on the method it exposes)

docs/guides/01_creating_an_aggregate.md
```diff
@@ -0,0 +1,190 @@
...
+        NotStarted,
+        InProgress,
+        Finished,
+    }
+}
+```
+
+Notice that a task is not an aggregate root - it can only be accessed through the project to which it belongs.
+
+Likewise, let's create a class representing an user which can be assigned to a task:
```
Dragemil: It's an exception.

```suggestion
Likewise, let's create a class representing a user which can be assigned to a task:
```

docs/guides/01_creating_an_aggregate.md
```diff
@@ -0,0 +1,190 @@
...
+        };
+    }
+}
+```
+
+Notice that the `User` class is an aggregate root - the user represents a standalone business domain object with its own meaning and behavior.
+
+## Adding logic
+
+So far, only the structure of objects has been defined but business domain objects usually has some set of behaviors which they can perform. In this section we will see how a logic can be added to the aggregates.
```
Dragemil: _perform behaviors_ sounds strange?

docs/guides/01_creating_an_aggregate.md
```diff
@@ -0,0 +1,190 @@
...
+Remember that the only way to access a task is to go through the relevant project which is the aggregate root. Because of this we will add some methods to the `Project` class which will allow us to interact with the tasks:
+
+```csharp
+public class Project : IIdentifiable<SId<Project>>
+{
+    . . .
+
+    public void EditTask(SId<Task> taskId, string name)
+    {
+        tasks.Single(t => t.Id == taskId).Edit(name);
```
Dragemil, line 149: I don't ever use `.Single(...)` in my projects in favor of `.First(...)`, is it a mistake by me, here or it doesn't matter?

docs/guides/01_creating_an_aggregate.md
```diff
@@ -0,0 +1,190 @@
...
+        tasks.Single(t => t.Id == taskId).ChangeTaskStatus(status);
+    }
+}
+```
+
+Notice that these added methods can and will throw an exception if a project does not contain any task with provided Id. This is an excepted behavior - checks for respecting domain logic should be performed in respective command (or operations) validators.
+
+## Rasing domain events
+
+The domain events make an important part of DDD. It is through the use of domain events that an aggregate might communicate with the rest of the system. We will see how to raise an event from the aggregate.
```
Dragemil:
```suggestion
The domain events make an important part of DDD. It is through the use of domain events that an aggregate may communicate with the rest of the system. In this section we will see how to raise an event from an aggregate.
```

docs/guides/01_creating_an_aggregate.md
```diff
@@ -0,0 +1,190 @@
...
+
+```csharp
+public class Project : IIdentifiable<SId<Project>>
+{
+    . . .
+
+    public void AssignUserToTask(SId<Task> taskId, SId<User> userId)
+    {
+        tasks.Single(t => t.Id == taskId).AssignUser(userId);
+        DomainEvents.Raise(new UserAssignedToTask(SId<Task> taskId, SId<User> userId));
```
Dragemil, line 185: Are we okay with raising domain events not from aggregate roots?

Dragemil (commented):
Nice and friendly walkthrough.

I think it would be nice if the final examples here were the examples from the fully fledged sample CoreLib app. Also it would be nice that if the app wouldn't build anymore in the CI due to CoreLib updates which it references than that could trigger the need to update the docs.

Other thing is maybe we would like to include some references to those basic DDD concepts we agree with?

docs/guides/01_creating_an_aggregate.md
FrackiewiczP: Thanks! Did not remember about it

docs/guides/01_creating_an_aggregate.md
jakubfijalkowski: Yes, at least that is what I did for the last couple of years ^^'

docs/guides/01_creating_an_aggregate.md
FrackiewiczP: I would say it doesn't matter that much. I find `Single(...)` more meaningful and in this this case it makes it less likely for two tasks with the same `Id` to be left unnoticed. But of course it is very unlikely in the first place.

docs/guides/01_creating_an_aggregate.md
FrackiewiczP: This is in `Project` but you are right, I marked some `Project`s as aggregate roots and others as `IIdentifiable` - all should be aggregate roots.

FrackiewiczP:
@Dragemil Added a story to Linear for a complete sample CoreLib app with CI configuration. Also added a note to the top level readme about DDD

docs/README.md
```diff
@@ -21,3 +21,10 @@ The CoreLib documentation is available here:
  3. [Benchmarks](./benchmarks/README.md).
  4. [Domain](./domain/README.md).
  5. [Architecture decision records](./adrs/README.md).
+
+## Domain Driven Design
+
+LeanCode Core Library is strongly based on concepts of Domain Driven Design. If you are not familiar with this approach to developing software please check out the following books:
```
jakubfijalkowski:
```suggestion
LeanCode Core Library is strongly based on concepts of Domain Driven Design. If you are not familiar with this approach to developing software, you can check these books:
```

less aggressive ;P

docs/guides/01_creating_an_aggregate.md
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    public void AssignUser(SId<User> userId)
+    {
+        AssignedUser = userId;
+    }
+
+    public void UnassignUser()
+    {
+        AssignUser = null;
```
lukaszgarstecki:
```suggestion
        AssignedUser = null;
```

docs/guides/src/Project.cs
lukaszgarstecki, line 1: Don't we want to add `namespace` and `using`'s here (and to other samples)?

lukaszgarstecki (approved):
Some NITs, other than that LGTM

docs/guides/src/Project.cs
FrackiewiczP: We want. In these first two PRs I focused on the documentation part. The third PR (in progress) is about creating a minimal app and will incorporate the samples in the app.

PR #481, opened 2023-05-18
Docs/writing commands

FrackiewiczP:


FrackiewiczP:
Next PR, I'll create a minimal app where these could be invoked.

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,233 @@
+# Writing commands
+
+## Intro
+
+Interacting with systems based with Core Library is based on a Command Query Responsibility Segregation (CQRS) principle. As such, there are two main types of actions which can be performed in the system: command, and query [^1]. In this guide we will focus on the former. A command is used the change the state of the system but it does not yield any results - it can be treated as write-only action.
```
lukaszgarstecki:
```suggestion
Interacting with systems based with Core Library is based on a Command Query Responsibility Segregation (CQRS) principle. As such, there are two main types of actions which can be performed in the system: command, and query [^1]. In this guide we will focus on the former. A command is used the change the state of the system but it does not yield any results - it can be treated as write-only action.
```
The `query` could be linked to the next document.

lukaszgarstecki: I'd say we can link some further reading docs about CQRS pattern like we did with DDD

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,233 @@
...
+```
+
+A validator for a command should inherit from `ContextualValidator<T>` where `T` is a type of command contract. The validation in CoreLibrary builds upon [FluentValidation](https://docs.fluentvalidation.net/en/latest/), adding `WithCode()` method which is responsible for adding an error code to the response in case the validation fails. In the above example, trying to change a project name to one which is over 500 characters long would cause a command to fail and return an appropriate error code to the client.
+
+## Command handler
+
+> **Info :information_source:**
+> A command validator and handler are usually put in a single file with a `CH` suffix, e.g. `CreateProjectCH.cs`.
+
+Finally, we arrive at a command handler which is a part which actually executes the command. For the `CreateProject` command, a handler may look like this:
```
lukaszgarstecki:
```suggestion
Finally, we arrive at a command handler which is a part that actually executes the command. For the `CreateProject` command, a handler may look like this:
```

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,233 @@
...
+The command for creating a project is a really basic one - it only takes a name for new project and all of the validation can be done without checking the state of the domain. We will now show a slightly more complex command which will be responsible for adding tasks to existing projects.
+
+```csharp
+[AllowUnauthorized]
+public class AddTasksToProject : ICommand
+{
+    public string ProjectId { get; set; }
+    public List<TaskDTO> Tasks { get; set; }
+
+    public static class ErrorCodes
```
lukaszgarstecki, line 132: Shouldn't we include `TaskDTO.ErrorCodes` here?

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,233 @@
...
+        RuleForEach(cmd => cmd.Tasks)
+            .ChildRules(child => child.RuleFor(c => c).SetValidator(new TaskDTOValidator()));
+
+        RuleForAsync(cmd => cmd, CheckProjectExistsAsync)
+            .Equal(true)
+            .WithCode(AddContactToSite.ErrorCodes.ProjectDoesNotExist)
+            .WithMessage("A project with given Id does not exist.");
+    }
+
+    private Task<ValidationResult> CheckProjectExistsAsync(
```
lukaszgarstecki:
```suggestion
    private Task<bool> CheckProjectExistsAsync(
```

docs/guides/03_writing_queries.md
```diff
@@ -0,0 +1,68 @@
+# Writing queries
+
+## Intro
+
+In the previous guide, we described how to write commands - actions which modify the data in the system. This guide shows how queries can be written. Contrary to commands, queries are responsible for retrieving data without modifying any objects.
```
lukaszgarstecki: We can add here that queries should be considered read only

docs/guides/03_writing_queries.md
```diff
@@ -0,0 +1,68 @@
...
+
+Each query consists of two parts:
+
+* Contract
+* Query handler
+
+This is analogous to how commands are implemented, except a validator.
```
lukaszgarstecki: link to writing command docs?

docs/guides/03_writing_queries.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Contract
+
+As it is the case with command contracts, a query contract defines the parameters passed to a query as well as the structure of the response. We will show a simple query for retrieving a list of projects.
+
+```csharp
+[AllowUnauthorized]
+public class AllProjects : IQuery<List<ProjectDTO>>
+{
+    public bool SortByDescending { get; set; }
```
lukaszgarstecki:
```suggestion
    public bool SortDescending { get; set; }
```

lukaszgarstecki: or `SortByNameDescending`

docs/guides/03_writing_queries.md
```diff
@@ -0,0 +1,68 @@
...
+public class ProjectDTO
+{
+    public string Id { get; set; }
+    public string Name { get; set; }
+}
+```
+
+A contract for a query is defined by creating a class implementing the `IQuery<T>` interface. The `T` is the result type of a query which is serialized in the HTTP response after a query has been executed. Also, web and mobile clients can use the Contracts Generator to generate classes corresponding to defined queries and their result types.
+
+Fields in a query class represent parameters which can be passed in a request to alter the query results. In the above example there is a single query parameter of boolean type `SortByDescending` which will be used to determine sorting order of the results.
```
lukaszgarstecki:
```suggestion
Fields in a query class represent parameters which can be passed in a request to alter the query results. In the above example there is a single query parameter of boolean type `SortDescending` which will be used to determine sorting order of the results.
```

docs/guides/03_writing_queries.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Handler
+
+A handler for a command is a class implementing `IQueryHandler<TContext, TQuery, TResult>` interface where `TQuery` is the contract type and `TResult` is the result type as defined in the contract. In our example, a query handler might look like this:
```
lukaszgarstecki, line 38: We silently ignore what `TContext` is in here (and in command docs). I believe they should be explained at least briefly

docs/guides/03_writing_queries.md
```diff
@@ -0,0 +1,68 @@
...
+
+    public async Task<List<ProjectDTO>> ExecuteAsync(CoreContext context, AllProjects query)
+    {
+        var q = dbContext.Projects.Select(p => new ProjectDTO
+        {
+            Id = p.Id,
+            Name = p.Name,
+        });
+
+        q = query.SortByDescending ? q.OrderByDescending(p => p.Name) : q.OrderBy(p => p.Name);
```
lukaszgarstecki:
```suggestion
        q = query.SortDescending ? q.OrderByDescending(p => p.Name) : q.OrderBy(p => p.Name);
```

docs/guides/03_writing_queries.md
```diff
@@ -0,0 +1,68 @@
...
+
+        return q.ToListAsync(context.CancellationToken);
+    }
+}
+```
+
+The `dbContext` used to query the database for data is passed in the constructor to be injected via Dependency Injection. Also, notice how `SortByDescending` parameter from the query is used to determine the sorting order of the result.
```
lukaszgarstecki:
```suggestion
The `dbContext` used to query the database for data is passed in the constructor to be injected via Dependency Injection. Also, notice how `SortDescending` parameter from the query is used to determine the sorting order of the result.
```

docs/guides/src/Domain/Project.cs
```diff
@@ -16,6 +16,11 @@ public static Project Create(string name)
         return new Project { Id = SId<Project>.New(), Name = name, };
     }
 
+    public void AddTasks(IEnumerable<Task> tasks)
```
lukaszgarstecki: Considering the fact that `Task` is identifiable and has a reference to `Project` I'd say we should not allow the `Task` creation outside the `Project` aggregate. 

If `Task` would be `ValueObject` the approach with passing it from service level is fine.

docs/guides/03_writing_queries.md
jakubfijalkowski: That might not be necessary - this _might_ change soon.

docs/guides/02_writing_commands.md
FrackiewiczP: I am not sure, I don't include such error codes in the command's error codes. Is it a mistake?

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,233 @@
+# Writing commands
+
+## Intro
+
+Interacting with systems based with Core Library is based on a Command Query Responsibility Segregation ([CQRS](../basics/02_cqrs.md)) principle. As such, there are two main types of actions which can be performed in the system: command, and [query](./03_writing_queries.md) [^1]. In this guide we will focus on the former. A command is used the change the state of the system but it does not yield any results - it can be treated as write-only action.
```
jakubfijalkowski:
```suggestion
Interacting with systems that are based on Core Library follows the Command Query Responsibility Segregation ([CQRS](../basics/02_cqrs.md)) principle. As such, there are two main types of actions which can be performed in the system: commands, and [queries](./03_writing_queries.md) [^1]. In this guide we will focus on the former. A command is used to change the state of the system but it does not yield any results - it can be treated as write-only action.
```

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,233 @@
...
+
+Each command consists of three parts:
+
+* Contract
+* Command validator
```
jakubfijalkowski:
```suggestion
* (Optional) Command validator
```

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,233 @@
...
+
+    public static class ErrorCodes
+    {
+        public const int NameCannotBeEmpty = 1;
+        public const int NameTooLong = 2;
+    }
+}
+```
+
+The contract for a command is simply a class implementing the `ICommand` interface. The fields of a contract class define how the body of an HTTP request should look like. In this example, an example of a valid request body would be:
```
jakubfijalkowski:
```suggestion
The contract for a command is simply a class implementing the `ICommand` interface. The fields of a contract class define how the body of an HTTP request should look like. In this case, an example of a valid request body would be:
```

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,233 @@
...
+{
+    "Name": "abc",
+}
+```
+
+Notice the `ErrorCodes` class. These define possible validation errors which can be expected by the clients. Each error code should be a unique (in a given contract) integer.
+
+Writing the contracts in that way makes it possible for web and mobile clients to generate their own version of contract classes with use of the Contract Generator which is also a part of CoreLibrary.
+
+Also, a contract can have authorization attributes which define rules which have to be satisfied for an actor to be able to invoke the command. To learn more about authorizers, see [authorization](./0X_authorization.md).
```
jakubfijalkowski: We should also note somewhere that authorization is an extension to the contracts and not an integral part of the contracts system. There can also be different extensions (we had, for example, a `CacheAttribute` at one point in time).

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Command validator
+
+Command validator's main role is to ensure that the command invoked by the user is valid. The command validator should check the structure of the command itself, for example, check for required fields, check whether passed strings do not exceed maximum length etc. Also the validator should check the business logic of the command for example, when editing a project, a command validator should check whether or not the project exists, see [more complex commands](#more-complex-commands) for an example. Let's see how a validator for `CreateProject` command might look like:
```
jakubfijalkowski:
```suggestion
The command validator's main role is to ensure that the command invoked by the user is valid. The command validator should check the structure of the command itself, for example, check for required fields, check whether passed strings do not exceed maximum length etc. Also the validator should check the business requirements of the command. For example, when editing a project, it should check whether or not the project exists, see [more complex commands](#more-complex-commands) for an example. Let's see how a validator for `CreateProject` command might look like:
```

We should also add a single sentence of "why" we split validation and execution.

Also, something to add to backlog - the guidelines of writing business logic validators.

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,233 @@
...
+        RuleFor(cmd => cmd.Name)
+            .NotEmpty()
+            .WithCode(CreateProject.ErrorCodes.NameCannotBeEmpty)
+            .MaximumLength(500)
+            .WithCode(CreateProject.ErrorCodes.NameTooLong);
+    }
+}
+```
+
+A validator for a command should inherit from `ContextualValidator<T>` where `T` is a type of command contract. The validation in CoreLibrary builds upon [FluentValidation](https://docs.fluentvalidation.net/en/latest/), adding `WithCode()` method which is responsible for adding an error code to the response in case the validation fails. In the above example, trying to change a project name to one which is over 500 characters long would cause a command to fail and return an appropriate error code to the client.
```
jakubfijalkowski: Let's remove the note about `ContextualValidator` - `ContextualValidator` is a helper that builds on top of FluentValidation for `RuleForAsync` (that is being deprecated, really), but is not necessary to use with our validation. You can (and should) use normal `Validator`s.

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Command handler
+
+> **Info :information_source:**
+> A command validator and handler are usually put in a single file with a `CH` suffix, e.g. `CreateProjectCH.cs`.
+
+Finally, we arrive at a command handler that is a part which actually executes the command. For the `CreateProject` command, a handler may look like this:
```
jakubfijalkowski:
```suggestion
Finally, we arrive at a command handler that is the part which actually executes the command. For the `CreateProject` command, a handler may look like this:
```

docs/guides/02_writing_commands.md
jakubfijalkowski: Not really a mistake, but having that might help clients (they will have all error codes in a single place then).

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,233 @@
...
+        RuleFor(cmd => cmd.Tasks)
+            .NotNull()
+            .WithCode(AddTasksToProject.ErrorCodes.TasksCannotBeNull)
+            .NotEmpty()
+            .WithCode(AddTasksToProject.ErrorCodes.TasksCannotBeEmpty);
+
+        RuleForEach(cmd => cmd.Tasks)
+            .ChildRules(child => child.RuleFor(c => c).SetValidator(new TaskDTOValidator()));
+
+        RuleForAsync(cmd => cmd, CheckProjectExistsAsync)
```
jakubfijalkowski: Okay, this will be "slight" breaking change here - instead of `RuleForAsync`, let's use `MustAsync`. It will be more natural and should be _the_ way of validating a single thing.

Also, `RuleForAsync` will probably be deprecated soon, as we don't want to stay on old version of FV (and we can't implement that in newer version).

docs/guides/02_writing_commands.md
Dragemil: > does not yield any results

Other then whether it succeeded - do we want to mention that here?

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,233 @@
...
+
+```json
+{
+    "Name": "abc",
+}
+```
+
+Notice the `ErrorCodes` class. These define possible validation errors which can be expected by the clients. Each error code should be a unique (in a given contract) integer.
+
+Writing the contracts in that way makes it possible for web and mobile clients to generate their own version of contract classes with use of the Contract Generator which is also a part of CoreLibrary.
```
Dragemil:
```suggestion
Writing contracts in that way makes it possible for web and mobile clients to generate their own version of contract classes with use of the Contract Generator which is also a part of the CoreLibrary.
```

docs/guides/02_writing_commands.md
Dragemil: Let's treat that as a suggestion to Fiołeks suggestion.

The first `the` doesn't seem right since we do not reference any particular validator. The other ones I didn't check anywhere as well, they're based on whether they feels proper to me, so surely they are debatable.

```suggestion
Command validator's main role is to ensure that a command invoked by a client is valid. Command validator should check the structure of the command itself, for example, check for required fields, check whether passed strings do not exceed maximum length, etc. Also the validator should check business requirements of the command. For example, when editing a project, it should check whether or not the project exists, see [more complex commands](#more-complex-commands) for an example. Let's see how a validator for `CreateProject` command might look like:
```

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,233 @@
...
+
+        logger.Information("Project {ProjectId} added", project.Id);
+    }
+}
+```
+
+> **Warning :warning:**
+> According to DDD rules, a single command should only modify a single aggregate as an aggregate should be a transaction boundary in the system. If there is a need to modify multiple aggregates as a part of a single operation, raise events and modify all aggregates one by one in event handlers.
+
+The logic of the handler is straightforward - a new project with a given name is created and added to `projects` repository. The `IRepository` interface is a part of Core Library which allows to abstract the logic of persisting domain entities. A repository (and other services) should be injected into a command handler using dependency injection. Also, notice that the `Add` method used to persist newly created project is synchronous. The actual asynchronous operation in the database is performed by `StoreAndPublishEvents()` element in the pipeline, see [pipeline](./0X_pipeline.md) for more info. A basic repository for projects allowing for operations on database may look like this:
```
Dragemil: I'd have a couple of things here:

1. AFAIK we use repositories not really for abstracting anything (the`DbSet`s are basically the same abstraction), but rather for avoiding dependency on project's `DbContext`, should one want to unit test the command and mock the dependencies (mocking `DbContext`s is not a very fun developer experience),
2. I'd rephrase why we use `Add` - yes it's synchronous, but most importantly it just marks entities to add to the DB and doesn't commit anything yet, which opens us to have transactional outbox (some reference here probably in the docs) as a part of the CQRS pipeline.

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,233 @@
...
+        : base(dbContext) { }
+
+    public override Task<Purchase?> FindAsync(SId<Purchase> id, CancellationToken cancellationToken = default)
+    {
+        return DbSet.AsTracking().FirstOrDefaultAsync(p => p.Id == id, cancellationToken)!;
+    }
+}
+```
+
+Here `EFRepository` is a class from Core Library containing default implementations for `Add`, `Delete` and `Update` methods which can be used to alter the contents of the database. Usually `FindAsync` method can be implemented as a simple retrieval of an entity based on Id but more complicated operations can also be done in the method if they are needed.
```
Dragemil: We basically want to have `the` probably everywhere when we mention the Core Library.

Dragemil: A few words why `FindAsync` is implemented here (and typically) without `DbSet`'s `FindAsync`?

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,233 @@
...
+
+    public class ErrorCodes
+    {
+        public const int NameCannotBeEmpty = 101;
+        public const int NameTooLong = 102; 
+    }
+}
+```
+
+Notice the use of `TaskDTO` class. DTO classes can be defined as part of contracts and matching classes will be generated for web and mobile clients. Also, notice that in `TaskDTO` we used error codes starting from 101 to avoid clashes with error codes in `AddTasksToProject`. Now, let's move onto validating the command:
```
Dragemil: A note in advance that ensuring uniqueness of shared error codes wasn't ever a problem for us so we don't have any sophisticated mechanism for that?

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,233 @@
...
+    private readonly IRepository<Project, SId<Project>> projects;
+
+    public AddTasksToProjectCH(IRepository<Project, SId<Project>> projects)
+    {
+        this.projects = projects;
+    }
+
+    public async Task ExecuteAsync(CoreContext context, AddTasksToProject command)
+    {
+        var project = await projects.FindAndEnsureExistsAsync(command.ProjectId, context.CancellationToken);
```
Dragemil, line 228: `FindAndEnsureExistsAsync` wasn't mentioned anywhere yet but probably is pretty self explanatory.

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,239 @@
...
+
+Interacting with systems that are based on the CoreLibrary follows the Command Query Responsibility Segregation ([CQRS](../basics/02_cqrs.md)) principle. As such, there are two main types of actions which can be performed in the system: commands, and [queries](./03_writing_queries.md) [^1]. In this guide we will focus on the former. A command is used to change the state of the system but it does not yield any results other then whether it succeeded - it can be treated as write-only action.
+
+Each command consists of three parts:
+
+* Contract
+* (Optional) Command validator
+* Command handler
+
+The validation is performed separately from the execution of the command. The reason is adherence to the single responsibility principle. Moreover, handler implementation is made easier, both to write and read, due to the fact that it contains only the actual business process.
```
mishioo: Maybe we could point out that this separation also leads to assumptions about data received being transferred implicitly from validator to handler (i.e. that compiler won't intercept that e.g. something is not `null`).

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,239 @@
...
+
+```json
+{
+    "Name": "abc",
+}
+```
+
+Notice the `ErrorCodes` class. These define possible validation errors which can be expected by the clients. Each error code should be a unique (in a given contract) integer.
+
+Writing contracts in that way makes it possible for web and mobile clients to generate their own version of contract classes with use of the Contract Generator which is also a part of the CoreLibrary.
```
mishioo:
```suggestion
Writing contracts in that way makes it possible for web and mobile clients to generate their own version of contract classes with use of the Contract Generator, which is also a part of the CoreLibrary.
```

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,239 @@
...
+
+Also, a contract can have authorization attributes which define rules which have to be satisfied for an actor to be able to invoke the command. Authorization, however, is an extension to the contracts systme and not an integral part of it. To learn more about authorizers, see [authorization](./0X_authorization.md).
```
mishioo:
```suggestion
Also, a contract can have authorization attributes, which define rules that have to be satisfied for an actor to be able to invoke the command. Authorization, however, is an extension to the contracts system and not an integral part of it. To learn more about authorizers, see [authorization](./0X_authorization.md).
```

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,239 @@
...
+        RuleFor(cmd => cmd.Name)
+            .NotEmpty()
+            .WithCode(CreateProject.ErrorCodes.NameCannotBeEmpty)
+            .MaximumLength(500)
+            .WithCode(CreateProject.ErrorCodes.NameTooLong);
+    }
+}
+```
+
+A validator for a command should inherit from `Validator<T>` where `T` is a type of command contract. The validation in the CoreLibrary builds upon [FluentValidation](https://docs.fluentvalidation.net/en/latest/), adding `WithCode()` method which is responsible for adding an error code to the response in case the validation fails. In the above example, trying to change a project name to one which is over 500 characters long would cause a command to fail and return an appropriate error code to the client.
```
mishioo:
```suggestion
A validator for a command should inherit from `Validator<T>` where `T` is a type of command contract. The validation in the CoreLibrary builds upon [FluentValidation](https://docs.fluentvalidation.net/en/latest/), adding `WithCode()` method, which is responsible for adding an error code to the response in case the validation fails. In the above example, trying to change a project name to one which is over 500 characters long would cause a command to fail and return an appropriate error code to the client.
```

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,239 @@
...
+
+The logic of the handler is straightforward - a new project with a given name is created and added to `projects` repository. The `IRepository` interface is a part of the CoreLibrary which allows to abstract the logic of persisting domain entities and to avoid dependency on the project's `DbContext` which allows for easier unit testing without having to mock the `DbContext`. A repository (and other services) should be injected into a command handler using dependency injection. Also, notice that the `Add` method used to persist newly created project is synchronous. It doesn't commit anything to the database which makes it possible to use transactional outbox pattern. The actual asynchronous operation in the database is performed by `StoreAndPublishEvents()` element in the pipeline, see [pipeline](./0X_pipeline.md) for more info. A basic repository for projects allowing for operations on database may look like this:
```
mishioo:
```suggestion
The logic of the handler is straightforward — a new project with a given name is created and added to `projects` repository. The `IRepository` interface is a part of the CoreLibrary which allows to abstract the logic of persisting domain entities and to avoid dependency on the project's `DbContext` which allows for easier unit testing without having to mock the `DbContext`. A repository (and other services) should be injected into a command handler using dependency injection. Also, notice that the `Add` method used to persist newly created project is synchronous. It doesn't commit anything to the database which makes it possible to use transactional outbox pattern. The actual asynchronous operation in the database is performed by `StoreAndPublishEvents()` element in the pipeline, see [pipeline](./0X_pipeline.md) for more info. A basic repository for projects allowing for operations on database may look like this:
```

mishioo: Also it would be nice to have a citation (i.e. a link with explanation) to the outbox pattern.

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,239 @@
...
+        : base(dbContext) { }
+
+    public override Task<Purchase?> FindAsync(SId<Purchase> id, CancellationToken cancellationToken = default)
+    {
+        return DbContext.Projects.FirstOrDefaultAsync(p => p.Id == id, cancellationToken)!;
+    }
+}
+```
+
+Here `EFRepository` is a class from the CoreLibrary containing default implementations for `Add`, `Delete` and `Update` methods which can be used to alter the contents of the database. Usually `FindAsync` method can be implemented as a simple retrieval of an entity based on Id but more complicated operations can also be done in the method if they are needed. Usually we don't wan to use `FindAsync` method of the corresponding `DbSet` to avoid concurrency errors resulting from entities being cached by the `DbSet`.
```
mishioo, line 116: TBH I think that repositories deserve their own chapter as well.

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,239 @@
...
+
+> **Warning :warning:**
+> If an implemented command does not work, make sure that all referenced types are properly registered in the dependency injection container.
+
+## More complex commands
+
+The command for creating a project is a really basic one - it only takes a name for new project and all of the validation can be done without checking the state of the domain. We will now show a slightly more complex command which will be responsible for adding tasks to existing projects.
```
mishioo:
```suggestion
The command for creating a project is a really basic one - it only takes a name for new project and all of the validation can be done without checking the state of the domain. We will now show a slightly more complex command, which will be responsible for adding tasks to existing projects.
```

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,239 @@
...
+
+    public class ErrorCodes
+    {
+        public const int NameCannotBeEmpty = 101;
+        public const int NameTooLong = 102; 
+    }
+}
+```
+
+Notice the use of `TaskDTO` class. DTO classes can be defined as part of contracts and matching classes will be generated for web and mobile clients. Also, notice that in `TaskDTO` we used error codes starting from 101 to avoid clashes with error codes in `AddTasksToProject`. Ensuring the uniqueness of error codes never had been a problem, so there does not exist any sophisticated mechanism for it. Now, let's move onto validating the command:
```
mishioo:
```suggestion
Notice the use of `TaskDTO` class. DTO classes can be defined as a part of contracts and matching classes will be generated for web and mobile clients. Also, notice that in `TaskDTO` we used error codes starting from 101 to avoid clashes with error codes in `AddTasksToProject`. Ensuring the uniqueness of error codes never had been a problem, so there does not exist any sophisticated mechanism for it. Now, let's move on to validating the command:
```

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,239 @@
...
+            return false;
+        }
+
+        var dbContext = ctx.ParentContext.GetService<CoreDbContext>();
+
+        return dbContext.Projects.AnyAsync(p => p.Id == pid, cancellationToken);
+    }
+}
+
+public class TaskDTOValidator : AbstractValidator<TaskDTO>
```
mishioo, line 196: I think it's worth explaining why it's `AbstractValidator<T>` and not the previously used `Validator<T>`.

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,239 @@
...
+```csharp
+public class AddTasksToProjectCV : Validator<AddTasksToProject>
+{
+    public AddTasksToProjectCV()
+    {
+        RuleFor(cmd => cmd.Tasks)
+            .NotNull()
+            .WithCode(AddTasksToProject.ErrorCodes.TasksCannotBeNull)
+            .NotEmpty()
+            .WithCode(AddTasksToProject.ErrorCodes.TasksCannotBeEmpty);
```
mishioo, line 166: Are `.WithMessage()` calls omitted intentionally? If so, maybe it's worth a comment.

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,239 @@
...
+        projects.Update(project);
+
+        logger.Information("{TaskCount} tasks added to project {ProjectId}.", command.Tasks.Count, project.Id);
+    }
+}
+```
+
+What happens is as follows: we retrieve the existing project via `projects` repository, add tasks to it and update it in the repository. Again, please notice that the `Update` method is called synchronously - the actual saving of the changes in the database is performed as a part of request pipeline. Also, notice the use of `FindAndEnsureExistsAsync` method. It behaves the same as `FindAsync` method in a repository class when requested object exists and throws an exception if it doesn't.
+
+[^1]: There is also a third type of action called 'operation' which can both modify object and return data. It should only be used out of necessity as it is a violation of CQRS principle.
```
mishioo: I think I'd be more specific about this, stating where the necessity may come from.

mishioo:
```suggestion
[^1]: There is also a third type of action called 'operation', which can both modify object and return data. It should only be used out of necessity as it is a violation of CQRS principle.
```

docs/guides/03_writing_queries.md
```diff
@@ -0,0 +1,68 @@
...
+}
+
+public class ProjectDTO
+{
+    public string Id { get; set; }
+    public string Name { get; set; }
+}
+```
+
+A contract for a query is defined by creating a class implementing the `IQuery<T>` interface. The `T` is the result type of a query which is serialized in the HTTP response after a query has been executed. Also, web and mobile clients can use the Contracts Generator to generate classes corresponding to defined queries and their result types.
```
mishioo: I may be wrong about the "the"s.
```suggestion
A contract for a query is defined by creating a class implementing the `IQuery<T>` interface. The `T` is the result type of the query, which is serialized in the HTTP response after the query has been executed. Also, web and mobile clients can use the Contracts Generator to generate classes corresponding to defined queries and their result types.
```

docs/guides/03_writing_queries.md
```diff
@@ -0,0 +1,68 @@
...
+
+Fields in a query class represent parameters which can be passed in a request to alter the query results. In the above example there is a single query parameter of boolean type `SortByNameDescending` which will be used to determine sorting order of the results.
```
mishioo:
```suggestion
Fields in a query class represent parameters which can be passed in a request to alter the query results. In the above example there is a single query parameter of boolean type `SortByNameDescending`, which will be used to determine sorting order of the results.
```

mishioo: Maybe mention that in a more complex query there could be fields for filtering results and whatnot, to state explicitly that queries are used for, well... all queries.

docs/guides/03_writing_queries.md
```diff
@@ -0,0 +1,68 @@
...
+        });
+
+        q = query.SortByNameDescending ? q.OrderByDescending(p => p.Name) : q.OrderBy(p => p.Name);
+
+        return q.ToListAsync(context.CancellationToken);
+    }
+}
+```
+
+The `dbContext` used to query the database for data is passed in the constructor to be injected via Dependency Injection. Also, notice how `SortByNameDescending` parameter from the query is used to determine the sorting order of the result.
```
mishioo: The example doesn't show it, but maybe it's a good idea to mention that, by convention, we do not throw when someone requests data that doesn't exists and we return an empty/`default` result instead.

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,239 @@
...
+```
+
+There are two important things to note here. First one is the use of a custom validator for validating `TaskDTO`. Writing such validators may allow for reusing the validation logic in commands using the same DTO classes. The second one is performing asynchronous validation - to check whether a project we want to add the tasks for exists, we perform a database query inside the `ProjectExistsAsync` method. This allows to check that the command to be executed does not violate the logic of the business domain.
+
+Now, let's implement the handler for the command:
+
+```csharp
+public class AddTasksToProjectCH : ICommandHandler<CoreContext, AddTasksToProject>
+{
+    private readonly Serilog.ILogger logger = Serilog.Log.ForContext<AddTasksToProjectCH>();
```
mishioo: We have Serilog in examples, but never mention it. I think we should either suggest it as a good practice (or standard, or whatever) or remove it from the examples to keep them minimal.

mishioo (commented):
A minor technical suggestion: let's use one sentence per line in the `.md` files. It's much easier to see a relevant change in the paragraph when working with these files in git.

Also, sorry for comma gatekeeping.

docs/guides/02_writing_commands.md
FrackiewiczP: I agree, but I wouldn't add it now. I would say that the basic retrieval by ID presented covers the vast majority of cases.

docs/guides/03_writing_queries.md
jakubfijalkowski: That's extra info that should be in the docs but not here.

docs/guides/02_writing_commands.md
jakubfijalkowski: We should have different doc for that (we don't focus here on how we write projects, we focus on how to write commands only).

docs/guides/02_writing_commands.md
jakubfijalkowski: That's _too much info_. `WithMessage` is used only for non-default validation rules (because FA does not provide any meaningful message then).

mishioo: I do not agree that it's too much. A single code comment like `// no WithMessage(), Fluent Assertions provide meaningful messages for it's built-in rules` would be enough to make sure we're on the same page with the reader, even if they don't know FA.

jakubfijalkowski: I disagree. This overloads the reader with too much information. This particular doc is about creating commands, not about FluentValidation nor validation in general. Docs to be readable need to be focused.

jakubfijalkowski: I disagree. This overloads the reader with too much information. This particular doc is about creating commands, not about FluentValidation nor validation in general. Docs to be readable need to be focused.

mishioo: I strongly agree with docs being focused, I just think that validating is a part of creating a command.

docs/guides/02_writing_commands.md
mishioo: Agreed. In such case I hold my suggestion to remove logging from those examples as irrelevant to the topic.

bartekpacia:
Do you plan to use some docs site generator (e.g. [Docusaurus](https://docusaurus.io/), [docs.page](https://docs.page/))?

jakubfijalkowski:
@bartekpacia that's not our aim for now.

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,238 @@
...
+
+```json
+{
+    "Name": "abc",
+}
+```
+
+Notice the `ErrorCodes` class. These define possible validation errors which can be expected by the clients. Each error code should be a unique (in a given contract) integer.
+
+Writing contracts in that way makes it possible for web and mobile clients to generate their own version of contract classes with use of the Contract Generator, which is also a part of the CoreLibrary.
```
bartekpacia: Hyperlink to Contract Generator here could be useful.

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,238 @@
...
+
+Also, a contract can have authorization attributes, which define rules that have to be satisfied for an actor to be able to invoke the command. Authorization, however, is an extension to the contracts system and not an integral part of it. To learn more about authorizers, see [authorization](./0X_authorization.md).
```
bartekpacia: nitpicky: "actor" – why not just "client"? I think "client" is far more common

docs/guides/02_writing_commands.md
```diff
@@ -0,0 +1,238 @@
...
+
+## Command validator
+
+Command validator's main role is to ensure that a command invoked by a client is valid. Command validator should check the structure of the command itself, for example, check for required fields, check whether passed strings do not exceed maximum length, etc. Also the validator should check business requirements of the command. For example, when editing a project, it should check whether or not the project exists, see [more complex commands](#more-complex-commands) for an example. Let's see how a validator for `CreateProject` command might look like:
```
bartekpacia: > For example, when editing a project, it should check whether or not the project exists, see [more complex commands](#more-complex-commands) for an example.

Shouldn't it be two sentences?

bartekpacia:
Also, this is nitpicky, but I like formatting Markdown to fit into 80 columns. It makes it easier to do code review, and to `cat` it. I use [this extension](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap) for automating the formatting.

jakubfijalkowski:
> Also, this is nitpicky, but I like formatting Markdown to fit into 80 columns. It makes it easier to do code review, and to `cat` it. I use [this extension](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap) for automating the formatting.

Let's focus on the _content_, not technicalities, for now.

FrackiewiczP:
@bartekpacia I'll make another PR with the wrapping after this one.

jakubfijalkowski:
Merging as-is, will adjust formatting in the following PR.

jakubfijalkowski (approved):

PR #482, opened 2023-05-19
Remove autofac

wojtek2288:
I removed all changes in `Domain` for now, because I know that @marcin-dardzinski is currently working on it to switch to asp.net endpoint routing and removing autofac while doing that, so keeping this as draft until that task is done then we can remove it completely.


marcin-dardzinski:
One very general remark.
I'd remove usage of `AppModule` entirely and covert everything to `IServiceCollection` extension methods.

jakubfijalkowski:
@marcin-dardzinski step by step changes :)

src/Core/LeanCode.Components/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+    private static IServiceCollection TryRegisterWithImplementedInterfacesBase<T>(
+        IServiceCollection services,
+        ServiceLifetime lifetime,
+        Func<IServiceProvider, object>? implementationFactory
+    )
+        where T : class
+    {
+        var implementationType = typeof(T);
+
+        TryAddServiceDescriptor(services, implementationType, implementationType, lifetime, implementationFactory);
```
marcin-dardzinski: This is potentially problematic for singletons. You might expect to have one registered instance where in fact you'd have instance per interface

E.g. the following test fails
```
    [Fact]
    public void TestMethod()
    {
        var serviceCollection = new ServiceCollection();

        serviceCollection.AddSingleton<IFoo, Impl>();
        serviceCollection.AddSingleton<IBar, Impl>();

        var sp = serviceCollection.BuildServiceProvider();

        var foo = sp.GetService<IFoo>();
        var bar = sp.GetService<IBar>();

        Assert.Same(foo, bar);
    }

    interface IFoo { }
    interface IBar { }
    class Impl : IFoo, IBar { }
```

https://stackoverflow.com/a/60375138 - this might be the solution though although we need to handle disposables.

marcin-dardzinski:
Due to [this](https://github.com/leancodepl/corelibrary/pull/482#discussion_r1201785396), I decided to remove `AsImplemetedInterfaces` replacement for service provider, as the feature mismatch is too big (it lead to wrong registrations in two places in this PR). Wasn't so much a burden to explicitly give the interface, at least at the corelib level.

Also, I'll remove the modules in a separate PR as in some cases it'll be more work with the configuration.

src/Infrastructure/LeanCode.Firebase.FCM/FCMModule.cs
```diff
@@ -1,26 +1,27 @@
...
 
 public class FCMModule<TStore> : AppModule
     where TStore : class, IPushNotificationTokenStore
 {
-    protected override void Load(ContainerBuilder builder)
+    public override void ConfigureServices(IServiceCollection services)
     {
-        builder.RegisterModule(new FCMModule());
-        builder.RegisterType<TStore>().AsImplementedInterfaces();
+        (new FCMModule()).ConfigureServices(services);
```
Saancreed, line 24: Or maybe derive `FCMModule<TStore>` from `FCMModule` and call `base.ConfigureServices(services);` here instead?

src/Infrastructure/LeanCode.ViewRenderer.Razor/RazorViewRendererModule.cs
```diff
@@ -12,8 +13,8 @@ public RazorViewRendererModule(RazorViewRendererOptions opts)
         this.opts = opts;
     }
 
-    protected override void Load(ContainerBuilder builder)
+    public override void ConfigureServices(IServiceCollection services)
     {
-        builder.Register(_ => new RazorViewRenderer(opts)).As<IViewRenderer>().SingleInstance();
+        services.TryAddSingleton<IViewRenderer>(_ => new RazorViewRenderer(opts));
```
Saancreed, line 18: Hm, probably a dumb question but if it's supposed to be a singleton, why not instanciate one here already?

src/Infrastructure/LeanCode.Firebase.FCM/FCMModule.cs
marcin-dardzinski: Ughh, I guess it's supposed to configure either `LeanCode.Firebase.FCM.SqlServer` or `LeanCode.Firebase.FCM.PostreSql`.

I think I'll leave it as is - there will be subsequent pr's removing modules at all - 
this will probably end up similar to `services.AddFCM().AddSqlServerTokenStore()`

src/Infrastructure/LeanCode.ViewRenderer.Razor/RazorViewRendererModule.cs
marcin-dardzinski: Just a dumb rewrite :see_no_evil: 
The only difference I can see is that if one's invoked the method twice the constructor will be called only once, but I doubt that's it's a heavy call, so we can switch to explicit constructor.

Saancreed: Right, I'm wondering if this isn't some legacy baggage because in general our Razor code is… not in good shape.

jakubfijalkowski (approved):

PR #484, opened 2023-05-24
Replace LeanCode.Pipelines with asp.net core middlewares, switch to endpoint routing 

marcin-dardzinski:
For a quick overview of how the app is built now take a look at `LeanCode.CQRS.AspNetCore.Tests.Integration.RemoteCQRSTestsBase`

### Endpoint routing 
Instead of registering three paths (`/api/command/FullName` etc) each command (I'll use command as command/query operation) is registered as a separate endpoint/path.
Each endpoint has associated `CQRSEndpointMetadata` 
- `CQRSObjectMetadata` (object type, handler type, result type)
- executor method - equivalent of `PipelineFinalizer` - method responsible for resolving command handler from DI and executing it. 

All the commands are enumerated at startup to register endpoints via a custom `CQRSEndpointsDataSource`

### ASP.NET Middlewares

CQRS objects registered via `MapCQRS` methods go through the following middleware pipeline:
- `CQRSRequestSerializer` 
- user defined middlewares
- `CQRSPipelineFinalizer` 

`CQRSRequestSerializer` deserializes requests according to `CQRSObjectMetadata` and sets `CQRSRequestPayload`. In the return phase it writes `ExecutionResult` to http response.
`CQRSPipelineFinalizer` just gets the executor method from endpoint metadata and sets execution result. 

The pipeline is shared for each object kind, however you can put an arbitrary filter on `CQRSObjectMetadata` if you need seperate pipelines for some commands.

#### ExecutionResult
`ExecutionResult` was kept so that `CQRSPipelineFinalizer` does not write response directly - the problem is that otherwise we wouldn't be able to have middlewares for commiting transctions / relaying events. For the sake of consistency, other middlewares like security and validation are not writing to `HttpContext` directly and set execution result instead. However this is open for discussion.

#### Removal of custom app context
All the handlers, validators, authorizers have now direct access to `HttpContext`. 

### Startup registration

Some further work will be needed here, currently the minimal setup is 
```
serviceCollection.AddCQRS(contractsCatalog, handlersCatalog);
serviceCollection.AddFluentValidation(handlersCatalog);

applicationBuilder.MapRemoteCQRS("/cqrs/", opts => { }); 
```

It's awkward, as there's no connection between `AddCQRS` and `AddFluentValidation`  also there will be more such cases (e.g. events relay)

There is already is `CQRSServicesBuilder` which can only override `ISerializer` now, but it might look like this later
```
serviceCollection.AddCQRS(contractsCatalog, handlersCatalog)
    .AddFluentValidation()
    .AddDomainEvents()
    ...
```

### Misc stuff
- some names might feel awkward - I'm open for adjustments
- removed `LeanCode.Benchmarks` and `LeanCode.IntegrationTests` - they're of no use today, will be restored later.

### TODO
- [ ] startup error logs for invalid commands / missing handlers
- [x] tests for cqrs objects scanning
- [x] Injecting `HttpContext` in scoped validators
- [x] Migrate MassTransit integration (orchestrate with #478)
- [x] Migrate leftover pipeline elements like `TracingElement` (separate PR)
- [x] Complete removal of `LeanCode.Pipelines` (separate PR)


src/Domain/LeanCode.CQRS.AspNetCore/CQRSEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,61 @@
...
+    public static void MapRemoteCqrs(
+        this IEndpointRouteBuilder builder,
+        string path,
+        Action<CQRSPipelineBuilder> config
+    )
+    {
+        var cqrsHandlers =
+            builder.ServiceProvider.GetService<CQRSObjectsRegistrationSource>()
+            ?? throw new InvalidOperationException(
+                "CQRS services were not registered, make sure you've called IServiceCollection.AddCQRS(...) first"
```
jakubfijalkowski:
```suggestion
                "CQRS services were not registered, make sure you've called IServiceCollection.AddCQRS(...) first."
```

src/Domain/LeanCode.CQRS.AspNetCore/LeanCode.CQRS.AspNetCore.csproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <ItemGroup>
+      <ProjectReference Include="../LeanCode.CQRS.Execution/LeanCode.CQRS.Execution.csproj" />
+      <ProjectReference Include="../LeanCode.CQRS.Security/LeanCode.CQRS.Security.csproj" />
+      <ProjectReference Include="../../Core/LeanCode.Components/LeanCode.Components.csproj" />
+      <ProjectReference Include="../../Infrastructure/LeanCode.Serialization/LeanCode.Serialization.csproj" />
+      <ProjectReference Include="..\LeanCode.CQRS.Validation\LeanCode.CQRS.Validation.csproj" />
```
jakubfijalkowski:
```suggestion
      <ProjectReference Include="../LeanCode.CQRS.Validation/LeanCode.CQRS.Validation.csproj" />
```

src/Domain/LeanCode.CQRS.AspNetCore/CQRSEndpointMetadata.cs
```diff
@@ -0,0 +1,17 @@
+using Microsoft.AspNetCore.Http;
+
+namespace LeanCode.CQRS.AspNetCore;
+
+public delegate Task<object?> ObjectExecutor(HttpContext httpContext, CQRSRequestPayload payload);
```
Dragemil: Why not `Func`?

src/Domain/LeanCode.CQRS.AspNetCore/HttpContextExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+using Microsoft.AspNetCore.Http.Features;
+
+namespace LeanCode.CQRS.AspNetCore;
+
+public static class HttpContextExtensions
+{
+    public static CQRSEndpointMetadata GetCQRSEndpoint(this HttpContext httpContext)
+    {
+        return httpContext.GetEndpoint()?.Metadata.GetMetadata<CQRSEndpointMetadata>()
+            ?? throw new InvalidOperationException("Request does not contain CQRSEndpointMetadata");
```
Dragemil:
```suggestion
            ?? throw new InvalidOperationException("Request does not contain CQRSEndpointMetadata.");
```

src/Domain/LeanCode.CQRS.AspNetCore/IApplicationBuilderCQRSExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+    public static IApplicationBuilder LogCQRSResponsesOnNonProduction(this IApplicationBuilder builder)
+    {
+        return builder.UseMiddleware<NonProductionResponseLoggerMiddleware>();
+    }
+
+    public static IApplicationBuilder LogCQRSResponses(this IApplicationBuilder builder)
+    {
+        return builder.UseMiddleware<ResponseLoggerMiddleware>();
+    }
+}
```
Dragemil: Maybe let's have one method here with some enum options?

src/Domain/LeanCode.CQRS.AspNetCore/Middleware/CQRSPipelineFinalizer.cs
```diff
@@ -0,0 +1,16 @@
...
+internal static class CQRSPipelineFinalizer
+{
+    public static async Task HandleAsync(HttpContext context)
+    {
+        var cqrsEndpoint = context.GetCQRSEndpoint();
+        var payload = context.GetCQRSRequestPayload();
+
+        var result = await cqrsEndpoint.ObjectExecutor(context, payload);
+
+        payload.SetResult(ExecutionResult.Success(result));
```
Dragemil: If the result is `Success` no matter if executor throws, then when the result is `Fail`?

src/Domain/LeanCode.CQRS.AspNetCore/Middleware/CQRSRequestSerializer.cs
```diff
@@ -0,0 +1,102 @@
...
+        try
+        {
+            await next(httpContext);
+        }
+        catch (Exception ex)
+        {
+            logger.Error(ex, "Cannot execute object {@Object} of type {Type}", obj, objectType);
+            httpContext.Response.StatusCode = StatusCodes.Status500InternalServerError;
+            return;
+        }
```
Dragemil: That doesn't seem quite _request serializer_ logic.

src/Domain/LeanCode.CQRS.AspNetCore/IApplicationBuilderCQRSExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+using Microsoft.AspNetCore.Builder;
+
+namespace LeanCode.CQRS.AspNetCore;
+
+public static class IApplicationBuilderCQRSExtensions
+{
+    public static IApplicationBuilder Validate(this IApplicationBuilder builder)
+    {
+        return builder.UseMiddleware<CQRSValidationMiddleware>();
+    }
```
Dragemil, line 12: Can we at this point ensure that it is added for command pipeline and throw otherwise?

src/Domain/LeanCode.CQRS.AspNetCore/Registration/CQRSObjectsRegistrationSource.cs
```diff
@@ -0,0 +1,150 @@
...
+
+    public CQRSObjectsRegistrationSource(TypesCatalog contractsCatalog, TypesCatalog handlersCatalog)
+    {
+        this.contractsCatalog = contractsCatalog;
+        this.handlersCatalog = handlersCatalog;
+
+        Objects = FindCqrsObjects();
+    }
+
+    private IEnumerable<CQRSObjectMetadata> FindCqrsObjects()
```
Dragemil: Inconsistent naming. However I could roll with `Cqrs` everywhere but not within this PR for sure.

Dragemil (commented):
Big boye this one, hopefully I'll be back stronger.

src/Domain/LeanCode.CQRS.AspNetCore/CQRSEndpointMetadata.cs
marcin-dardzinski: Well, actually the signature of this method changed three or four times during the PR - so it was easier to have a separate one, but maybe we can change it as it stabilized.

src/Domain/LeanCode.CQRS.AspNetCore/LeanCode.CQRS.AspNetCore.csproj
marcin-dardzinski: Rider :woman_facepalming:

src/Domain/LeanCode.CQRS.AspNetCore/IApplicationBuilderCQRSExtensions.cs
marcin-dardzinski: Not really. 
You can do some dirty tricks to get the built pipeline, but I don't actually know what implementation of `IApplicationBuilder` is fed there. 
Also, one can always use `UseMiddleware<CQRSValidationMiddleware>()` which we cannot control.

src/Domain/LeanCode.CQRS.AspNetCore/IApplicationBuilderCQRSExtensions.cs
marcin-dardzinski: I don't understand :confused: .

src/Domain/LeanCode.CQRS.AspNetCore/Middleware/CQRSPipelineFinalizer.cs
marcin-dardzinski: Yeah, this is an evolutional baggage from remote cqrs ancestor, maybe it's time to correct the design :stuck_out_tongue: 

If exception is thrown, success wont be set - the exception will bubble up instead. Now, the old `ExecutionResult` had an `Success` and `Fail` methods, but looking back - it all boiled down to setting a status code or/and setting a result to serialize.

So, maybe it's better to rebrand it as `ExecutionResult.Empty(statusCode)` and `ExecutionResult.Payload(payload, statusCode = 200)`.

src/Domain/LeanCode.CQRS.AspNetCore/Middleware/CQRSRequestSerializer.cs
marcin-dardzinski: As I said the naming might be bad, but the idea is to read the request and then write a response. I don't see any gain from having a separate middleware for reading and writing.

src/Domain/LeanCode.CQRS.AspNetCore/IApplicationBuilderCQRSExtensions.cs
Saancreed: I think Emil meant something like

```csharp
public enum Environments
{
    Development = 1 << 0,
    Test = 1 << 1,
    Production = 1 << 2,
}

public static IApplicationBuilder LogCQRSResponses(this IApplicationBuilder builder, Environments environments)
```

src/Domain/LeanCode.CQRS.AspNetCore/Middleware/CQRSRequestSerializer.cs
Saancreed: Then let's figure out a better name? `CQRS{Content/Payload}Converter`? :stuck_out_tongue:

src/Domain/LeanCode.CQRS.AspNetCore/IApplicationBuilderCQRSExtensions.cs
Saancreed: We could add an analyzer that naively looks at action assigned to `CQRSPipelineBuilder.Commands` but it's probably not worth the effort.

src/Domain/LeanCode.CQRS.AspNetCore/Middleware/CQRSRequestSerializer.cs
Dragemil: Maybe something like `CQRSPipelineInitializer` since it's expected to always wrap all CQRS related middlewares?

src/Domain/LeanCode.CQRS.AspNetCore/IApplicationBuilderCQRSExtensions.cs
jakubfijalkowski: Let's add that to the backlog - changing the way this middleware works is out of scope for now.

src/Domain/LeanCode.CQRS.AspNetCore/Middleware/CQRSRequestSerializer.cs
jakubfijalkowski: Just `CQRSMiddleware`?

src/Domain/LeanCode.CQRS.AspNetCore/IApplicationBuilderCQRSExtensions.cs
jakubfijalkowski: Let's wage the benefit here - you _can_ use it wrongly, but you configure that once per project basically, so the pain of mistake will be very low (and very rare).

src/Domain/LeanCode.CQRS.AspNetCore/Middleware/CQRSRequestSerializer.cs
marcin-dardzinski: `CQRSMiddleware` it is.

test/Domain/LeanCode.DomainModels.MassTransitRelay.Tests/ScopedFiltersTests.cs
```diff
@@ -10,34 +10,44 @@
 
 namespace LeanCode.DomainModels.MassTransitRelay.Tests;
 
-public class ScopedFiltersTests : IAsyncLifetime
+public class ScopedFiltersTests : IAsyncLifetime, IDisposable
```
Dragemil, line 13: `IAsyncDisposable` as well?

test/Domain/LeanCode.DomainModels.MassTransitRelay.Tests/ScopedFiltersTests.cs
```diff
@@ -10,34 +10,44 @@
...
+                    cfg.ReceiveEndpoint(
+                        "queue",
+                        rcv =>
+                        {
+                            Filter1Observer.UseFilter1(rcv, ctx);
+                            Filter2Observer.UseFilter2(rcv, ctx);
+
+                            rcv.ConfigureConsumers(ctx);
+                        }
+                    );
```
Dragemil: Shouldn't we somehow rewrite that in an _endpoint per consumer_ way to promote good practices? Also that should be our main case so the test would be more relevant.

test/Domain/LeanCode.CQRS.AspNetCore.Tests.Integration/RemoteCQRSCommandsTests.cs
```diff
@@ -0,0 +1,113 @@
...
+    public async Task Returns_UnprocessableEntity_with_errors_when_validation_fails()
+    {
+        var (body, statusCode) = await SendAsync(
+            "/cqrs/command/LeanCode.CQRS.AspNetCore.Tests.Integration.TestCommand",
+            @"{ ""FailValidation"": true }"
+        );
+
+        var commandResult = JsonSerializer.Deserialize<CommandResult>(body);
+
+        // Assert.Equal(HttpStatusCode.UnprocessableEntity, statusCode);
```
Dragemil: Why commented?

test/Domain/LeanCode.CQRS.AspNetCore.Tests.Integration/RemoteCQRSCommandsTests.cs
Dragemil, line 1: We want to use FluentAssertions here, right?

test/Domain/LeanCode.DomainModels.MassTransitRelay.Tests/ScopedFiltersTests.cs
marcin-dardzinski: xUnit does not use `IAsyncDisposable` at all.

test/Domain/LeanCode.CQRS.AspNetCore.Tests.Integration/RemoteCQRSCommandsTests.cs
marcin-dardzinski: In general yes, but I'd keep it out of scope now.

jakubfijalkowski (approved):
Let's merge this one as-is. It is too big already.

PR #485, opened 2023-05-29
Rework mass transit relay to use asp.net middlewares

marcin-dardzinski:
Targetting #484

Nothing fancy, but one quirk.
`AddCQRSMassTransitIntegration` actually does not add mass transit at all, you still need to call `serviceCollection.AddMassTransit` yourself (similar to what was now, except that you _had_ to override a method that prompted you to do so). We can make our method take `Action<IBusRegistrationConfigurator>` and register MT here, but I feel that's a bit too restrictive, e.g. you can't use `AddMassTransitTestHarness` version, which might be usable in tests.


src/Domain/LeanCode.DomainModels.MassTransitRelay/MassTransitRelayServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+    public static void AddCQRSMassTransitIntegration(
+        this IServiceCollection services,
+        Action<MassTransitHostOptions>? hostCfg = null
+    )
+    {
+        var interceptor = new AsyncEventsInterceptor();
+        interceptor.Configure();
+        services.AddSingleton(interceptor);
+
+        hostCfg ??= opts =>
```
jakubfijalkowski, line 19: This should be controlled by the parent IMO, we shouldn't force that.

test/Domain/LeanCode.DomainModels.MassTransitRelay.Tests/ScopedFiltersTests.cs
```diff
@@ -249,4 +232,6 @@ public void SetConsumer(Guid messageId, Guid value)
             return Data.TryGetValue($"{messageId}_consumer", out var value) ? value : null;
         }
     }
+
+    public void Dispose() => serviceProvider.Dispose();
```
Saancreed, line 236: It's weird how this doesn't do the same thing as `DisposeAsync`.

src/Domain/LeanCode.DomainModels.MassTransitRelay/Middleware/CommitDatabaseTransactionMiddleware.cs
```diff
@@ -0,0 +1,21 @@
+using Microsoft.AspNetCore.Http;
+using Microsoft.EntityFrameworkCore;
+
+namespace LeanCode.DomainModels.MassTransitRelay.Middleware;
+
+public class CommitDatabaseTransactionMiddleware<TDbContext>
```
Saancreed, line 6: Now that it has been split away from `PublishEvents`, this could be put in some separate package, closer to some generic EF stuff. I can imagine this being useful in a vacuum, even without MT (and even though we don't have backends without MT nowadays).

src/Domain/LeanCode.DomainModels.MassTransitRelay/MassTransitRelayServiceCollectionExtensions.cs
marcin-dardzinski: I feel that lack of it will be a painpoint in our projects - we will forget about it. 
Combined with what I've said in the description maybe we need a split ->

`serviceCollection.AddAsyncEventsInterceptor()` - self explanatory

`serviceColllection.AddCQRSMassTransitIntegration(Action<MassTransitHostOptions>, Action<IBusRegistrationConfigurator>)` - this one will call the first one + register MT and will be intended for regular use.

But if sb needs other way to register mass transit, e.g. through test harness, they'll have a possibility.

Saancreed: There is nothing being forced here as far as I can tell, unless @jakubfijalkowski meant to say that we shouldn't `AddOptions<MassTransitHostOptions>()`. But having `opts.WaitUntilStarted = true;` sounds like a reasonable default and callers can always opt out of it by passing their own `hostCfg`.

test/Domain/LeanCode.DomainModels.MassTransitRelay.Tests/ScopedFiltersTests.cs
marcin-dardzinski: Well, that xUnit's `IAsyncLifetime` not me :stuck_out_tongue:

src/Domain/LeanCode.DomainModels.MassTransitRelay/Middleware/CommitDatabaseTransactionMiddleware.cs
marcin-dardzinski: I don't think it's worth it now...

test/Domain/LeanCode.DomainModels.MassTransitRelay.Tests/ScopedFiltersTests.cs
Saancreed: Ah, and `IAsyncLifetime` is not `IAsyncDisposable`, right :confused:

src/Domain/LeanCode.DomainModels.MassTransitRelay/MassTransitRelayServiceCollectionExtensions.cs
jakubfijalkowski: > I feel that lack of it will be a painpoint in our projects - we will forget about it.

I agree, we will forget that, but this is no excuse to break SRP - you're saying that you configure the relay and force user to configure MT separately, yet you are configuring MT to some extent.

> There is nothing being forced here as far as I can tell

You are forced to opt out of it (which it seemingly does not do), which is not intuitive.

src/Domain/LeanCode.DomainModels.MassTransitRelay/Middleware/CommitDatabaseTransactionMiddleware.cs
jakubfijalkowski: I also think it is not worth it. It is usable separately, but the idea is that it is part of the relay/outbox integration.

src/Domain/LeanCode.DomainModels.MassTransitRelay/MassTransitRelayServiceCollectionExtensions.cs
Saancreed: > which is not intuitive

You're not incorrect, but it's still something we should probably encourage by default. If not here, maybe in the template?

src/Domain/LeanCode.DomainModels.MassTransitRelay/Middleware/CommitDatabaseTransactionMiddleware.cs
Saancreed: Fair enough.

src/Domain/LeanCode.DomainModels.MassTransitRelay/MassTransitRelayServiceCollectionExtensions.cs
jakubfijalkowski: I'm only arguing that it should not be forced here. We might have some other way of ensuring that (even some `AddDefaultMTConfiguration`) or put that in template, but not _here_.

marcin-dardzinski: :popcorn: 

So, as I said maybe sth like this:
```
    public static IServiceCollection AddCQRSMassTransitIntegration(this IServiceCollection services,
        Action<IBusRegistrationConfigurator> busCfg,
        Action<MassTransitHostOptions>? hostCfg = null)
    {
        services.AddAsyncEventsInterceptor();

        hostCfg ??= opts =>
        {
            opts.WaitUntilStarted = true;
        };

        services.AddOptions<MassTransitHostOptions>().Configure(hostCfg);
        services.AddMassTransit(busCfg);

        return services;
    }

    public static IServiceCollection AddAsyncEventsInterceptor(
        this IServiceCollection services)
    {
        var interceptor = new AsyncEventsInterceptor();
        interceptor.Configure();
        services.AddSingleton(interceptor);

        return services;
    }
```

src/Domain/LeanCode.DomainModels.MassTransitRelay/Middleware/EventsPublisherMiddleware.cs
```diff
@@ -0,0 +1,58 @@
...
+            await PublishEventsAsync(publishEndpoint, events, httpContext.RequestAborted);
+        }
+    }
+
+    private Task PublishEventsAsync(IPublishEndpoint publishEndpoint, List<IDomainEvent> events, CancellationToken ct)
+    {
+        logger.Debug("Publishing {Count} raised events", events.Count);
+        var conversationId = Guid.NewGuid();
+
+        var publishTasks = events.Select(evt => PublishEventAsync(publishEndpoint, evt, conversationId, ct));
```
Dragemil, line 35: Can't we use some `PublishBatch` here?

test/Domain/LeanCode.DomainModels.MassTransitRelay.Tests/Integration/TestApp.cs
```diff
@@ -60,79 +46,101 @@ public TestApp()
...
+                        app.UseRouting();
+                        app.UseEndpoints(ep =>
+                        {
+                            ep.MapRemoteCqrs(
+                                "/cqrs",
+                                cqrs =>
+                                {
+                                    cqrs.Queries = q => q.Secure();
+                                    cqrs.Commands = c =>
+                                        c.Secure().Validate().CommitTransaction<TestDbContext>().PublishEvents();
```
Dragemil, line 77: I don't get it - in order for Transactional Outbox to work shouldn't we `.PublishEvents()` first which would probably mark the events to store in the DB, then `CommitTransaction` which would actually save the data transactionally to the DB and then MT would publish them sometime?

src/Domain/LeanCode.DomainModels.MassTransitRelay/Middleware/EventsPublisherMiddleware.cs
marcin-dardzinski: Under the hood it does exactly the same :stuck_out_tongue: 
Especially that we'll primarily use it with outbox which would just append a message to a list, so not much of a difference here.

test/Domain/LeanCode.DomainModels.MassTransitRelay.Tests/Integration/TestApp.cs
marcin-dardzinski: Middlewares are bidirectional so in this case event publishing will be first in the return phase, then commit.

And yes, since the outbox is configured `PublishEvents` will just add them to db context. When the request scope ends the background outbox service will be pulsed to publish events to the transport.

src/Domain/LeanCode.DomainModels.MassTransitRelay/Middleware/EventsPublisherMiddleware.cs
jakubfijalkowski: If this "just" appends to an outbox (so - dbcontext), is it possible to switch to sync?

marcin-dardzinski: No, there're no sync counterparts.
Digging through MT onion of abstraction I see that indeed it's 'just' adding to db context, but with `AddAsync` :confused:

jakubfijalkowski (approved):

Dragemil (approved):

PR #486, opened 2023-05-30
Remove AppModule from Infrastructure projects

marcin-dardzinski:
In applicable places the required configuration is being asked for directly.
Also IdentityServer key vault client registers the signing key on it's own (still the client app should configure the credentials though)


jakubfijalkowski:
Let's add to our TODO to remove IdentityServer-related things, as we won't be using it any longer (in this form).

jakubfijalkowski (approved):

PR #487, opened 2023-05-30
Use harness inactivity task to wait for completion in tests

marcin-dardzinski:


jakubfijalkowski:
What's the reason?

marcin-dardzinski:
There is flakiness with the test - I just wanted to see if the fix works on the CI w/o other changes. On the other hand we might as well merge it now to not wait for #485.

PR #489, opened 2023-06-01
Pass HttpContext to IValidationContext in LeanCode.CQRS.Validation.Fl…

marcin-dardzinski:
…uent.Scoped.

Since with `IServiceProvider` we can't inject it via parameter hints, restoring the `RootContextData` injection. One can still use an `IHttpContextAccessor`, but I still think it's easier that way.

### DIFF
--- a/src/Domain/LeanCode.CQRS.Validation.Fluent.Scoped/ValidationContextExtensions.cs
+++ b/src/Domain/LeanCode.CQRS.Validation.Fluent.Scoped/ValidationContextExtensions.cs
@@ -17,7 +17,7 @@ public static class ValidationContextExtensions
         else
         {
             throw new InvalidOperationException(
-                "`HttpContext` is not available in validation context. Ensure you are calling the validator through FluentValidationCommandValidatorAdapter"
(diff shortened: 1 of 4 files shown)


src/Domain/LeanCode.CQRS.Validation.Fluent.Scoped/ValidationContextExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+    public static HttpContext HttpContext(this IValidationContext ctx)
+    {
+        if (ctx.RootContextData.TryGetValue(HttpContextKey, out var httpContext))
+        {
+            return (HttpContext)httpContext;
+        }
+        else
+        {
+            throw new InvalidOperationException(
+                "`HttpContext` is not available in validation context. Ensure you are calling the validator through FluentValidationCommandValidatorAdapter"
```
jakubfijalkowski:
```suggestion
                "`HttpContext` is not available in validation context. Ensure you are calling the validator through FluentValidationCommandValidatorAdapter."
```

jakubfijalkowski (approved):

PR #491, opened 2023-06-05
Remove Autofac leftovers 

marcin-dardzinski:
No projects depends now on Autofac, except `LeanCode.Startup.Autofac`, `LeanCode.Components.Autofac`.
Projects can choose between `LeanCode.Startup.Autofac` and `LeanCode.Startup.MicrosoftDI`. The latter does not have any `IAppModule` counterpart - any logical 'module' is supposed to be registered as a plain extension method on `IServiceCollection`


Dragemil (approved):
AspNetCoreDI sounds better to me, rather than MicrosoftDI.

marcin-dardzinski:
Well, it's independent of asp net core really, but it's not a big deal imo.

jakubfijalkowski (approved):

PR #492, opened 2023-06-06
Add scoped transaction middleware with controllable optimistic/pessim…

### LINKED ISSUE #383: Allow configuring database transaction level per handler
Currently, when using the implicit handler transaction, we use the default transaction level. This is not always the best approach. There are cases when you want to use higher transaction level (to e.g. get make the concurrency pessimistic).

What I would propose - an attribute that is attached to the command handler that allows specifying transaction level. Then, the middleware that guards the transaction reads it at runtime.

marcin-dardzinski:
…istic concurrency per handler.

Now that we now the handler type up front we can make use of it. In this case - optional PC based on attribute on a handler. This is more of a demo, but if we decide to proceed with it, I'd remove the corresponding middleware in `LeanCode.CQRS.MassTransitRelay`.

There is problem with testing it though - in mem / mocking does not really work here - we'd need to wait for integration tests restoration.


src/CQRS/LeanCode.CQRS.ScopedTransaction/ScopedTransactionMiddleware.cs
```diff
@@ -0,0 +1,89 @@
...
+    private bool UsePessimisticConcurrency(
+        HttpContext httpContext,
+        [NotNullWhen(true)] out PessimisticConcurrencyAttribute? config
+    )
+    {
+        var cqrsMetadata = httpContext.GetCQRSEndpoint().ObjectMetadata;
+        var handlerType = cqrsMetadata.HandlerType;
+
+        if (
+            Attribute.GetCustomAttribute(handlerType, typeof(PessimisticConcurrencyAttribute))
```
jakubfijalkowski, line 35: How is the performance of this? Should we plan on having some kind of cache?

marcin-dardzinski: Yeah, this can be wrapped in ConcurrentDictionary and cached

jakubfijalkowski: Let's put that on our backlog, with more systemic approach (I really hope that this approach will be used thoroughly :D).

jakubfijalkowski (approved):

PR #496, opened 2023-06-13
Propagate user id and roles in tracing baggage and attach it to tags

marcin-dardzinski:


src/Infrastructure/LeanCode.OpenTelemetry/IdentityTraceAttributesFromBaggageProcessor.cs
```diff
@@ -0,0 +1,22 @@
...
+        if (activity.GetBaggageItem(IdentityTraceBaggageHelpers.UserIdKey) is string userId)
+        {
+            activity.SetTag(IdentityTraceBaggageHelpers.UserIdKey, userId);
+        }
+
+        var roles = activity.GetUserRoleBaggage();
+
+        if (roles is not null)
+        {
+            activity.SetTag(IdentityTraceBaggageHelpers.RoleKey, roles);
```
Dragemil, line 19: Have you checked how queryable is  array in the DD search?

Dragemil (approved):

src/Infrastructure/LeanCode.OpenTelemetry/IdentityTraceAttributesFromBaggageProcessor.cs
marcin-dardzinski: Don't have easy way to upload a trace to DD. But in general it can filter logs by array members so I guess it's all the same.

jakubfijalkowski: Idea - add (semi-public) integration with DD to example app.

jakubfijalkowski (approved):

PR #498, opened 2023-06-20
Switch to .net8.0 preview

marcin-dardzinski:
We have a very nice naming conflict with `System.TimeProvider` that will need addressing

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -1,8 +1,8 @@
 <Project>
   <PropertyGroup>
-    <EntityFrameworkCoreVersion>7.0.0</EntityFrameworkCoreVersion>
-    <AspNetCoreVersion>7.0.0</AspNetCoreVersion>
-    <ExtensionsVersion>7.0.0</ExtensionsVersion>
+    <EntityFrameworkCoreVersion>8.0.0-preview.5.23280.1</EntityFrameworkCoreVersion>
+    <AspNetCoreVersion>8.0.0-preview.5.23302.2</AspNetCoreVersion>
+    <ExtensionsVersion>8.0.0-preview.5.23280.8</ExtensionsVersion>
 
     <GoogleApisVersion>1.58.0</GoogleApisVersion>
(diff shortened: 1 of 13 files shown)


Directory.Build.targets
```diff
@@ -95,7 +95,8 @@
     <PackageReference Update="Microsoft.Extensions.Logging.Debug" Version="$(ExtensionsVersion)" />
     <PackageReference Update="Microsoft.Extensions.Options" Version="$(ExtensionsVersion)" />
     <PackageReference Update="Microsoft.Extensions.Options.ConfigurationExtensions" Version="$(ExtensionsVersion)" />
-    <PackageReference Update="Microsoft.Extensions.WebEncoders" Version="$(ExtensionsVersion)" />
+
+    <PackageReference Update="Microsoft.Extensions.WebEncoders" Version="$(AspNetCoreVersion)" />
```
jakubfijalkowski, line 99: Considering this is `Microsoft.Extensions` package, why are we changing the version to be `AspNetCoreVersion` variable?

jakubfijalkowski (approved):

Directory.Build.targets
marcin-dardzinski: Well, the project is in the [aspnetcore](https://github.com/dotnet/aspnetcore) repo and versioning is aligned. 
I can leave a comment or move it to asp net section of this file.

Saancreed:
I'd say let's dump our implementation of `TimeProvider` class and instead let the package be just a static source of `System.TimeProvider` for our purpose that is initialized with `TimeProvider.System` plus maybe a `FixedTimeProvider` that would be a subclass of `System.TimeProvider` for tests, but maybe the latter could exist in a separate package for testing.

Directory.Build.targets
jakubfijalkowski: Ahh, got it. No need to move, thx.

jakubfijalkowski (approved):

marcin-dardzinski:
I've noted time provider adjustments in backlog, let's address them separately.

PR #500, opened 2023-06-21
Restore integration tests 

marcin-dardzinski:
Rework of `LeanCodeTestFactory`:
- removed authorization related stuff
- allow configuring http client & json serialization options
- added helpers for MassTransit 
- added a `TestAuthorizationHandler` allowing to pass arbitrary `ClaimsPrincipal` in http request - this is not registered by default though

Also, restored old integration tests and enabled integrations tests on CI builds


.github/workflows/core_build_publish.yml
```diff
@@ -97,7 +105,12 @@ jobs:
       - name: Test
         run: dotnet msbuild /t:RunTestsWithCoverage
         working-directory: test
-
+      - name: Integration Test
```
jakubfijalkowski, line 136: TBH I would move integration tests to a separate workflow. It will be easier to manage/see the problem there.

test/LeanCode.IntegrationTests/App/Startup.cs
```diff
@@ -0,0 +1,61 @@
...
+        services.AddRouting();
+        services.AddCQRS(CQRSTypes, CQRSTypes);
+        services.AddFluentValidation(CQRSTypes);
+
+        services
+            .AddAuthentication(TestAuthenticationHandler.SchemeName)
+            .AddScheme<AuthenticationSchemeOptions, TestAuthenticationHandler>(
+                TestAuthenticationHandler.SchemeName,
+                _ => { }
+            );
```
jakubfijalkowski: Can we wrap this in an extension method?

jakubfijalkowski (approved):

.github/workflows/core_build_publish.yml
marcin-dardzinski: Can we orchestrate workflows so that publish job is done only if integration tests pass?

jakubfijalkowski: Yup, jobs can have dependencies (and workflows are somewhat transparent groups of jobs and can reference cross-workflows). [Example](https://github.com/jakubfijalkowski/kvenv/blob/main/.github/workflows/ci.yaml#L93).

src/Testing/LeanCode.IntegrationTestHelpers/ConfigurationOverrides.cs
```diff
@@ -7,8 +7,6 @@ public class ConfigurationOverrides : IConfigurationSource
 {
     public const LogEventLevel MinimumLevelDefault = LogEventLevel.Warning;
     public const bool EnableInternalLogsDefault = false;
-    public const string ConnectionStringBaseDefault = "SqlServer__ConnectionStringBase";
-    public const string ConnectionStringKeyDefault = "SqlServer:ConnectionString";
     public const string InternalBaseKeyDefault = "InternalBase";
     public const string PublicBaseKeyDefault = "PublicBase";
```
Saancreed: I think these are unused now that we don't have IS4 anymore.

src/Testing/LeanCode.IntegrationTestHelpers/DbContextInitializer.cs
Saancreed, line 1: Aren't we removing SQL Server specific stuff? Or at least can we make it not assume the driver used? Or let it handle both relevant cases?

src/Testing/LeanCode.IntegrationTestHelpers/LeanCodeTestFactory.cs
```diff
@@ -16,60 +15,57 @@ namespace LeanCode.IntegrationTestHelpers;
...
+    public virtual JsonSerializerOptions JsonOptions { get; } =
+        new()
+        {
+            Converters =
+            {
+                new JsonLaxDateOnlyConverter(),
+                new JsonLaxTimeOnlyConverter(),
+                new JsonLaxDateTimeOffsetConverter(),
+            },
+        };
```
Saancreed, line 36: `MakeReadOnly()` returns `void` so we can't chain it nicely :cry:

src/Testing/LeanCode.IntegrationTestHelpers/LeanCodeTestFactory.cs
```diff
@@ -16,60 +15,57 @@ namespace LeanCode.IntegrationTestHelpers;
...
-        return new HttpCommandsExecutor(CreateApiClient());
+        return new HttpOperationsExecutor(CreateApiClient(config), JsonOptions);
     }
 
-    public virtual async Task<bool> AuthenticateAsync(PasswordTokenRequest tokenRequest)
+    public virtual async Task WaitForBusAsync(TimeSpan? delay = null)
     {
-        // FIXME: what if the auth server is outside of the app?
-        using var client = CreateAuthClient();
+        delay ??= TimeSpan.FromSeconds(5);
```
Saancreed, line 68: Assign to another local of type `TimeSpan` to avoid awkward `.Value` later.

src/Testing/LeanCode.IntegrationTestHelpers/TestAuthenticationHandler.cs
```diff
@@ -0,0 +1,83 @@
...
+
+public class TestAuthenticationHandler : AuthenticationHandler<AuthenticationSchemeOptions>
+{
+    public const string SchemeName = "Test";
+
+    public TestAuthenticationHandler(
+        IOptionsMonitor<AuthenticationSchemeOptions> options,
+        ILoggerFactory logger,
+        UrlEncoder encoder,
+        ISystemClock clock
```
Saancreed: Nice warning :upside_down_face:

src/Testing/LeanCode.IntegrationTestHelpers/DbContextInitializer.cs
marcin-dardzinski: Is there a similar problem in postgres? I think for the sake of simplicity we should keep it as is.

src/Testing/LeanCode.IntegrationTestHelpers/LeanCodeTestFactory.cs
marcin-dardzinski: Don't see anything awkward in it

src/Testing/LeanCode.IntegrationTestHelpers/DbContextInitializer.cs
Saancreed: Not that I'm aware, but there _are_ some transient errors that can be thrown with Npgsql there as well, I'm just not sure how common they are when running migrations in integration tests. At least they are more reasonable to detect: `NpgsqlException.IsTransient`

src/Testing/LeanCode.IntegrationTestHelpers/LeanCodeTestFactory.cs
Saancreed: I do, you're kind of relying on the compiler to optimize away null checks.

src/Testing/LeanCode.IntegrationTestHelpers/LeanCodeTestFactory.cs
```diff
@@ -17,9 +18,15 @@ public abstract class LeanCodeTestFactory<TStartup> : WebApplicationFactory<TSta
...
 
-    public virtual JsonSerializerOptions JsonOptions { get; } =
-        new()
+    public virtual JsonSerializerOptions JsonOptions { get; }
+
+    protected virtual string ApiBaseAddress => "api/";
+
+    protected LeanCodeTestFactory()
+    {
+        var options = new JsonSerializerOptions
```
Saancreed, line 27: You can just assign to `JsonOptions` immediately then do `JsonOptions.MakeReadOnly();`.

Saancreed (approved):

src/Testing/LeanCode.IntegrationTestHelpers/LeanCodeTestFactory.cs
marcin-dardzinski: Then you'll get an ugly virtual call in constructor warning.

PR #502, opened 2023-06-26
Kratos web hooks

Saancreed:
~~Those `#if NET8_0_OR_GREATER` checks are probably just unnecessary noise, I'll get rid of them soon (after some manual testing on, well, ASP.NET Core 7.0 :see_no_evil:).~~

### DIFF
--- a/src/Infrastructure/LeanCode.Kratos/KratosAuthenticationHandler.cs
+++ b/src/Infrastructure/LeanCode.Kratos/KratosAuthenticationHandler.cs
@@ -30,10 +30,9 @@ public class KratosAuthenticationHandler<TOptions> : AuthenticationHandler<TOpti
         IOptionsMonitor<TOptions> options,
         ILoggerFactory logger,
         UrlEncoder encoder,
-        ISystemClock clock,
         IFrontendApi api
     )
-        : base(options, logger, encoder, clock)
+        : base(options, logger, encoder)
     {
         this.api = api;
     }
@@ -164,8 +163,7 @@ public class KratosAuthenticationHandler : KratosAuthenticationHandler<KratosAut
         IOptionsMonitor<KratosAuthenticationOptions> options,
         ILoggerFactory logger,
         UrlEncoder encoder,
-        ISystemClock clock,
         IFrontendApi api
     )
-        : base(options, logger, encoder, clock, api) { }
+        : base(options, logger, encoder, api) { }
 }
--- a/src/Infrastructure/LeanCode.Kratos/KratosWebHookHandlerBase.cs
+++ b/src/Infrastructure/LeanCode.Kratos/KratosWebHookHandlerBase.cs
@@ -85,15 +85,15 @@ public abstract partial class KratosWebHookHandlerBase
 [JsonSerializable(typeof(ErrorResponseBody))]
 public partial class KratosWebHookHandlerBaseJsonSerializerContext : JsonSerializerContext;
 
-#pragma warning disable CA2227 // Collection properties should be read only
-
 public record struct IdentityResponseBody([property: JsonPropertyName("identity")] Identity Identity);
 
-public record struct ErrorResponseBody([property: JsonPropertyName("messages")] List<ErrorMessage> Messages);
+public record struct ErrorResponseBody(
+    [SuppressMessage("?", "CA2227")] [property: JsonPropertyName("messages")] List<ErrorMessage> Messages
+);
 
 public record struct ErrorMessage(
     [property: JsonPropertyName("instance_ptr")] string? InstancePtr,
-    [property: JsonPropertyName("messages")] List<DetailedMessage> Messages
+    [SuppressMessage("?", "CA2227")] [property: JsonPropertyName("messages")] List<DetailedMessage> Messages
 );
 
 public record struct DetailedMessage(
--- a/src/Infrastructure/LeanCode.Kratos/KratosWebHookHandlerBase.cs
+++ b/src/Infrastructure/LeanCode.Kratos/KratosWebHookHandlerBase.cs
@@ -1,0 +1,104 @@
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.InteropServices;
+using System.Security.Cryptography;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using LeanCode.Kratos.Model;
+using Microsoft.AspNetCore.Http;
+
+namespace LeanCode.Kratos;
+
+public sealed record class KratosWebHookHandlerConfig(string ApiKey);
+
+public abstract partial class KratosWebHookHandlerBase
+{
+    private readonly Serilog.ILogger logger = Serilog.Log.ForContext<KratosWebHookHandlerBase>();
+
+    private readonly KratosWebHookHandlerConfig config;
+
+    protected KratosWebHookHandlerBase(KratosWebHookHandlerConfig config)
+    {
+        this.config = config;
+    }
+
+    protected virtual string ApiKeyHeaderName => "X-Api-Key";
+
+    protected abstract Task HandleCoreAsync(HttpContext ctx);
+
+    [SuppressMessage(
+        "?",
+        "CA1031",
+        Justification = "Exception is logged, rethrowing it for ASP.NET to handle is unnecessary."
+    )]
+    public async Task HandleAsync(HttpContext ctx)
+    {
+        try
+        {
+            if (VerifyApiKey(ctx.Request.Headers[ApiKeyHeaderName].ToString(), config.ApiKey))
+            {
+                await HandleCoreAsync(ctx);
+            }
+            else
+            {
+                logger.Error("Invalid Api Key");
+                ctx.Response.StatusCode = 403;
+            }
+        }
+        catch (Exception e)
+        {
+            logger.Error(e, "Failed to process webhook");
+            ctx.Response.StatusCode = 500;
+        }
+    }
+
+    private static bool VerifyApiKey(string left, string right)
+    {
+        var leftBytes = MemoryMarshal.AsBytes(left.AsSpan());
+        var rightBytes = MemoryMarshal.AsBytes(right.AsSpan());
+
+        return CryptographicOperations.FixedTimeEquals(leftBytes, rightBytes);
+    }
+
+    public static Task WriteIdentityResponseAsync(HttpContext ctx, Identity identity)
+    {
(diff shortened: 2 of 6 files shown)


Saancreed:
Seems to work :slightly_smiling_face:

src/Infrastructure/LeanCode.Kratos/KratosAuthenticationHandler.cs
```diff
@@ -26,6 +26,18 @@ public class KratosAuthenticationHandler<TOptions> : AuthenticationHandler<TOpti
 
     private readonly IFrontendApi api;
 
+#if NET8_0_OR_GREATER
```
jakubfijalkowski: Caaan we get rid of this ifdefs? We don't need them, really, since we will target .NET 8 onwards only here.

src/Infrastructure/LeanCode.Kratos/KratosWebHookHandlerBase.cs
```diff
@@ -0,0 +1,106 @@
...
+
+    private static bool VerifyApiKey(string left, string right)
+    {
+        var leftBytes = MemoryMarshal.AsBytes(left.AsSpan());
+        var rightBytes = MemoryMarshal.AsBytes(right.AsSpan());
+
+        return CryptographicOperations.FixedTimeEquals(leftBytes, rightBytes);
+    }
+
+    public static Task WriteIdentityResponseAsync(HttpContext ctx, Identity identity)
```
jakubfijalkowski, line 62: What is the intended usage here? Only the derived class should call it? If yes - let's make this a `protected` method.

src/Infrastructure/LeanCode.Kratos/KratosWebHookHandlerBase.cs
```diff
@@ -0,0 +1,106 @@
...
+}
+
+#if NET8_0_OR_GREATER
+[JsonSourceGenerationOptions(PropertyNamingPolicy = JsonKnownNamingPolicy.SnakeCaseLower)]
+#endif
+[JsonSerializable(typeof(IdentityResponseBody))]
+[JsonSerializable(typeof(ErrorResponseBody))]
+public partial class KratosWebHookHandlerBaseJsonSerializerContext : JsonSerializerContext;
+
+#pragma warning disable CA2227 // Collection properties should be read only
```
jakubfijalkowski: Can we `SuppressMessage` instead of going the pragma way? It's more readable.

src/Infrastructure/LeanCode.Kratos/Model/Identity.cs
```diff
@@ -0,0 +1,54 @@
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using LeanCode.Serialization;
+
+namespace LeanCode.Kratos.Model;
+
+#nullable disable warnings
```
jakubfijalkowski: Let's move this to the top?

src/Infrastructure/LeanCode.Kratos/Model/Identity.cs
```diff
@@ -0,0 +1,54 @@
...
+#pragma warning disable CA2227 // Collection properties should be read only
```
jakubfijalkowski: And make this a `SuppressMessage`?

jakubfijalkowski:
Plus, can we have tests for this one? At least the key validation should be tested.

Saancreed:
I'll see what can be done about tests :+1:

src/Infrastructure/LeanCode.Kratos/KratosWebHookHandlerBase.cs
Saancreed: In theory, there wouldn't be anything wrong with someone rolling their own base class (or not using any base class at all) and calling this there. But the intent was indeed about calling from subclasses so I can make it `protected` with little value being lost in the process.

src/Infrastructure/LeanCode.Kratos/KratosWebHookHandlerBase.cs
Saancreed: It really isn't, lots of noise just to get rid of a warning that doesn't make sense in this context.

src/Infrastructure/LeanCode.Kratos/KratosWebHookHandlerBase.cs
Saancreed: But it makes testing them much more annoying :slightly_frowning_face:

test/Infrastructure/LeanCode.Kratos.Tests/KratosWebHookHandlerTests.cs
```diff
@@ -0,0 +1,173 @@
...
+        Assert.Equal(timestampString, id.GetProperty("created_at"u8).GetString());
+        Assert.Equal(timestampString, id.GetProperty("updated_at"u8).GetString());
+        Assert.Equal("user", id.GetProperty("schema_id"u8).GetString());
+        Assert.Equal("https://auth.local.lncd.pl/schemas/dXNlcg", id.GetProperty("schema_url"u8).GetString());
+        Assert.Equal("active", id.GetProperty("state"u8).GetString());
+        Assert.Equal(email, id.GetProperty("traits"u8).GetProperty("email"u8).GetString());
+        Assert.Equal(JsonValueKind.Null, id.GetProperty("metadata_public"u8).ValueKind);
+        Assert.Equal(JsonValueKind.Null, id.GetProperty("metadata_admin"u8).ValueKind);
+
+        var recoveryAddresses = id.GetProperty("recovery_addresses");
```
Saancreed: Oops, should be `u8` here and below, will fix :see_no_evil:

test/Infrastructure/LeanCode.Kratos.Tests/KratosWebHookHandlerTests.cs
```diff
@@ -0,0 +1,173 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features;
+using Microsoft.Extensions.Primitives;
+using Xunit;
+
+namespace LeanCode.Kratos.Tests;
+
+public class KratosWebHookHandlerTests
+{
+    private sealed class KratosWebHookTestHandler : KratosWebHookHandlerBase
```
jakubfijalkowski: Let's move this out of the test class - it makes understanding the tests harder than it needs to be.

test/Infrastructure/LeanCode.Kratos.Tests/KratosWebHookHandlerTests.cs
```diff
@@ -0,0 +1,173 @@
...
+        features.Set<IHttpRequestFeature>(new HttpRequestFeature() { Headers = new HeaderDictionary(headers) });
+        features.Set<IHttpResponseFeature>(new HttpResponseFeature());
+        features.Set<IHttpResponseBodyFeature>(new StreamResponseBodyFeature(new MemoryStream()));
+        var httpContext = new DefaultHttpContext(features);
+        await handler.HandleAsync(httpContext);
+        return httpContext;
+    }
+
+    [Fact]
+    public async Task Responds_with_status_code_403_without_running_inner_handler_if_api_key_is_invalid()
```
jakubfijalkowski, line 15: And make the tests the first thing that you see. They matter, everything else is just a helper.

test/Infrastructure/LeanCode.Kratos.Tests/KratosWebHookHandlerTests.cs
```diff
@@ -0,0 +1,173 @@
...
+            MetadataAdmin = null,
+        };
+
+        var handler = PrepareHandler(ctx => KratosWebHookHandlerBase.WriteIdentityResponseAsync(ctx, identity));
+        var ctx = await RunAsync(handler, new() { [handler.ApiKeyHeaderName] = ApiKey });
+
+        Assert.Equal(200, ctx.Response.StatusCode);
+
+        ctx.Response.Body.Position = 0L;
+        using var document = await JsonDocument.ParseAsync(ctx.Response.Body);
```
jakubfijalkowski: Can you deserialize to `Identity` and use FluentAssertions to compare them? Would be much easier to read :)

test/Infrastructure/LeanCode.Kratos.Tests/KratosWebHookHandlerTests.cs
```diff
@@ -0,0 +1,173 @@
...
+                    Value = email,
+                    Status = AddressStatus.Completed,
+                    VerifiedAt = timestamp,
+                    Verified = true,
+                },
+            },
+            Traits = JsonSerializer.SerializeToElement(new Dictionary<string, string> { ["email"] = email }),
+            MetadataPublic = null,
+            MetadataAdmin = null,
+        };
```
jakubfijalkowski: Let's move this out of this test method.

src/Infrastructure/LeanCode.Kratos/KratosWebHookHandlerBase.cs
jakubfijalkowski: `new public` to the rescue, but whatever - we can live with these being public.

test/Infrastructure/LeanCode.Kratos.Tests/KratosWebHookHandlerTests.cs
Saancreed: I can deserialize to `Identity`, I don't have to use FluentAsssertions to compare them, but in any case I think it would defeat the purpose because it won't allow me to verify over-the-wire format, if something was broken in the serialization code and e.g. enums were serialized as numbers, Kratos could fail to understand our response and we would be unable to detect it in tests.

test/Infrastructure/LeanCode.Kratos.Tests/KratosWebHookHandlerTests.cs
Saancreed: Wouldn't moving it to the very bottom be enough? There's little reason for this class to be visible outside `KratosWebHookHandlerTests`.

test/Infrastructure/LeanCode.Kratos.Tests/KratosWebHookHandlerTests.cs
jakubfijalkowski: Ah, got it. Valid point. So, let's leave it as-is indeed.

jakubfijalkowski (approved):

PR #504, opened 2023-06-28
Forced/suggested updates

wojtek2288:
Still need to check if this will work with contracts generators.


src/Infrastructure/LeanCode.ClientsUpdates/LeanCode.ClientsUpdates.csproj
jakubfijalkowski, line 1: Nit: `LeanCode.ClientUpdates` or rather `LeanCode.ForceUpdate`.

src/Infrastructure/LeanCode.ClientsUpdates/LeanCode.ClientsUpdates.csproj
```diff
@@ -0,0 +1,9 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <ItemGroup>
+    <PackageReference Include="LeanCode.Contracts" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\CQRS\LeanCode.CQRS.Execution\LeanCode.CQRS.Execution.csproj" />
```
jakubfijalkowski:
```suggestion
    <ProjectReference Include="../../CQRS/LeanCode.CQRS.Execution/LeanCode.CQRS.Execution.csproj" />
```

src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSObjectsRegistrationSource.cs
```diff
@@ -10,17 +10,32 @@ internal class CQRSObjectsRegistrationSource
...
     {
         this.contractsCatalog = contractsCatalog;
         this.handlersCatalog = handlersCatalog;
 
-        Objects = FindCQRSObjects();
+        Objects = FindCQRSObjects(contractsCatalog, handlersCatalog);
     }
 
-    private IEnumerable<CQRSObjectMetadata> FindCQRSObjects()
+    public IEnumerable<CQRSObjectMetadata> AddAdditionalCQRSObjects(
```
jakubfijalkowski: I'm not sure about that, IMO we should not allow to modify the registration source. @marcin-dardzinski , what is your opinion on that?

src/CQRS/LeanCode.CQRS.AspNetCore/ServiceCollectionCQRSExtensions.cs
```diff
@@ -30,6 +32,28 @@ TypesCatalog handlersCatalog
 
         return new CQRSServicesBuilder(serviceCollection);
     }
+
+    public static CQRSServicesBuilder AddClientsUpdates(this IServiceCollection serviceCollection)
```
jakubfijalkowski: This should be placed in `LeanCode.ClientsUpdates` project, not here.

src/Infrastructure/LeanCode.ClientsUpdates/Contracts/PlatformDTO.cs
```diff
@@ -0,0 +1,7 @@
+namespace LeanCode.ClientsUpdates.Contracts;
+
+public enum PlatformDTO
+{
+    IOS,
+    Android,
```
jakubfijalkowski:
```suggestion
    IOS = 0,
    Android = 1,
```

test/CQRS/LeanCode.CQRS.AspNetCore.Tests.Integration/ClientsUpdatesTests.cs
```diff
@@ -0,0 +1,118 @@
+using System.Net;
```
marcin-dardzinski: I'm not sure if such tests should belong to this project - you can unit test handlers without the http fuss. 

Your one has the benefit of testing DI registrations, but I still think this can be done separately.

src/Infrastructure/LeanCode.ClientsUpdates/Contracts/VersionSupport.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    public PlatformDTO Platform { get; set; }
+    public string Version { get; set; } = default!;
+}
+
+public enum VersionSupportDTO
+{
+    UpdateRequired,
+    UpdateSuggested,
+    UpToDate,
```
jakubfijalkowski:
```suggestion
    UpdateRequired = 0,
    UpdateSuggested = 1,
    UpToDate = 2,
```

`enum`s in DTO should always be explicitly numbered, otherwise it is very easy to break backwards compatibility.

src/Infrastructure/LeanCode.ClientsUpdates/LeanCode.ClientsUpdates.csproj
jakubfijalkowski: Also, we _probably_ (not fully sure 'bout that) should split contracts and implementation projects. Contracts need to also be embedded in the project (just like `LeanCode.Chat` does). Ask @Saancreed how to do that.

src/Infrastructure/LeanCode.ClientsUpdates/LeanCode.ClientsUpdates.csproj
marcin-dardzinski: Flip'em to the other direction :upside_down_face:

test/CQRS/LeanCode.CQRS.AspNetCore.Tests.Integration/RemoteCQRSTestsBase.cs
```diff
@@ -16,6 +17,8 @@ namespace LeanCode.CQRS.AspNetCore.Tests.Integration;
 public abstract class RemoteCQRSTestsBase : IDisposable, IAsyncLifetime
 {
     private const string IsAuthenticatedHeader = "is-authenticated";
+    protected const string MinimumRequiredVersion = "1.0";
```
jakubfijalkowski: This is terribly, terribly not related with `RemoteCQRS`.

src/CQRS/LeanCode.CQRS.AspNetCore/ServiceCollectionCQRSExtensions.cs
marcin-dardzinski: I think we'll need more clever way of registering it. 
In essence you just need to add one assembly to the types catalog plus register some service. 

Maybe some extension method on CQRSServicesBuilder, but I need to think about it.

src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSObjectsRegistrationSource.cs
marcin-dardzinski: Well, at this point I think it's better to remove constructor parameters and have only one method `AddCQRSObjects(TypesCatalog, TypesCatalog)`.

By the way, what would happen now if you add the same assembly twice - I think it's gonna crash at some point (probably routing).

LeanCode.CoreLibrary.sln
```diff
@@ -173,11 +173,7 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "LeanCode.CQRS.Validation.Fl
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "LeanCode.CQRS.ScopedTransaction", "src\CQRS\LeanCode.CQRS.ScopedTransaction\LeanCode.CQRS.ScopedTransaction.csproj", "{6D962351-01E4-4A32-85C3-1B5A36B0C11C}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "LeanCode.Startup.MicrosoftDI", "src\Core\LeanCode.Startup.MicrosoftDI\LeanCode.Startup.MicrosoftDI.csproj", "{5243E4DB-E668-42C3-88F2-38F6135C5F66}"
-EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "LeanCode.Components.Autofac", "src\Core\LeanCode.Components.Autofac\LeanCode.Components.Autofac.csproj", "{47EBE467-DDAC-47D5-8669-F5B45AADF9E6}"
-EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "LeanCode.IntegrationTests", "test\LeanCode.IntegrationTests\LeanCode.IntegrationTests.csproj", "{061874B2-0AD2-4452-800D-723BBC75277D}"
```
marcin-dardzinski, line 180: Looks suspicious, those projects should be in solution.

src/Infrastructure/LeanCode.ClientsUpdates/LeanCode.ClientsUpdates.csproj
marcin-dardzinski: `LeanCode.ForceUpdate` :heart:

src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSObjectsRegistrationSource.cs
marcin-dardzinski: It's not an intrinsic problem - as long as we do it before we build routes. We just need to make sure no object is registered twice. 

I think this would be easier to achieve then to have injection points for `TypesCatalog`.

marcin-dardzinski: We decided to make registration source modifiable and expose it in `CQRSServicesBuilder` so that clients could extend it easily.

Client setup will look like this
```
services.AddCQRS(...).AddForceUpdate();
```

src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSObjectsRegistrationSource.cs
```diff
@@ -57,9 +61,10 @@ private IEnumerable<CQRSObjectMetadata> FindCQRSObjects()
             );
 
             resultObjects.Add(metadata);
+            Objects = Objects.Append(metadata);
```
marcin-dardzinski: 1. This looks like a perf nighmare - new collection per _every_ added item.
2. You still can add the same object twice.

I'd have a mutable hashset here (with proper equality comparer) and ignore the fact that `Objects` collection can be silently updated. Or just use `ImmutableHashSet`  .

src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSObjectsRegistrationSource.cs
```diff
@@ -2,25 +2,29 @@
...
-    private readonly TypesCatalog handlersCatalog;
+    private readonly IServiceCollection services;
 
-    public IEnumerable<CQRSObjectMetadata> Objects { get; }
+    public IEnumerable<CQRSObjectMetadata> Objects { get; private set; }
 
-    public CQRSObjectsRegistrationSource(TypesCatalog contractsCatalog, TypesCatalog handlersCatalog)
+    public CQRSObjectsRegistrationSource(
+        IServiceCollection services,
+        TypesCatalog contractsCatalog,
```
marcin-dardzinski: I think having this in constructor is redundant, let the caller explicitly call `AddCQRSObjects`

test/Infrastructure/LeanCode.ForceUpdate.Tests/ForceUpdateTests.cs
```diff
@@ -0,0 +1,118 @@
...
+using LeanCode.ForceUpdate.Contracts;
+using FluentAssertions;
+using Xunit;
+
+namespace LeanCode.ForceUpdate.Tests;
+
+public class ForceUpdateTests : ForceUpdateTestsBase
+{
+    [Fact]
+    public async Task Versions_correctly_returns_all_versions()
```
marcin-dardzinski: My previous comment still holds - you don't need the whole asp.net pipeline here, calling QH directly would be far less awkward.

My proposition is to build a container, resolve the handler and work on it, like this:
```
var services = new ServiceCollection();
services.AddCQRS().AddForceUpdate()
(register configs)

var sp = services.BuildServiceProvider();
var qh = services.GetRequiredService<QH>()
```
Then you can test di registration and handler easily.

One more nit: if you were to keep the full http pipeline, I'd pack the call in better method - there's no need to specify path and raw json content in every tests.

test/Infrastructure/LeanCode.ForceUpdate.Tests/ForceUpdateTestsBase.cs
```diff
@@ -0,0 +1,109 @@
...
+                            );
+                        });
+                    });
+            })
+            .Build();
+
+        server = host.GetTestServer();
+    }
+
+    private class Query1 : IQuery<QueryResult1> { }
```
marcin-dardzinski: I see the reason for this but empty `TypesCatalog` would be much better.

test/CQRS/LeanCode.CQRS.AspNetCore.Tests/CQRSObjectsRegistrationSourceTests.cs
```diff
@@ -62,6 +63,7 @@ private void AssertNotRegistered<T>()
     private static List<CQRSObjectMetadata> GetCQRSObjects()
     {
         return new CQRSObjectsRegistrationSource(
+            new ServiceCollection(),
```
marcin-dardzinski: Given that the responsibility of this class changed I'd also test if handlers are properly registered, either by building this service collection and trying to resolve handlers or via a mock.

src/Infrastructure/LeanCode.ForceUpdate/LeanCode.ForceUpdate.Services/CQRS/VersionSupportQH.cs
```diff
@@ -0,0 +1,57 @@
...
+    )
+    {
+        this.iOSConfiguration = iOSConfiguration;
+        this.androidConfiguration = androidConfiguration;
+        this.versionHandler = versionHandler;
+    }
+
+    public Task<VersionSupportDTO?> ExecuteAsync(HttpContext context, VersionSupport query)
+    {
+        if (!Version.TryParse(query.Version, out var version) || !Enum.IsDefined(typeof(PlatformDTO), query.Platform))
```
Saancreed:
```suggestion
        if (!Version.TryParse(query.Version, out var version) || !Enum.IsDefined<PlatformDTO>(query.Platform))
```

src/Infrastructure/LeanCode.ForceUpdate/LeanCode.ForceUpdate.Services/CQRS/VersionSupportQH.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            return Task.FromResult<VersionSupportDTO?>(null);
+        }
+
+        return query.Platform switch
```
Saancreed: I'd assign to a variable here, like `var result = query.Platform switch` and at the bottom do `return Task.FromResult(result);`

src/Infrastructure/LeanCode.ForceUpdate/LeanCode.ForceUpdate.Services/CQRS/VersionSupportQH.cs
```diff
@@ -0,0 +1,57 @@
...
+                            context
+                        )
+                ),
+            PlatformDTO.Android
+                => Task.FromResult<VersionSupportDTO?>(
+                    (VersionSupportDTO)
+                        versionHandler.CheckVersion(
+                            version,
+                            iOSConfiguration.MinimumRequiredVersion,
+                            iOSConfiguration.CurrentlySupportedVersion,
```
Saancreed: Shouldn't this use `androidConfiguration`?

```suggestion
                            androidConfiguration.MinimumRequiredVersion,
                            androidConfiguration.CurrentlySupportedVersion,
```

Saancreed (commented):
I feel like doing this without involving a manual call to some magic CQRS endpoint would be a bit more preferable but I could be the minority here. Anyway, see comments.

About contracts, generate them to a file called `LeanCode.Contracts.pb` and embed that in the assembly like so: https://github.com/leancodepl/contractsgenerator/blob/20f2bf1095a5b27095e2258090dc24e2a092e038/examples/project/embedded/embedded.csproj#L8-L10

src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSObjectsRegistrationSource.cs
```diff
@@ -2,25 +2,23 @@
...
 namespace LeanCode.CQRS.AspNetCore.Registration;
 
 internal class CQRSObjectsRegistrationSource
 {
-    private readonly TypesCatalog contractsCatalog;
-    private readonly TypesCatalog handlersCatalog;
+    private readonly IServiceCollection services;
 
-    public IEnumerable<CQRSObjectMetadata> Objects { get; }
+    public IEnumerable<CQRSObjectMetadata> Objects { get; private set; }
```
Saancreed:
```suggestion
    private readonly HashSet<CQRSObjectMetadata> objects = new(CQRSObjectMetadataEqualityComparer.Instance);

    public IReadOnlySet<CQRSObjectMetadata> Objects => objects;
```

And then below just do
```csharp
if (objects.Add(metadata))
{
    services.AddCQRSHandler(metadata);
}
```

`AddCQRSHandler` that accepts a single `CQRSObjectMetadata` doesn't seem to exist yet so go ahead and add one (and make `AddCQRSHandlers` call it in a loop instead).

src/CQRS/LeanCode.CQRS.Execution/CQRSObjectMetadata.cs
```diff
@@ -16,6 +16,32 @@ public CQRSObjectMetadata(CQRSObjectKind objectKind, Type objectType, Type resul
...
+
+        return x.ObjectKind == y.ObjectKind
+            && x.ObjectType == y.ObjectType
+            && x.ResultType == y.ResultType
+            && x.HandlerType == y.HandlerType;
+    }
+
+    public int GetHashCode(CQRSObjectMetadata obj)
+    {
+        return (obj.ObjectKind, obj.ObjectType, obj.ResultType, obj.HandlerType).GetHashCode();
```
Saancreed:
```suggestion
        return HashCode.Combine(obj.ObjectKind, obj.ObjectType, obj.ResultType, obj.HandlerType);
```

Maybe `CQRSObjectMetadata` should have been a `record class` instead?

src/Infrastructure/LeanCode.ForceUpdate.Contracts/VersionSupport.cs
```diff
@@ -0,0 +1,16 @@
+using LeanCode.Contracts;
+
+namespace LeanCode.ForceUpdate.Contracts;
+
+public class VersionSupport : IQuery<VersionSupportDTO?>
+{
+    public PlatformDTO Platform { get; set; }
+    public string Version { get; set; } = default!;
```
Saancreed: Imho `= default!;` in contracts is just noise, instead consider disabling nullable warnings in the entire `Contracts` project (assuming that `<Nullable>enable</Nullable>` in `Directory.Build.props` won't interfere :see_no_evil:).

src/CQRS/LeanCode.CQRS.AspNetCore/ServiceCollectionCQRSExtensions.cs
```diff
@@ -19,31 +19,39 @@ TypesCatalog handlersCatalog
...
 
-        return new CQRSServicesBuilder(serviceCollection);
+        return new CQRSServicesBuilder(serviceCollection, objectsSource);
     }
 }
 
 public class CQRSServicesBuilder
 {
-    private readonly IServiceCollection services;
+    public readonly IServiceCollection services;
```
Saancreed: Make it a get-only property if you need this to be `public`.

src/Infrastructure/LeanCode.ForceUpdate/LeanCode.ForceUpdate.Services/CQRS/VersionSupportQH.cs
```diff
@@ -0,0 +1,53 @@
...
+                    context
+                ),
+            PlatformDTO.Android
+                => versionHandler.CheckVersion(
+                    version,
+                    androidConfiguration.MinimumRequiredVersion,
+                    androidConfiguration.CurrentlySupportedVersion,
+                    context
+                ),
+            _ => throw new InvalidOperationException("Invalid platform"),
```
Saancreed: Throwing an exception on invalid input is just going to confuse clients if they get 500 in response when it's really their fault. How about logging an error _that includes the invalid value we got_ and returning `null` instead?

src/Infrastructure/LeanCode.ForceUpdate/LeanCode.ForceUpdate.Services/CQRS/VersionsQH.cs
```diff
@@ -0,0 +1,38 @@
...
+        this.iOSConfiguration = iOSConfiguration;
+        this.androidConfiguration = androidConfiguration;
+    }
+
+    public Task<List<VersionsDTO>> ExecuteAsync(HttpContext context, Versions query)
+    {
+        return Task.FromResult(
+            new List<VersionsDTO>
+            {
+                new VersionsDTO
```
Saancreed:
```suggestion
                new()
```
Same below.

src/Infrastructure/LeanCode.ForceUpdate.Contracts/PlatformDTO.cs
```diff
@@ -0,0 +1,7 @@
+namespace LeanCode.ForceUpdate.Contracts;
+
+public enum PlatformDTO
+{
+    IOS = 0,
+    Android = 1,
```
Saancreed: Any particular reason why is this not in lexicographic order? :smile:

src/Infrastructure/LeanCode.ForceUpdate/LeanCode.ForceUpdate.Services/VersionHandler.cs
```diff
@@ -0,0 +1,34 @@
+using Microsoft.AspNetCore.Http;
+
+namespace LeanCode.ForceUpdate.Services;
+
+public class VersionHandler
+{
+    public virtual VersionSupport CheckVersion(
```
Saancreed: Actually, I can probably think of some cases where overrides of this method would want to perform some async call (e.g. we want to force update for specific tenants only and we store that information in the database). Can we make it `virtual Task<VersionSupport> CheckVersionAsync` instead?

(Aaand that makes my previous suggestion about refactoring `Task.FromResult` no longer valid, sorry :sweat_smile:)

test/Infrastructure/LeanCode.ForceUpdate.Tests/ForceUpdateTests.cs
```diff
@@ -0,0 +1,110 @@
...
+        var result = await handler.ExecuteAsync(new DefaultHttpContext(), new());
+
+        result
+            .Should()
+            .BeEquivalentTo(
+                new List<VersionsDTO>
+                {
+                    new VersionsDTO
+                    {
+                        Platform = PlatformDTO.IOS,
```
Saancreed: We probably shouldn't rely on order of entries here, but making it a dictionary won't work without changing `Platform` to be `string`, or at least something serializable to `string` :weary:

I'd suggest requiring the client to pass `Platform` as query parameter for `Versions` but that puts in question separating that from `VersionSupport` query :weary:²

src/Infrastructure/LeanCode.ForceUpdate/LeanCode.ForceUpdate.Services/CQRS/VersionSupportQH.cs
wojtek2288: It will never go there as it's checked in `if (!Version.TryParse(query.Version, out var version) || !Enum.IsDefined<PlatformDTO>(query.Platform))` added log there.

Saancreed: Ah, the fact that it's a proper `InvalidOperationException` instead of something like `UnreachableException` led me the believe that path was actually reachable. My bad.

src/CQRS/LeanCode.CQRS.AspNetCore/ServiceCollectionCQRSExtensions.cs
```diff
@@ -19,31 +19,39 @@ TypesCatalog handlersCatalog
...
 
-        return new CQRSServicesBuilder(serviceCollection);
+        return new CQRSServicesBuilder(serviceCollection, objectsSource);
     }
 }
 
 public class CQRSServicesBuilder
 {
-    private readonly IServiceCollection services;
+    public IServiceCollection services { get; }
```
Saancreed:
```suggestion
    public IServiceCollection Services { get; }
```

Convention, though I'm surprised your IDE didn't scream at you for leaving this all lowercase.

src/Infrastructure/LeanCode.ForceUpdate/LeanCode.ForceUpdate.Services/CQRS/VersionSupportQH.cs
```diff
@@ -0,0 +1,68 @@
...
+        this.iOSConfiguration = iOSConfiguration;
+        this.androidConfiguration = androidConfiguration;
+        this.versionHandler = versionHandler;
+    }
+
+    public async Task<VersionSupportDTO?> ExecuteAsync(HttpContext context, VersionSupport query)
+    {
+        if (!Version.TryParse(query.Version, out var version) || !Enum.IsDefined<PlatformDTO>(query.Platform))
+        {
+            logger.Warning("Invalid input: Version={Version}, Platform={Platform}", query.Version, query.Platform);
```
Saancreed: Technically, the `Version=` and `Platform=` parts are not needed because the log is already enriched with property names taken from `{Version}` and `{Platform}`.

src/Infrastructure/LeanCode.ForceUpdate.Contracts/VersionSupport.cs
```diff
@@ -0,0 +1,29 @@
+using LeanCode.Contracts;
+
+namespace LeanCode.ForceUpdate.Contracts;
+
+public class VersionSupport : IQuery<VersionSupportDTO>
+{
+    public PlatformDTO Platform { get; set; }
+    public string Version { get; set; }
```
Saancreed, line 8: If this and `Result` below were made nullable, we could allow querying for supported version range for callers that aren't actually those clients and don't have any version assigned.

But I'm not sure how much value is there in this use case (e.g. calling from Postman without faking the version?) so feel free to tell me there's no point and ignore this comment.

Saancreed (approved):
Just nitpicks, otherwise LGTM (but make sure Dardzin and/or Fiołek are okay with this too).

src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSObjectsRegistrationSource.cs
```diff
@@ -2,25 +2,23 @@
...
+using Microsoft.Extensions.DependencyInjection;
 
 namespace LeanCode.CQRS.AspNetCore.Registration;
 
 internal class CQRSObjectsRegistrationSource
 {
-    private readonly TypesCatalog contractsCatalog;
-    private readonly TypesCatalog handlersCatalog;
+    private readonly IServiceCollection services;
+    private readonly HashSet<CQRSObjectMetadata> objects = new();
```
marcin-dardzinski: I wonder if we should care more about equality comparison here - _technically_ you are now able to register one command with two handlers from separate assemblies.

This is a rather not probable case, but I'd compare only the object type here and silently ignore duplicates.

src/Infrastructure/LeanCode.ForceUpdate/LeanCode.ForceUpdate.Services/VersionHandler.cs
```diff
@@ -0,0 +1,34 @@
+using Microsoft.AspNetCore.Http;
+
+namespace LeanCode.ForceUpdate.Services;
+
+public class VersionHandler
+{
+    public virtual Task<VersionSupportResult> CheckVersionAsync(
```
marcin-dardzinski: I'll be very nitty here, but I don't like how this work with `VersionSupportQH` :upside_down_face: 

I guess this is virtual so that projects may change the logic here freely. I think the interface exposed is rigid though, because it expects you to pass minimum required versions.

What if you have just `CheckVersionAsync(Version version, Platform platform, HttpContext context)` and get minimal versions as dependencies?

src/Infrastructure/LeanCode.ForceUpdate/CQRSServicesBuilderCQRSExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+using LeanCode.ForceUpdate.Services;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace LeanCode.ForceUpdate;
+
+public static class CQRSServicesBuilderCQRSExtensions
+{
+    public static CQRSServicesBuilder AddForceUpdate(this CQRSServicesBuilder cqrsServicesBuilder)
+    {
+        cqrsServicesBuilder.services.AddTransient<VersionHandler>();
```
marcin-dardzinski: Since `VersionSupportQH` will throw if you don't register version configurations and I don't see any way to dynamically reload support versions now, maybe let's require versions to be passed explicitly here and register them as singletons?

src/Infrastructure/LeanCode.ForceUpdate/CQRSServicesBuilderCQRSExtensions.cs
```diff
@@ -9,9 +9,24 @@ namespace LeanCode.ForceUpdate;
 
 public static class CQRSServicesBuilderCQRSExtensions
 {
-    public static CQRSServicesBuilder AddForceUpdate(this CQRSServicesBuilder cqrsServicesBuilder)
+    public static CQRSServicesBuilder AddForceUpdate(
+        this CQRSServicesBuilder cqrsServicesBuilder,
+        Version androidMinimumRequiredVersion,
```
marcin-dardzinski: It would be easier for the caller to have `AndroidVersionsConfiguration` / `IOSVersionsConfiguration` here

src/Infrastructure/LeanCode.ForceUpdate/LeanCode.ForceUpdate.Services/VersionHandler.cs
```diff
@@ -0,0 +1,52 @@
...
+    private readonly AndroidVersionsConfiguration androidConfiguration;
+    private readonly IOSVersionsConfiguration iOSConfiguration;
+
+    public VersionHandler(AndroidVersionsConfiguration androidConfiguration, IOSVersionsConfiguration iOSConfiguration)
+    {
+        this.androidConfiguration = androidConfiguration;
+        this.iOSConfiguration = iOSConfiguration;
+    }
+
+    public virtual Task<VersionSupportResult> CheckVersionAsync(Version version, Platform platform, HttpContext context)
```
jakubfijalkowski: Why is this `async`?

src/Infrastructure/LeanCode.ForceUpdate/LeanCode.ForceUpdate.Services/VersionHandler.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        var (minimumRequiredVersion, currentlySupportedVersion) = platform switch
+        {
+            Platform.Android
+                => (androidConfiguration.MinimumRequiredVersion, androidConfiguration.CurrentlySupportedVersion),
+            Platform.IOS => (iOSConfiguration.MinimumRequiredVersion, iOSConfiguration.CurrentlySupportedVersion),
+            _ => throw new InvalidOperationException($"Invalid platform: {platform}"),
```
jakubfijalkowski:
```suggestion
            _ => throw new InvalidOperationException($"Invalid platform: {platform}."),
```

src/Infrastructure/LeanCode.ForceUpdate/LeanCode.ForceUpdate.Services/CQRS/VersionSupportQH.cs
```diff
@@ -0,0 +1,58 @@
...
+            PlatformDTO.Android
+                => new VersionSupportDTO
+                {
+                    CurrentlySupportedVersion = androidConfiguration.CurrentlySupportedVersion.ToString(),
+                    MinimumRequiredVersion = androidConfiguration.MinimumRequiredVersion.ToString(),
+                    Result = (VersionSupportResultDTO)(
+                        await versionHandler.CheckVersionAsync(version, (Platform)query.Platform, context)
+                    ),
+                },
+            _ => throw new InvalidOperationException($"Invalid platform: {query.Platform}"),
```
jakubfijalkowski:
```suggestion
            _ => throw new InvalidOperationException($"Invalid platform: {query.Platform}."),
```

src/Infrastructure/LeanCode.ForceUpdate/LeanCode.ForceUpdate.Services/VersionHandler.cs
wojtek2288: As @Saancreed said above:
> Actually, I can probably think of some cases where overrides of this method would want to perform some async call (e.g. we want to force update for specific tenants only and we store that information in the database). Can we make it `virtual Task<VersionSupport> CheckVersionAsync` instead?
> 
> (Aaand that makes my previous suggestion about refactoring `Task.FromResult` no longer valid, sorry sweat_smile)

jakubfijalkowski: @Saancreed what kind of usecases you had in mind? I can't find one :(

Saancreed: > we want to force update for specific tenants only and we store that information in the database

Like, we have a new feature that's not yet available for everyone which needs a new version of the application to work correctly, so we'd need to force an update only for some users that actually have that feature enabled and there's no need for everyone else to update just yet.

Yes, this is probably not something that will come up very often but restricting this to just sync calls feels a bit shortsighted to me. Even having this as `ValueTask` that usually completes synchronously would be preferable to reinventing the wheel once an actual usecase is found.

jakubfijalkowski: I fear that this is not _the_ feature for this. The package is meant to be super-simple, without many features (esp. that I think more advanced features would need _more_ data in the query).

Nevertheless, I agree - we should not limit ourselves.

But let's move to `ValueTask` @wojtek2288 :P

src/Infrastructure/LeanCode.ForceUpdate/LeanCode.ForceUpdate.Services/VersionHandler.cs
```diff
@@ -0,0 +1,52 @@
...
+            return Task.FromResult(VersionSupportResult.UpdateSuggested);
+        }
+        else
+        {
+            return Task.FromResult(VersionSupportResult.UpToDate);
+        }
+    }
+}
+
+public enum Platform
```
jakubfijalkowski: We don't need to duplicate the `enum`s, we don't gain anything for having them in contracts and here. Let's use normal DTOs.

jakubfijalkowski (changes requested):

src/Infrastructure/LeanCode.ForceUpdate/LeanCode.ForceUpdate.Services/CQRS/VersionSupportQH.cs
```diff
@@ -0,0 +1,54 @@
...
+    {
+        if (!Version.TryParse(query.Version, out var version) || !Enum.IsDefined<PlatformDTO>(query.Platform))
+        {
+            logger.Warning("Invalid input: {Version}, {Platform}", query.Version, query.Platform);
+            return null;
+        }
+
+        return query.Platform switch
+        {
+            PlatformDTO.IOS
```
marcin-dardzinski: Nit: I think extracting `var (min, current) = GetVersions(platform)` would read nicer.

marcin-dardzinski (approved):

src/Infrastructure/LeanCode.ForceUpdate.Contracts/VersionSupport.cs
jakubfijalkowski: > But I'm not sure how much value is there in this use case (e.g. calling from Postman without faking the version?) so feel free to tell me there's no point and ignore this comment.

That's not the use-case for this lib. It solves one single thing and we should stick with that.

jakubfijalkowski (approved):

docs/guides/02_force_update.md
```diff
@@ -0,0 +1,119 @@
...
+                new Version(AndroidMinimumRequiredVersion),
+                new Version(AndroidCurrentlySupportedVersion),
+                new Version(IOSMinimumRequiredVersion),
+                new Version(IOSCurrentlySupportedVersion));
+    // ...
+}
+```
+
+## Version support
+After configuation above. `VersionSupport` query is created and available at `/cqrs/query/LeanCode.ForceUpdate.Contracts.VersionSupport`.
```
jakubfijalkowski:
```suggestion

After configuation above. `VersionSupport` query is created and available at `/cqrs/query/LeanCode.ForceUpdate.Contracts.VersionSupport`.
```

And let's link to the `VersionSupport` file, not copy the code - it will get outdated.

docs/guides/02_force_update.md
```diff
@@ -0,0 +1,119 @@
...
+{
+    UpdateRequired = 0,
+    UpdateSuggested = 1,
+    UpToDate = 2,
+}
+```
+The query takes `Platform` (either IOS or Android) and the `Version` of the client app as parameters. It returns whether the client's version is supported, `CurrentlySupportedVersion` and `MinimumRequiredVersion` for specified provider. If the client's version is below the `MinimumRequiredVersion`, the response will indicate that an update is needed. If the client's version is greater than or equal to the `MinimumRequiredVersion` and less than the `CurrentlySupportedVersion`, the response will suggest an update. If the app version is greater than or equal to the `CurrentlySupportedVersion`, the response will indicate that the app is up to date. It's also possible to override this behavior by overriding the `CheckVersion` method in the `VersionHandler` class, responsible for version checking:
+
+```csharp
+public class VersionHandler
```
jakubfijalkowski: Let's describe _intent_ and general idea in docs, not the real code - it will get outdated soon.

docs/guides/02_force_update.md
```diff
@@ -0,0 +1,119 @@
...
+}
+
+public enum VersionSupportResult
+{
+    UpdateRequired,
+    UpdateSuggested,
+    UpToDate,
+}
+```
+Below is an example of overriding the `CheckVersion` method:
```
jakubfijalkowski: We don't need to be _that_ elaborate in docs - saying that you can provide your own `VersionHandler` will be enough. :)

jakubfijalkowski (changes requested):
Okay, we need to adjust docs (forgot to check that previously).

jakubfijalkowski (approved):

PR #505, opened 2023-06-28
Prefixed ulid source generator

marcin-dardzinski:
I suggest starting the reading from the second commit onwards, the first one is vendoring Ulid package (with some auto refactoring)


Directory.Build.targets
```diff
@@ -110,6 +110,8 @@
     <PackageReference Update="MassTransit.EntityFrameworkCore" Version="$(MassTransitVersion)" />
 
     <PackageReference Update="OpenTelemetry" Version="1.3.1" />
+
+    <PackageReference Update="NUlid" Version="1.3.2" />
```
Saancreed, line 114: I understand this is an alternative implementation used only for testing purposes?

src/Domain/LeanCode.DomainModels/LeanCode.DomainModels.csproj
```diff
@@ -1,3 +1,5 @@
 <Project Sdk="Microsoft.NET.Sdk">
-
+  <PropertyGroup>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
Saancreed: :thinking:

Let's see if we can get rid of it.

src/Domain/LeanCode.DomainModels/Ulids/Ulid.cs
```diff
@@ -0,0 +1,884 @@
...
+
+/// <summary>
+/// Represents a Universally Unique Lexicographically Sortable Identifier (ULID).
+/// Spec: https://github.com/ulid/spec
+/// </summary>
+[StructLayout(LayoutKind.Explicit, Size = 16)]
+[DebuggerDisplay("{ToString(),nq}")]
+[System.Text.Json.Serialization.JsonConverter(typeof(UlidJsonConverter))]
+[System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]
+public struct Ulid : IEquatable<Ulid>, IComparable<Ulid>, ISpanFormattable, ISpanParsable<Ulid>
```
Saancreed: `readonly`? Possibly also make it a `record`? Possibly also implement `IUtf8SpanFormattable`?

src/Domain/LeanCode.DomainModels/Ulids/Ulid.cs
```diff
@@ -0,0 +1,884 @@
...
+{
+    // https://en.wikipedia.org/wiki/Base32
+    private static readonly char[] Base32Text = "0123456789ABCDEFGHJKMNPQRSTVWXYZ".ToCharArray();
```
Saancreed: I'd make all the private readonly collections immutable if possible.

src/Domain/LeanCode.DomainModels/Ulids/Ulid.cs
```diff
@@ -0,0 +1,884 @@
...
+        26,
+        255,
+        27,
+        28,
+        29,
+        30,
+        31
+    };
+
+    private static readonly DateTimeOffset UnixEpoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
```
Saancreed: Consider using [`DateTimeOffset.UnixEpoch`](https://learn.microsoft.com/en-us/dotnet/api/system.datetimeoffset.unixepoch?view=net-8.0) instead?

src/Domain/LeanCode.DomainModels/Ulids/Ulid.cs
```diff
@@ -0,0 +1,884 @@
...
+
+    [FieldOffset(14)]
+    private readonly byte randomness8;
+
+    [FieldOffset(15)]
+    private readonly byte randomness9;
+
+    [IgnoreDataMember]
+    [SuppressMessage("?", "CA1819", Justification = "Fresh array is allocated per call")]
+    public byte[] Random =>
```
Saancreed, line 252: I'd change return type to `ROS<byte>` or `ROM<byte>` or something similar and reimplement it as a read-only view over current instance's data. The copy is probably not always needed.

src/Domain/LeanCode.DomainModels.Generators/IdSource.cs
```diff
@@ -140,6 +144,114 @@ public static bool IsValid([NotNullWhen(true)] string? v)
...
+                && Ulid.TryParse(span[{{prefix.Length + 1}}..], out rawUlid);
+        }
+
+        public static bool IsValid([NotNullWhen(true)] string? v)
+        {
+            return TryDeconstruct(v.AsSpan(), out _);
+        }
+
+        public bool Equals({{data.TypeName}} other) => Value.Equals(other.Value, StringComparison.Ordinal);
+        public int CompareTo({{data.TypeName}} other) => string.Compare(Value, other.Value, StringComparison.Ordinal);
```
Dragemil, line 238: I guess you could compare just the `Ulid` parts here, which might've been more efficient.

src/Domain/LeanCode.DomainModels/Ulids/Ulid.cs
```diff
@@ -0,0 +1,884 @@
...
+        {
+            Span<byte> buffer = stackalloc byte[8];
+            buffer[0] = timestamp5;
+            buffer[1] = timestamp4;
+            buffer[2] = timestamp3;
+            buffer[3] = timestamp2;
+            buffer[4] = timestamp1;
+            buffer[5] = timestamp0; // [6], [7] = 0
+
+            var timestampMilliseconds = Unsafe.As<byte, long>(ref MemoryMarshal.GetReference(buffer));
```
Saancreed: What about endianness here? Wouldn't it be safer to use something like `BinaryPrimitives.ReadInt64LittleEndian` instead?

src/Domain/LeanCode.DomainModels.Generators/IdSource.cs
```diff
@@ -140,6 +144,114 @@ public static bool IsValid([NotNullWhen(true)] string? v)
...
+                && span[{{prefix.Length}}] == Separator
+                && Ulid.TryParse(span[{{prefix.Length + 1}}..], out rawUlid);
+        }
+
+        public static bool IsValid([NotNullWhen(true)] string? v)
+        {
+            return TryDeconstruct(v.AsSpan(), out _);
+        }
+
+        public bool Equals({{data.TypeName}} other) => Value.Equals(other.Value, StringComparison.Ordinal);
```
Dragemil, line 237: You here could compare only `Ulid` parts.

src/Domain/LeanCode.DomainModels/Ulids/Ulid.cs
```diff
@@ -0,0 +1,884 @@
...
+            return DateTimeOffset.FromUnixTimeMilliseconds(timestampMilliseconds);
+        }
+    }
+
+    internal Ulid(long timestampMilliseconds, XorShift64 random)
+        : this()
+    {
+        // Get memory in stack and copy to ulid(Little->Big reverse order).
+        ref var firstByte = ref Unsafe.As<long, byte>(ref timestampMilliseconds);
```
Saancreed: Again, what if `long` is already BE?

src/Domain/LeanCode.DomainModels/Ulids/Ulid.cs
```diff
@@ -0,0 +1,884 @@
...
+    }
+
+    // HACK: We assume the layout of a Guid is the following:
+    // Int32, Int16, Int16, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8
+    // source: https://github.com/dotnet/runtime/blob/4f9ae42d861fcb4be2fcd5d3d55d5f227d30e723/src/libraries/System.Private.CoreLib/src/System/Guid.cs
+    public Ulid(Guid guid)
+    {
+        Span<byte> buf = stackalloc byte[16];
+        MemoryMarshal.Write(buf, ref guid);
+        if (BitConverter.IsLittleEndian)
```
Saancreed, line 362: I think `Guid.TryWriteBytes` handles some of that nonsense for you, even if that would make the code a bit slower on BE machines.

Dragemil (approved):

src/Domain/LeanCode.DomainModels/Ulids/Ulid.cs
```diff
@@ -0,0 +1,884 @@
...
+        return true;
+    }
+
+    public string ToBase64(Base64FormattingOptions options = Base64FormattingOptions.None)
+    {
+        var buffer = ArrayPool<byte>.Shared.Rent(16);
+        try
+        {
+            TryWriteBytes(buffer);
+            return Convert.ToBase64String(buffer, options);
```
Saancreed: `Convert.ToBase64String` has an overload that accepts `ROS<byte>` so you could avoid this by stackalloc'ing 16 bytes, writing there and converting that.

src/Domain/LeanCode.DomainModels/Ulids/Ulid.cs
```diff
@@ -0,0 +1,884 @@
...
+            return false;
+        }
+    }
+
+    // Comparable/Equatable
+
+    public override unsafe int GetHashCode()
+    {
+        // Simply XOR, same algorithm of Guid.GetHashCode
+        fixed (void* p = &this.timestamp0)
```
Saancreed: Uhh okay, this is a bit mindbending. I think we could get away with `Unsafe.ReadUnaligned(Unsafe.Add(…))` but I'll have to verify that.

src/Domain/LeanCode.DomainModels/Ulids/Ulid.cs
```diff
@@ -0,0 +1,884 @@
...
+        {
+            var a = (int*)p;
+            return (*a) ^ *(a + 1) ^ *(a + 2) ^ *(a + 3);
+        }
+    }
+
+    public unsafe bool Equals(Ulid other)
+    {
+        // readonly struct can not use Unsafe.As...
+        fixed (byte* a = &this.timestamp0)
```
Saancreed: How about we just created a `ROS<byte>` over both instances and then compared their content?

src/Domain/LeanCode.DomainModels/Ulids/Ulid.cs
```diff
@@ -0,0 +1,884 @@
...
+    public static bool operator >(Ulid a, Ulid b) => a.CompareTo(b) > 0;
+
+    public static bool operator >=(Ulid a, Ulid b) => a.CompareTo(b) >= 0;
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    private static int GetResult(byte me, byte them) => me < them ? -1 : 1;
+
+    public int CompareTo(Ulid other)
+    {
+        if (this.timestamp0 != other.timestamp0)
```
Saancreed: And here, reinterpret memory as `ROS<byte>` and run compare on that with `SequenceCompareTo`. Might be a bit slower, but I'd be okay with that.

src/Domain/LeanCode.DomainModels/Ulids/UlidRandomProvider.cs
```diff
@@ -0,0 +1,71 @@
...
+    public static XorShift64 GetXorShift64()
+    {
+        return xorShift ??= CreateXorShift64();
+    }
+
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    private static XorShift64 CreateXorShift64()
+    {
+        using var rng = RandomNumberGenerator.Create();
+        Span<byte> buffer = stackalloc byte[sizeof(ulong)];
```
Saancreed, line 46: Isn't `sizeof` unsafe? :open_mouth:

If it is, you could replace this with `Unsafe.SizeOf<ulong>()` or just dump literal `8` there.

Saancreed (commented):
I have a few suggestions :see_no_evil:

By the way, didn't we want to keep vendored `Ulid` internal or was that just my imagination?

src/Domain/LeanCode.DomainModels.Generators/IdSource.cs
marcin-dardzinski: Nah, string comparison is vectorized anyway, little to zero gain but a readability cost

Directory.Build.targets
marcin-dardzinski: Yes

src/Domain/LeanCode.DomainModels/Ulids/Ulid.cs
marcin-dardzinski: Some marshal code was complaining after making it `readonly`

marcin-dardzinski (commented):
>By the way, didn't we want to keep vendored Ulid internal or was that just my imagination?

It has to be `public` to work with source generators.

Noted all the _modernization_ comments - I'll create a sub PR addressing them, they'll be better trackable.

src/Domain/LeanCode.DomainModels/Ulids/Ulid.cs
marcin-dardzinski: Giving it a second thought I don't think there's a safe way to convert it into ROS. If the method is not inlined the stack frame that we'll be pointing to will be invalidated. Maybe a static method with `ref` arg would help.

Still, I think it might be risky, say you get a span to the id in some array and then swap it with some other id.

On the other hand, nobody uses this property now so we can change it into `TryGetRandom(Span<byte>)` or just get rid of it :upside_down_face:

jakubfijalkowski (approved):
:+1: to all the comments @Saancreed left. Overall LGTM.

Saancreed (approved):

PR #506, opened 2023-07-03
Ulid modernization

marcin-dardzinski:
Addressing ulid intrinsic comments from #505

Some more tests should probably be necessary...

### DIFF
--- a/src/Domain/LeanCode.DomainModels/Ulids/Ulid.cs
+++ b/src/Domain/LeanCode.DomainModels/Ulids/Ulid.cs
@@ -22,6 +22,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
  */
 
+using System.Buffers.Binary;
 using System.Collections.Immutable;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
@@ -40,7 +41,12 @@ namespace LeanCode.DomainModels.Ulids;
 [DebuggerDisplay("{ToString(),nq}")]
 [System.Text.Json.Serialization.JsonConverter(typeof(UlidJsonConverter))]
 [System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)]
-public struct Ulid : IEquatable<Ulid>, IComparable<Ulid>, ISpanFormattable, ISpanParsable<Ulid>, IUtf8SpanFormattable
+public readonly record struct Ulid
+    : IEquatable<Ulid>,
+        IComparable<Ulid>,
+        ISpanFormattable,
+        ISpanParsable<Ulid>,
+        IUtf8SpanFormattable
 {
     public const int BytesLength = 26;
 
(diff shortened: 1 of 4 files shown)


src/Domain/LeanCode.DomainModels/Ulids/Ulid.cs
```diff
@@ -576,179 +573,131 @@ public bool TryWriteBytes(Span<byte> destination)
...
-        span[18] = Base32Bytes[(randomness5 & 248) >> 3];
-        span[19] = Base32Bytes[((randomness5 & 7) << 2) | ((randomness6 & 192) >> 6)];
-        span[20] = Base32Bytes[(randomness6 & 62) >> 1];
-        span[21] = Base32Bytes[((randomness6 & 1) << 4) | ((randomness7 & 240) >> 4)];
-        span[22] = Base32Bytes[((randomness7 & 15) << 1) | ((randomness8 & 128) >> 7)];
-        span[23] = Base32Bytes[(randomness8 & 124) >> 2];
-        span[24] = Base32Bytes[((randomness8 & 3) << 3) | ((randomness9 & 224) >> 5)];
-
-        return true;
+        return string.Create(BytesLength, this, (span, ulid) => ulid.TryFormat(span, out _, "", null));
```
Saancreed: Using `BytesLength` in this context suggests (and rightfully so) that this isn't actually length in _bytes_ but in base32 digits which all happen to be encodable with a single UTF8 code unit or a single UTF16 code unit each.

`LengthInChars` would be less incorrect, if we allow for overloaded use of term "character" :stuck_out_tongue:

Or [`LengthInTextElements`](https://learn.microsoft.com/en-us/dotnet/api/system.globalization.stringinfo.lengthintextelements?view=net-7.0) :see_no_evil:

src/Domain/LeanCode.DomainModels/Ulids/Ulid.cs
```diff
@@ -759,92 +708,19 @@ public override bool Equals(object? obj)
...
+        var thisSpan = AsSpan(in this);
+        var otherSpan = AsSpan(in other);
+        return thisSpan.SequenceCompareTo(otherSpan);
+    }
 
-        if (this.randomness9 != other.randomness9)
-        {
-            return GetResult(this.randomness9, other.randomness9);
-        }
+    private static ReadOnlySpan<byte> AsSpan(in Ulid ulid)
```
Saancreed, line 721: `[MemoryImpl(MemoryImplOptions.AggressingInlining)]` on this? :see_no_evil:

src/Domain/LeanCode.DomainModels/Ulids/Ulid.cs
```diff
@@ -367,23 +361,26 @@ public Ulid(Guid guid)
...
-            tmp = buf[1];
-            buf[1] = buf[2];
-            buf[2] = tmp;
-            tmp = buf[4];
-            buf[4] = buf[5];
-            buf[5] = tmp;
-            tmp = buf[6];
-            buf[6] = buf[7];
-            buf[7] = tmp;
+            buf[0..4].Reverse();
```
jakubfijalkowski, line 366: Can we get a comment here? The non-symmetry is quite mind-bending. ^^'

jakubfijalkowski (approved):

Saancreed (approved):
Two suggestions, otherwise LGTM.

PR #507, opened 2023-07-04
Use plain Task.Delay when waiting for MT outbox in tests

marcin-dardzinski:
Two problems with `IBusOutboxNotification`:
1. It's not thread safe
2. Actually I misunderstood it's purpose - it's used by outbox to wait for messages to send so it marks the start of outbox publication and not the end :see_no_evil:   

So, we resort to good'ol sleepin'.


jakubfijalkowski (approved):
I hope that this is The Right Thing :tm: .

jakubfijalkowski (approved):

jakubfijalkowski (approved):

PR #509, opened 2023-07-06
Rename `TimeProvider` -> `Time`

marcin-dardzinski:
Since `System.TimeProvider` violently intruded and it doesn't fit our use case, let's roll back to `Time`


Saancreed:
> it doesn't fit our use case

Please explain? Why can't we instead make our `FixedTimeProvider` subclass from `System.TimeProvider`? Or even dump it entirely in favor of `FakeTimeProvider`?

marcin-dardzinski:
Summing up offline discussion - we still need `Time` class for ambient context, but we can remove `ITimeProvider` interface and use `System.TimeProvider` instead. We will also need async local aware `FakeTimeProvider` class to replicate existing behavior.

@Saancreed will pick up the topic.

jakubfijalkowski (approved):

Saancreed (approved):

PR #510, opened 2023-07-06
Replace ITimeProvider with System.TimeProvider

Saancreed:
I love deleting code.

Some tests were kept to prove that it fits our use case, but I suppose we can dump them as well.


src/Domain/LeanCode.TimeProvider/Time.cs
```diff
@@ -2,10 +2,12 @@ namespace LeanCode.TimeProvider;
...
 
-    public static DateTime Now => timeProvider.Now;
-    public static DateTimeOffset NowWithOffset => timeProvider.NowWithOffset;
+    public static System.TimeProvider CurrentProvider => Provider.Value ?? System.TimeProvider.System;
 
-    public static void Use(ITimeProvider newProvider) => timeProvider = newProvider;
+    public static DateTime Now => NowWithOffset.UtcDateTime;
+    public static DateTimeOffset NowWithOffset => CurrentProvider.GetUtcNow();
+
+    public static void Use(System.TimeProvider? newProvider) => Provider.Value = newProvider;
```
marcin-dardzinski: I'm not convinced this is a good place for async locals - what if you really want to change provider globally? 
On the other hand, setting it on startup _should_ be respected and I don't see a use case to change it later, except tests.

jakubfijalkowski: I don't think you will be able to set this globally if the async local is at this level. The async machine at the start of the generic host program is really tricky, with multiple independent tasks and breaks in the async chain (last time I checked, so couple of years). :(

src/Domain/LeanCode.TimeProvider/Time.cs
```diff
@@ -2,10 +2,12 @@ namespace LeanCode.TimeProvider;
...
-    private static ITimeProvider timeProvider = new UtcTimeProvider();
+    private static readonly AsyncLocal<System.TimeProvider?> Provider = new();
 
-    public static DateTime Now => timeProvider.Now;
-    public static DateTimeOffset NowWithOffset => timeProvider.NowWithOffset;
+    public static System.TimeProvider CurrentProvider => Provider.Value ?? System.TimeProvider.System;
 
-    public static void Use(ITimeProvider newProvider) => timeProvider = newProvider;
+    public static DateTime Now => NowWithOffset.UtcDateTime;
+    public static DateTimeOffset NowWithOffset => CurrentProvider.GetUtcNow();
```
jakubfijalkowski: This should not force you to UTC. Same with `Now`. The provider should define if you get +00:00 or any other offset.

Saancreed: In case of `DateTime` I disagree, the only kind of timestamp that it makes sense to return is UTC one, everything else is worthless. In case of `DateTimeOffset`, sure, I can swap this to call `GetLocalNow` instead. However, do we want the default implementation to return non-UTC timestamps or should I write a `TimeProvider` that pretends to be in `+00:00` timezone for our purposes?

src/Domain/LeanCode.TimeProvider/Time.cs
marcin-dardzinski: Well, I was thinking about dedicated `AsyncLocalTimeProvider` adapter from the very beginning, so that `Time` has only one static instance.

The question is, do we need to care about async startup issues, given our use case (system provider by default, overridden only for tests)?

Saancreed: Wouldn't exposing and actually using a global override be incredibly thread-unsafe in all but the most trivial use cases? 

Okay, how about two fields, one `static TimeProvider` and one `static AsyncLocal<TimeProvider>` with separate methods to replace them, `UseGlobally` and `UseLocally`, where the former fully resets the latter?

src/Domain/LeanCode.TimeProvider/Time.cs
jakubfijalkowski: Datetime has two kinds, and we should not force one or the other. There are use cases where you do want it to be local (although you would ignore the kind, you would not convert to UTC), and not converted to UTC (we haven't used that in a long time, but sometimes you have to). That's the very reason why time provider exists in the first place - to not force you to go UTC/non-UTC in the first place.

We should have a UTC provider by default, and system one should not be used.

src/Domain/LeanCode.TimeProvider/Time.cs
jakubfijalkowski: @marcin-dardzinski The problem with init is that you would set it the the init, and then it won't be propagated to the request handlers, so we can't ignore that. :)

@Saancred it worked like that for the last couple of years without that much of a problem. The async local is only helpful in tests, so only the fake time provider needs that. 

Let's not introduce the async local at the global level, let's add an adapter that Marcin is talking about.

src/Domain/LeanCode.TimeProvider/Time.cs
Saancreed: > Datetime has two kinds, and we should not force one or the other. There are use cases where you do want it to be local (although you would ignore the kind, you would not convert to UTC), and not converted to UTC (we haven't used that in a long time, but sometimes you have to). That's the very reason why time provider exists in the first place - to not force you to go UTC/non-UTC in the first place.

The way it is now, it's still possible to retrieve local `DateTime` by calling `Time.NowWithOffset.DateTime`, I think UTC timestamps are still the primary use case for `DateTime` and so I'd prefer `Now` to always return in UTC. At least when you actually want local timestamp, it will be more obvious to see what you're doing and why.

In fact, we used to have a test that did

```csharp
[Fact]
public void SetTo_overload_for_DateTimeOffset_correctly_alters_return_value_of_Now()
{
    FixedTimeProvider.SetTo(expectedTime);
    Assert.Equal(expectedTime.UtcDateTime, LeanCode.TimeProvider.Time.Now);
}
```

so it really isn't new behavior.

> We should have a UTC provider by default, and system one should not be used.

:+1:

src/Domain/LeanCode.TimeProvider/Time.cs
Saancreed: I added a wrapper for that, let me know what you're thinking because I might have overengineered it a bit :see_no_evil:

src/Domain/LeanCode.TimeProvider/AsyncLocalTimeProvider.cs
```diff
@@ -0,0 +1,40 @@
+namespace LeanCode.TimeProvider;
+
+public sealed class AsyncLocalTimeProvider(System.TimeProvider fallback) : System.TimeProvider
+{
+    public static AsyncLocalTimeProvider SystemSharedInstance { get; } = new(System);
```
Saancreed: I suppose this one won't be of much use to us so I could get rid of it and keep only `UtcSystemSharedInstance`.

marcin-dardzinski: IMO, calling it `TestTimeProvider` (fixed no longer fits), having only one static instance, and making `Activate` method static would make better dx for our use case.

marcin-dardzinski (approved):

src/Domain/LeanCode.TimeProvider/AsyncLocalTimeProvider.cs
Saancreed: Right, that would also make it closer to previous `FixedTimeProvider.SetTo`, and having multiple shared static instances doesn't really work anyway; you could pick one to use in your project's tests but then you can never use the other because `Time` only has a single reference so they would evict each another.

On the other hand, being more explicit about which instance we want to use is imo a good thing, even if there's most commonly only one: `TestTimeProvider.Shared.Activate` isn't much longer than `TestTimeProvider.Activate` and doesn't favor our single instance that much… but maybe it should? :thinking:

I'm indifferent about the name, one explains the behavior and the other explains intent. Both work for me but let's see what @jakubfijalkowski says about this.

jakubfijalkowski: > but maybe it should?

IMO it should. "Shared" instance, where you really don't need shared but a singleton, hints that there might be a case for more instances. Here, _really_ static instance makes for a better interface IMO.

src/Domain/LeanCode.TimeProvider/Time.cs
jakubfijalkowski: > The way it is now, it's still possible to retrieve local DateTime by calling Time.NowWithOffset.DateTime, I think UTC timestamps are still the primary use case for DateTime and so I'd prefer Now to always return in UTC. At least when you actually want local timestamp, it will be more obvious to see what you're doing and why.

That's not really the case. The main idea is to allow changing the effective timezone _globally_, in a single place (e.g. setting env variable with the timezone), without the code knowing that it is working in a different timezone. Forcing you to use UTC _all the time_ defies the point of this abstraction a little. The main usecase is to have `Time.Now` and `Time.NowWithOffset` behave the same, i.e. if one returns zoned datetime, the other should also return _the same time_. And you should be able to switch between the behavior without changing the code.

 > In fact, we used to have a test that did

We did not. This code tests `FixedTimeProvider`, not general time provider. In tests that works, but that is the only place where the simplification makes sense.

jakubfijalkowski (approved):

jakubfijalkowski (approved):

PR #511, opened 2023-07-07
Postgres AD auth helper

marcin-dardzinski:


src/Infrastructure/LeanCode.Npgsql.ActiveDirectory/NpgsqlDataSourceBuilderExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace LeanCode.Npgsql.ActiveDirectory;
+
+public static class NpgsqlDataSourceBuilderExtensions
+{
+    public static NpgsqlDataSourceBuilder UseAzureActiveDirectoryAuthentication(
+        this NpgsqlDataSourceBuilder builder,
+        TokenCredential credential
+    )
+    {
+        return builder.UsePeriodicPasswordProvider(
```
jakubfijalkowski, line 15: Can we have a test if this works? :P 

Yeah, yeah, I know, but I think it is high time to have also Postgres-specific tests.

src/Infrastructure/LeanCode.Npgsql.ActiveDirectory/NpgsqlDataSourceBuilderExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+            async (_, ct) =>
+            {
+                var token = await credential.GetTokenAsync(
+                    new(new[] { "https://ossrdbms-aad.database.windows.net" }, null),
```
Saancreed: Locally some analyzer complained to me that this `new[] { "https://ossrdbms-aad.database.windows.net" }` should be a `private static readonly string[]` field for performance reasons.

Saancreed (approved):

src/Infrastructure/LeanCode.Npgsql.ActiveDirectory/NpgsqlDataSourceBuilderExtensions.cs
marcin-dardzinski: Yeah I saw it too, but TBH I don't have any guarantees regarding how this array is consumed. I _expect_ it not to be modified, but I'm not sure.

src/Infrastructure/LeanCode.Npgsql.ActiveDirectory/NpgsqlDataSourceBuilderExtensions.cs
marcin-dardzinski: What kind of test do you expect? A unit (say non-azure) one with some mocked `TokenCredential` or a E2E one? 
I can think of a test skipped by default, unless you set up credentials with some provisioning script to be run if necessary.

jakubfijalkowski: If you can develop a unit test, then unit test. If you can't - then integration test that is run on CI will be OK.

src/Infrastructure/LeanCode.Npgsql.ActiveDirectory/NpgsqlDataSourceBuilderExtensions.cs
jakubfijalkowski: I think we can safely assume that :)

src/Infrastructure/LeanCode.Npgsql.ActiveDirectory/NpgsqlDataSourceBuilderExtensions.cs
marcin-dardzinski: I did a e2e test connecting to postgres but it requires a database. We can skip it just like other (send grid/fcm/smsapi) tests and run it only when we touch the integration, or I can set up all the resources in the cloud and run it every time.

jakubfijalkowski: We should have that on CI (separate workflow) and probably have two targets in the local flow.

marcin-dardzinski:
Added cloud tests in github actions

.github/workflows/core_build_publish.yml
```diff
@@ -133,6 +133,11 @@ jobs:
       - name: Test
         run: dotnet msbuild /t:RunTestsWithCoverage
         working-directory: test
+        env:
```
jakubfijalkowski, line 139: Can we run the cloud/azure tests as a separate job? Will be easier to diagnose.

jakubfijalkowski (approved):

PR #516, opened 2023-07-12
Register one command under multiple paths

marcin-dardzinski:


src/CQRS/LeanCode.CQRS.Annotations/PathAliasAttribute.cs
```diff
@@ -0,0 +1,27 @@
+﻿namespace LeanCode.CQRS.Annotations;
+
+/// <summary>
+/// Register command/query/operation under additional HTTP path. This could be useful for reorganizing contracts namespaces without introducing a braking change.
```
jakubfijalkowski:
```suggestion
/// Register command/query/operation under additional HTTP path. This could be useful for reorganizing contracts namespaces without introducing a breaking change.
```
I love the driving reference.

jakubfijalkowski (approved):

PR #520, opened 2023-07-13
Make case-specific JsonStringEnumConverters generic

Saancreed:


marcin-dardzinski (approved):

src/Infrastructure/LeanCode.Serialization/JsonConverters/JsonStringEnumConverters.cs
```diff
@@ -1,34 +1,85 @@
+using System.Diagnostics.CodeAnalysis;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 
 namespace LeanCode.Serialization;
 
+[RequiresDynamicCode(
```
jakubfijalkowski: Since we have the generic variant now, is there a reason for keeping these (for our usecase)?

test/Infrastructure/LeanCode.Serialization.Tests/LeanCode.Serialization.Tests.csproj
```diff
@@ -1,5 +1,9 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
+  <PropertyGroup>
+    <NoWarn>SYSLIB1040</NoWarn>
```
jakubfijalkowski: Reason?

test/Infrastructure/LeanCode.Serialization.Tests/JsonStringEnumConvertersTests.cs
```diff
@@ -96,10 +96,120 @@ public void Snake_case_upper_converter_correctly_converts_enum_name(bool useSour
...
+    }
+
+    [JsonConverter(typeof(JsonKebabCaseLowerStringEnumConverter<KebabCaseLowerGeneric>))]
+    public enum KebabCaseLowerGeneric
+    {
+        ExampleValue
+    }
+
+    [Theory]
+    [InlineData(true)]
```
jakubfijalkowski: This is, IMO, the wrong usage for theories - these should be two separate tests as these test two separate codepaths.

src/Infrastructure/LeanCode.Serialization/JsonConverters/JsonStringEnumConverters.cs
Saancreed: Probably not, or at least I can't see one. I suppose I'll just remove them.

(Funny how they survived a total of 2 weeks :upside_down_face:)

test/Infrastructure/LeanCode.Serialization.Tests/LeanCode.Serialization.Tests.csproj
Saancreed: It's because I'm using non-generic converters on some enums here, I'll be able to drop this once I drop those converters and related tests.

test/Infrastructure/LeanCode.Serialization.Tests/JsonStringEnumConvertersTests.cs
Saancreed: This will become (imo, unnecessarily) verbose but ye, you're right :disappointed:

src/Infrastructure/LeanCode.Serialization/JsonConverters/JsonStringEnumConverters.cs
jakubfijalkowski: Yeah, since you're a fan of removing code, let's remove the code ;P

test/Infrastructure/LeanCode.Serialization.Tests/LeanCode.Serialization.Tests.csproj
jakubfijalkowski: One last question - why the non-generic variant generates this? [MSDN](https://learn.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib1040-1049) has such a sparse description of the error that I can't really find out why it is raised in the first place. :(

Saancreed: Docs seem to be outdated, the message I'm seeing is:

> CSC : error SYSLIB1040: The member 'LeanCode.Kratos.Model.IdentityState' has been annotated with 'JsonStringEnumConverter' which is not supported in native AOT. Consider using the generic 'JsonStringEnumConverter<TEnum>' instead.

jakubfijalkowski: Thx!

jakubfijalkowski (approved):

PR #526, opened 2023-07-18
Fix Dependabot's config

jakubfijalkowski:
It's `exclude-patterns`, not `exclude_patterns`...


marcin-dardzinski (approved):
:see_no_evil:

PR #535, opened 2023-07-24
Use UTC timestamps where appropriate

Saancreed:
While this is a bit more than we discussed, there could be similar issues whenever we want to persist any `DateTime` in PostgreSQL (and MSSQL change is just for consistency, we can drop that one). If you believe that we shouldn't do some of the changes proposed here, let me know, preferably with an explanation (and I'll try to explain why not doing them could be problematic).

Also, maybe we should introduce `UtcNow` property in `Time` that is guaranteed to return UTC timestamps?


marcin-dardzinski (approved):
Not sure if we already discussed it, but maybe `Time.UtcNow => Time.NowWithOffset.UtcDateTime`?

tomaszlaskowskiLC (approved):
LGTM, although I agree that we might have it under `Time.UtcNow`.

PR #549, opened 2023-08-23
Introduce CollectionsExtensions

lukaszgarstecki:


jakubfijalkowski:
I'm not sure that a new project is needed here. For me, this could be placed in `LeanCode.DomainModels.EF` and it would be equally discoverable/usable.

Also, we need tests.

src/Infrastructure/LeanCode.CollectionExtensions/CollectionExtensions.cs
```diff
@@ -0,0 +1,122 @@
...
+            )
+            .SelectMany(g => g.Values.DefaultIfEmpty(), (g, r) => resultSelector(g.Key, r));
+    }
+
+    private struct Grouping<TKey, TValue>
+    {
+        public TKey Key { get; set; }
+        public IEnumerable<TValue> Values { get; set; }
+
+        public static System.Reflection.PropertyInfo KeyPropertyInfo = typeof(Grouping<TKey, TValue>).GetProperty(
```
jakubfijalkowski:
```suggestion
        public static readonly System.Reflection.PropertyInfo KeyPropertyInfo = typeof(Grouping<TKey, TValue>).GetProperty(
```

src/Infrastructure/LeanCode.CollectionExtensions/CollectionExtensions.cs
```diff
@@ -0,0 +1,122 @@
...
+        return left.GroupJoin(
+                right,
+                leftKeySelector,
+                rightKeySelector,
+                (l, r) => new Grouping<TLeft, TRight> { Key = l, Values = r, }
+            )
+            .SelectMany(g => g.Values.DefaultIfEmpty(), lambda);
+    }
+
+    public static IEnumerable<TResult> LeftJoin<TLeft, TRight, TKey, TResult>(
```
jakubfijalkowski: I'm not sure we should have this - you might use it by accident and I haven't had to use `LeftJoin` on `IEnumerable` ever (although I might be biased).

lukaszgarstecki: True, we didn't use it in the project I found it in. I will remove it

Saancreed: :slightly_frowning_face:

lukaszgarstecki: Could you express what you have in mind @Saancreed ?

jakubfijalkowski: Okay, there _are_ cases where this might come in handy, so let's keep it.

test/Domain/LeanCode.DomainModels.EF.Tests/CollectionExtensionsTests.cs
```diff
@@ -0,0 +1,137 @@
...
+                    new { Animal = Cat, Species = CatS, },
+                    new { Animal = Dog, Species = DogS1, },
+                    new { Animal = Dog, Species = DogS2, },
+                    new { Animal = Dash, Species = null as Species, },
+                }
+            );
+    }
+}
+
+internal class Animal(string name, int legsCount)
```
marcin-dardzinski: Apply those hints, they'll accumulate and spam later :disappointed:

marcin-dardzinski (approved):

PR #556, opened 2023-09-07
Create migration without connection string

michal12334:


src/Infrastructure/LeanCode.EFMigrator/BaseFactory.cs
```diff
@@ -21,22 +21,27 @@ public abstract class BaseFactory<TContext, TFactory> : IDesignTimeDbContextFact
 
     public TContext CreateDbContext(string[] args)
     {
-        var connectionString =
-            MigrationsConfig.GetConnectionString()
-            ?? throw new InvalidOperationException("Failed to find connection string.");
+        var connectionString = MigrationsConfig.GetConnectionString();
+        var optionsAction = (SqlServerDbContextOptionsBuilder cfg) =>
```
jakubfijalkowski: I wonder if inlining it won't be slightly more readable.

jakubfijalkowski (approved):

PR #557, opened 2023-09-07
Introduce TimestampTz value object for use with PostgreSQL

Saancreed:
`TimestampTzLocalTimestampExpressionInterceptor` is there for ease of use, allowing access to `LocalTimestamp` property by translating it to `"UtcTimestamp" AT TIME ZONE "TimeZoneId"` in queries. Of course, we don't have access to the offset this way so care should be taken to use `GetLocalTimestampWithOffset()` when offset is actually needed (that will be evaulated client-side).

Creating indexes on columns corresponding to members of `OwnsOne`'d types is still a bit of a pain but should be doable with raw SQL in migrations.

I believe this + something like `Microsoft.CodeAnalysis.BannedApiAnalyzers` configured to forbid use of implicit cast of `DateTime` to `DateTimeOffset` (when mapping to DTOs) should be enough to cover most of our use cases.

### DIFF
--- a/src/Domain/LeanCode.DomainModels.EF/TimestampTzExpressionInterceptor.cs
+++ b/src/Domain/LeanCode.DomainModels.EF/TimestampTzExpressionInterceptor.cs
@@ -11,23 +11,23 @@ public sealed class TimestampTzExpressionRewriter : ExpressionVisitor
     private static readonly Type TimeZoneInfoType = typeof(TimeZoneInfo);
     private static readonly Type DateTimeOffsetType = typeof(DateTimeOffset);
 
-    private static readonly PropertyInfo LocalTimestampWithoutOffsetProperty =
+    public static readonly PropertyInfo LocalTimestampWithoutOffsetProperty =
         TimestampTzType.GetProperty(nameof(TimestampTz.LocalTimestampWithoutOffset))
         ?? throw new MissingMemberException(TimestampTzType.FullName, nameof(TimestampTz.LocalTimestampWithoutOffset));
 
-    private static readonly PropertyInfo UtcTimestampProperty =
+    public static readonly PropertyInfo UtcTimestampProperty =
         TimestampTzType.GetProperty(nameof(TimestampTz.UtcTimestamp))
         ?? throw new MissingMemberException(TimestampTzType.FullName, nameof(TimestampTz.UtcTimestamp));
 
-    private static readonly PropertyInfo TimeZoneIdProperty =
+    public static readonly PropertyInfo TimeZoneIdProperty =
         TimestampTzType.GetProperty(nameof(TimestampTz.TimeZoneId))
         ?? throw new MissingMemberException(TimestampTzType.FullName, nameof(TimestampTz.TimeZoneId));
 
-    private static readonly PropertyInfo UtcDateTimeProperty =
+    public static readonly PropertyInfo UtcDateTimeProperty =
         DateTimeOffsetType.GetProperty(nameof(DateTimeOffset.UtcDateTime))
         ?? throw new MissingMemberException(DateTimeOffsetType.FullName, nameof(DateTimeOffset.UtcDateTime));
 
-    private static readonly MethodInfo ConvertDateTimeBySystemTimeZoneIdMethod =
+    public static readonly MethodInfo ConvertDateTimeBySystemTimeZoneIdMethod =
         TimeZoneInfoType.GetMethod(
             nameof(TimeZoneInfo.ConvertTimeBySystemTimeZoneId),
             new[] { typeof(DateTime), typeof(string) }
--- a/src/Domain/LeanCode.DomainModels.EF/TimestampTzExpressionInterceptor.cs
+++ b/src/Domain/LeanCode.DomainModels.EF/TimestampTzExpressionInterceptor.cs
@@ -1,0 +1,69 @@
+using System.Linq.Expressions;
+using System.Reflection;
+using LeanCode.DomainModels.Model;
+using Microsoft.EntityFrameworkCore;
+
+namespace LeanCode.DomainModels.EF;
+
+public sealed class TimestampTzExpressionRewriter : ExpressionVisitor
+{
+    private static readonly Type TimestampTzType = typeof(TimestampTz);
+    private static readonly Type TimeZoneInfoType = typeof(TimeZoneInfo);
+    private static readonly Type DateTimeOffsetType = typeof(DateTimeOffset);
+
+    private static readonly PropertyInfo LocalTimestampWithoutOffsetProperty =
+        TimestampTzType.GetProperty(nameof(TimestampTz.LocalTimestampWithoutOffset))
+        ?? throw new MissingMemberException(TimestampTzType.FullName, nameof(TimestampTz.LocalTimestampWithoutOffset));
+
+    private static readonly PropertyInfo UtcTimestampProperty =
+        TimestampTzType.GetProperty(nameof(TimestampTz.UtcTimestamp))
+        ?? throw new MissingMemberException(TimestampTzType.FullName, nameof(TimestampTz.UtcTimestamp));
+
+    private static readonly PropertyInfo TimeZoneIdProperty =
+        TimestampTzType.GetProperty(nameof(TimestampTz.TimeZoneId))
+        ?? throw new MissingMemberException(TimestampTzType.FullName, nameof(TimestampTz.TimeZoneId));
+
+    private static readonly PropertyInfo UtcDateTimeProperty =
+        DateTimeOffsetType.GetProperty(nameof(DateTimeOffset.UtcDateTime))
+        ?? throw new MissingMemberException(DateTimeOffsetType.FullName, nameof(DateTimeOffset.UtcDateTime));
+
+    private static readonly MethodInfo ConvertDateTimeBySystemTimeZoneIdMethod =
+        TimeZoneInfoType.GetMethod(
+            nameof(TimeZoneInfo.ConvertTimeBySystemTimeZoneId),
+            new[] { typeof(DateTime), typeof(string) }
+        )
+        ?? throw new MissingMemberException(
+            TimeZoneInfoType.FullName,
+            nameof(TimeZoneInfo.ConvertTimeBySystemTimeZoneId)
+        );
+
+    protected override Expression VisitMember(MemberExpression node)
+    {
+        if (node.Member == LocalTimestampWithoutOffsetProperty)
+        {
+            return Expression.Call(
+                null,
+                ConvertDateTimeBySystemTimeZoneIdMethod,
+                Expression.Property(Expression.Property(node.Expression, UtcTimestampProperty), UtcDateTimeProperty),
+                Expression.Property(node.Expression, TimeZoneIdProperty)
+            );
+        }
+        else
+        {
+            return node;
+        }
+    }
+}
+
+public static class TimestampTzExpressionInterceptorDbContextOptionsBuilderExtensions
+{
+    public static readonly QueryExpressionVisitorInterceptor Interceptor = new(new TimestampTzExpressionRewriter());
+
+    public static DbContextOptionsBuilder AddTimestampTzExpressionInterceptor(
+        this DbContextOptionsBuilder optionsBuilder
+    )
+    {
+        optionsBuilder.AddInterceptors(Interceptor);
+        return optionsBuilder;
+    }
+}
--- a/src/Domain/LeanCode.DomainModels/Model/TimestampTz.cs
+++ b/src/Domain/LeanCode.DomainModels/Model/TimestampTz.cs
@@ -1,0 +1,93 @@
+using System.Text.Json.Serialization;
+
+namespace LeanCode.DomainModels.Model;
+
+/// <remarks>
+/// This type is intended to be used with PostgreSQL which supports AT TIME ZONE operator
+/// with IANA time zone IDs but cannot store both timestamp and offset in a single column.
+/// </remarks>
+public sealed record class TimestampTz : ValueObject
+{
+    [JsonIgnore]
+    public TimeZoneInfo TimeZoneInfo => TimeZoneInfo.FindSystemTimeZoneById(TimeZoneId);
+
+    /// <summary>
+    /// Returns contained timestamp in local time as DateTime.
+    /// Can be evaluated database-side with PostgreSQL but the offset is lost during conversion.
+    /// </summary>
+    [JsonIgnore]
+    public DateTime LocalTimestampWithoutOffset =>
+        TimeZoneInfo.ConvertTimeBySystemTimeZoneId(UtcTimestamp, TimeZoneId).DateTime;
+
+    /// <summary>
+    /// Returns contained timestamp in local time as DateTimeOffset.
+    /// Cannot be evaluated database-side with PostgreSQL.
+    /// </summary>
+    [JsonIgnore]
+    public DateTimeOffset LocalTimestampWithOffset =>
+        TimeZoneInfo.ConvertTimeBySystemTimeZoneId(UtcTimestamp, TimeZoneId);
+
+    /// <summary>
+    /// Always stored with offset equal to zero.
+    /// </summary>
+    [JsonInclude]
+    public DateTimeOffset UtcTimestamp { get; private init; }
+
+    /// <summary>
+    /// Always expressed using IANA identifier.
+    /// </summary>
+    [JsonInclude]
+    public string TimeZoneId { get; private init; }
+
+    public static bool IsValidTimeZoneId(string timeZoneId)
+    {
+        if (TimeZoneInfo.TryConvertWindowsIdToIanaId(timeZoneId, out var ianaId))
+        {
+            timeZoneId = ianaId;
+        }
+
+        return TimeZoneInfo.TryFindSystemTimeZoneById(timeZoneId, out var timeZoneInfo) && timeZoneInfo.HasIanaId;
+    }
+
+    [JsonConstructor]
+    private TimestampTz()
+    {
+        TimeZoneId = default!;
+    }
+
+    public TimestampTz(DateTimeOffset timestamp, TimeZoneInfo timeZoneInfo)
+        : this(timestamp.UtcDateTime, timeZoneInfo.Id) { }
+
+    public TimestampTz(DateTimeOffset timestamp, string timeZoneId)
+        : this(timestamp.UtcDateTime, timeZoneId) { }
+
+    public TimestampTz(DateTime utcTimestamp, TimeZoneInfo timeZoneInfo)
+        : this(utcTimestamp, timeZoneInfo.Id) { }
+
+    public TimestampTz(DateTime utcTimestamp, string timeZoneId)
+    {
+        if (utcTimestamp.Kind != DateTimeKind.Utc)
+        {
+            throw new ArgumentException("Timestamp's Kind must be Utc.", nameof(utcTimestamp));
+        }
+
+        if (TimeZoneInfo.TryConvertWindowsIdToIanaId(timeZoneId, out var ianaId))
+        {
+            timeZoneId = ianaId;
+        }
+
+        if (!TimeZoneInfo.FindSystemTimeZoneById(timeZoneId).HasIanaId)
+        {
+            throw new ArgumentException("TimeZoneId must have associated IANA identifier.", nameof(timeZoneId));
+        }
+
+        UtcTimestamp = new(utcTimestamp, TimeSpan.Zero);
+        TimeZoneId = timeZoneId;
+    }
+
+    public void Deconstruct(out DateTimeOffset utcTimestamp, out string timeZoneId)
+    {
+        utcTimestamp = UtcTimestamp;
+        timeZoneId = TimeZoneId;
+    }
+}
--- a/test/Domain/LeanCode.DomainModels.Tests/Model/TimestampTzTests.cs
+++ b/test/Domain/LeanCode.DomainModels.Tests/Model/TimestampTzTests.cs
@@ -1,0 +1,179 @@
+using System.Text.Json;
+using FluentAssertions;
+using LeanCode.DomainModels.Model;
+using Xunit;
+
+namespace LeanCode.DomainModels.Tests.Model;
+
+public class TimestampTzTests
+{
+    [Theory]
+    [InlineData("UTC")]
+    [InlineData("Etc/UTC")]
+    [InlineData("Etc/GMT")]
+    [InlineData("Europe/Warsaw")]
+    [InlineData("Central European Standard Time")]
+    public void Both_IANA_and_Windows_time_zone_ids_are_valid(string timeZoneId)
+    {
+        AssertionExtensions.Should(TimestampTz.IsValidTimeZoneId(timeZoneId)).BeTrue();
+    }
+
+    [Theory]
+    [InlineData("")]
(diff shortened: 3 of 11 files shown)


src/Domain/LeanCode.DomainModels.EF/TimestampTzLocalTimestampExpressionInterceptor.cs
```diff
@@ -0,0 +1,56 @@
+using System.Linq.Expressions;
+using System.Reflection;
+using LeanCode.DomainModels.Model;
+using Microsoft.EntityFrameworkCore.Diagnostics;
+
+namespace LeanCode.DomainModels.EF;
+
+public sealed class TimestampTzLocalTimestampExpressionInterceptor : IQueryExpressionInterceptor
```
marcin-dardzinski: An extension setup helper would be nice.

src/Domain/LeanCode.DomainModels/Model/TimestampTz.cs
```diff
@@ -0,0 +1,16 @@
+namespace LeanCode.DomainModels.Model;
+
+public sealed record class TimestampTz(DateTime UtcTimestamp, string TimeZoneId) : ValueObject
```
marcin-dardzinski: I think some helper methods `FromUtc` / `FromLocal` would be nice (with `string` and `TimeZoneInfo` overloads). 
Also possibly time zone should be validated (unless it's a design choice).

Speaking of it, wouldn't it be worth to have a `TimeZoneInfo` property instead of class - it shouldn't complicate much EF and STJ _I hope_

src/Domain/LeanCode.DomainModels.EF/TimestampTzLocalTimestampExpressionInterceptor.cs
```diff
@@ -0,0 +1,56 @@
...
+            )
+            ?? throw new MissingMemberException(
+                TimeZoneInfoType.FullName,
+                nameof(TimeZoneInfo.ConvertTimeBySystemTimeZoneId)
+            );
+
+        protected override Expression VisitMember(MemberExpression node)
+        {
+            return node.Member == LocalTimestampProperty
+                ? Expression.Call(
```
marcin-dardzinski: This works for both SQL and postgres?

src/Domain/LeanCode.DomainModels/Model/TimestampTz.cs
Saancreed: > `FromUtc`

Would it be just the same thing as `new(timestamp, "Etc/UTC")`? If so then sure, we can add one.

> `FromLocal`

I see no value in this, shouldn't be too common.

> (with `string` and `TimeZoneInfo` overloads)

What if the `TimeZoneInfo` object we were passed has `HasIanaId == false`? Throw an exception?

> Also possibly time zone should be validated (unless it's a design choice).

This is tricky. How do you verify that passed `TimeZoneId` is a valid IANA ID using only .NET APIs? (and without calling into libicu)

> Speaking of it, wouldn't it be worth to have a `TimeZoneInfo` property instead of class - it shouldn't complicate much EF and STJ _I hope_

I think it would complicate EF, because afaik it's not possible to store `TimeZoneInfo` directly in a column without writing a converter and with one, rewriting expressions could become much harder than it is right now. Also, how is STJ relevant here?

src/Domain/LeanCode.DomainModels.EF/TimestampTzLocalTimestampExpressionInterceptor.cs
Saancreed: Not sure about SQL Server, it's not the target of this PR but if we wanted to explicitly support it, we could provide a translation for `GetLocalTimestampWithOffset` using [EF.Functions.AtTimeZone](https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.sqlserverdbfunctionsextensions.attimezone?view=efcore-7.0#microsoft-entityframeworkcore-sqlserverdbfunctionsextensions-attimezone(microsoft-entityframeworkcore-dbfunctions-system-datetime-system-string)).

src/Domain/LeanCode.DomainModels/Model/TimestampTz.cs
marcin-dardzinski: > > `FromLocal`
> 
> I see no value in this, shouldn't be too common.

I mean `FromLocal(DateTime/DateTimeOffset, TimeZoneInfo)`, not the local machine time- this seems to be a common use case - do you expect clients to always send utc + tz pair? 

>> (with string and TimeZoneInfo overloads)
>
> What if the TimeZoneInfo object we were passed has HasIanaId == false? Throw an exception?

What if I pass `XD` string as timezone id? How would postgres `AT TIME ZONE` react to that? I mean, we'd need to validate this at some point.

>>  Also possibly time zone should be validated (unless it's a design choice).
>
> This is tricky. How do you verify that passed TimeZoneId is a valid IANA ID using only .NET APIs? (and without calling into libicu)

What is wrong with .NET APIs? Is it not up to date / enough for our use cases? I think usually it's the backend that dictates available time zones anyway.

> > Speaking of it, wouldn't it be worth to have a `TimeZoneInfo` property instead of class - it shouldn't complicate much EF and STJ _I hope_
> 
> I think it would complicate EF, because afaik it's not possible to store `TimeZoneInfo` directly in a column without writing a converter and with one, rewriting expressions could become much harder than it is right now. Also, how is STJ relevant here?

Okay, yeah this would call for custom translation. Speaking of `STJ` - this VO may still be serialized in events, json columns, etc.

src/Domain/LeanCode.DomainModels.EF/TimestampTzLocalTimestampExpressionInterceptor.cs
jakubfijalkowski: If we put it in non-Postgres specific package, we should support SQL Server (someone will use it there, and be very surprised that it does not work).

src/Domain/LeanCode.DomainModels/Model/TimestampTz.cs
Saancreed: > I mean `FromLocal(DateTime/DateTimeOffset, TimeZoneInfo)`, not the local machine time- this seems to be a common use case - do you expect clients to always send utc + tz pair?

No, I'd expect the client to always send timezone ID and either some kind of `DateTimeOffset` (because contracts forbid `DateTime`), in which case using the constructor is about as complicated as calling `.UtcDateTime` on that `DateTimeOffset`. If you believe that deserves a helper method, sure, I can add one.

> What if I pass `XD` string as timezone id? How would postgres `AT TIME ZONE` react to that? I mean, we'd need to validate this at some point.

We could, but it will never be foolproof. Let me know when you figure out how to extract IANA ID (and not just any ID that could be a Windows ID) from any arbitrary `TimeZoneInfo` object and maybe we will give this a try.

> What is wrong with .NET APIs? Is it not up to date / enough for our use cases? I think usually it's the backend that dictates available time zones anyway.

I recall there being some project that attempted to validate timezone IDs only to fail horribly when passed non-canonical/linked IDs or Windows IDs.

> Okay, yeah this would call for custom translation? Speaking of `STJ` - this VO may still be serialized in events, json columns, etc.

I'm just not sure if EF won't scream at us for attempting to construct an method expression that expects a parameter of type `string` but is given a parameter of type `TimeZoneInfo` and backend isn't aware that `TimeZoneInfo` is actually a lie.

About STJ, good point, less about JSON columns but certainly about events. At worst we will add a new converter to `LeanCode.Serialization` but I'll try to check first it it would be needed at all.

src/Domain/LeanCode.DomainModels.EF/TimestampTzLocalTimestampExpressionInterceptor.cs
Saancreed: That should be doable, but… can we somehow recognize which provider is used from inside our interceptor? If not, we would probably have to split them into two versions, one for SQL Server and one for Postgres :weary:

jakubfijalkowski: I'm pretty sure it is possible, but probably not officially supported ^^'

I think two separate extension methods would be OK.

src/Domain/LeanCode.DomainModels/Model/TimestampTz.cs
marcin-dardzinski: OK, having the rest resolved let's focus on just `TimeZoneInfo` validation applicability.

I believe accepting plain string without any validation is just escaping from the problem. _At some point_ you'll need to convert something, either client/backend/db and you'll encounter undefined behaviour. 

I agree that is might be easy to shoot oneself in the feet, probably we'd need to extract validation to some static class that can be overridden in the project if necessary. 

 > We could, but it will never be foolproof. Let me know when you figure out how to extract IANA ID (and not just any ID that could be a Windows ID) from any arbitrary TimeZoneInfo object and maybe we will give this a try.

Isn't `TimeZoneInfo.TryConvertWindowsIdToIanaId` + `TimeZoneInfo.HasIanaId` enough / enough to filter out all the zones that we don't need? Also, in some of the projects I took an approach where the backend returns supported timezones to clients (clients send local time and tz id and can even be unaware of the offsets write time).

I'm leaning toward some `ITimeZoneProvider` interface with sensible defaults, that will list all supported timezones + let you validate id.

Saancreed: I implemented validation of time zone IDs but I'd say I'm against making it configurable. I don't think there are any reasonable alternatives other than what .NET already provides via libicu.

Let's hope nobody will try to run this code in minimal Alpine container.

marcin-dardzinski: I mean, this could be an escape hatch if you really need to override this but I can do without.

src/Domain/LeanCode.DomainModels/Model/TimestampTz.cs
```diff
@@ -0,0 +1,47 @@
...
+        TimeZoneInfo.ConvertTimeBySystemTimeZoneId(new DateTimeOffset(UtcTimestamp, TimeSpan.Zero), TimeZoneId);
+
+    public DateTime LocalTimestamp => TimeZoneInfo.ConvertTimeBySystemTimeZoneId(UtcTimestamp, TimeZoneId);
+
+    public DateTime UtcTimestamp { get; init; } =
+        UtcTimestamp.Kind == DateTimeKind.Utc
+            ? UtcTimestamp
+            : throw new ArgumentException("UtcTimestamp.Kind must be Utc.", nameof(UtcTimestamp));
+
+    public string TimeZoneId { get; init; } = NormalizeAndValidateTimeZoneId(TimeZoneId);
```
marcin-dardzinski: At this moment I think a classic constructor would be more readable.

Saancreed: But then I'd have to implement deconstructors manually :disappointed:

jakubfijalkowski: I sometimes really don't know if you are sarcastic or just dead serious xD

jakubfijalkowski: Assuming you are dead serious - I 100% agree with Dardzin. Now, there is logic sprinkled here and there, when assigning to properties, and every time I look at the file I find out new rules.

jakubfijalkowski (commented):
I am totally OK with no test in the draft phase, but what I would expect from this solution if we agree that it is OK is to have:
1) unit test for `VisitMember` (should be doable with just lambdas + Compile + call),
2) unit tests for the logic in `TimestampTz`,
3) integration tests for both SQL Server and PostgreSQL.

src/Domain/LeanCode.DomainModels/Model/TimestampTz.cs
Saancreed: Nah, you guys have a point. I'll rewrite this into something where the flow should be easier to follow.

Saancreed: All the boring stuff is now being done explicitly inside the constructor.

src/Domain/LeanCode.DomainModels.EF/TimestampTzLocalTimestampExpressionInterceptor.cs
Saancreed: @marcin-dardzinski We have decided that there will be no support for SQL Server at this time (and the type has a comment explaining that it's meant to be used only with PostgreSQL).

src/Domain/LeanCode.DomainModels/Model/TimestampTz.cs
```diff
@@ -0,0 +1,93 @@
...
+
+namespace LeanCode.DomainModels.Model;
+
+/// <remarks>
+/// This type is intended to be used with PostgreSQL which supports AT TIME ZONE operator
+/// with IANA time zone IDs but cannot store both timestamp and offset in a single column.
+/// </remarks>
+public sealed record class TimestampTz : ValueObject
+{
+    public static bool IsValidTimeZoneId(string timeZoneId)
```
jakubfijalkowski: Nit: methods (no matter if static or not) should be placed after properties.

test/Domain/LeanCode.DomainModels.Tests/Model/TimestampTzTests.cs
```diff
@@ -0,0 +1,171 @@
+using System.Text.Json;
+using LeanCode.DomainModels.Model;
+using Xunit;
+
+namespace LeanCode.DomainModels.Tests.Model;
+
+public class TimestampTzTests
```
jakubfijalkowski, line 8: FluentAssertions please. From now on, we should use it instead of Xunit ones.

test/Domain/LeanCode.DomainModels.Tests/Model/TimestampTzTests.cs
```diff
@@ -0,0 +1,171 @@
...
+    {
+        var source = new TimestampTz(DateTime.UtcNow, "Europe/Warsaw");
+        var (timestamp, timeZoneId) = source;
+        var reconstructed = new TimestampTz(timestamp, timeZoneId);
+
+        Assert.Equal(source, reconstructed);
+    }
+
+    [Fact]
+    public void Json_serialization_roundtrips()
```
jakubfijalkowski: Let's split this test into two - one that check if the serialization format is indeed like we expect, and the other that checks the roundtripping.

test/LeanCode.IntegrationTests/TimestampTzTests.cs
```diff
@@ -0,0 +1,67 @@
...
+    {
+        app = new TestApp();
+    }
+
+    [PostgresFact]
+    public async Task TimestampTz_local_timestamp_access_is_correctly_translated()
+    {
+        using var scope = app.Services.CreateScope();
+        var dbContext = scope.ServiceProvider.GetRequiredService<TestDbContext>();
+        var today = DateOnly.FromDateTime(DateTime.UtcNow);
```
jakubfijalkowski: I'm not sure we should rely on "current date" in tests - although we don't really depend on the exact date (hopefully), making it dynamic might make changes more tricky.

test/LeanCode.IntegrationTests/TimestampTzTests.cs
```diff
@@ -0,0 +1,67 @@
...
+{
+    private readonly TestApp app;
+
+    public TimestampTzTests()
+    {
+        app = new TestApp();
+    }
+
+    [PostgresFact]
+    public async Task TimestampTz_local_timestamp_access_is_correctly_translated()
```
jakubfijalkowski: Let's refactor this one, as this test isn't that self-explanatory.

From what I understand, it checks:
1) That the generated SQL has correct `ORDER BY` statement,
2) That the behavior of the query is correct, in two versions.

We should split that into two tests. Performance, for now (and I presume - for later also), is not really a problem and we can favor readability instead.

Also, let's remove from sight what doesn't really matter, mainly the "arrange" part that adds data to the database. And let's name the assertions, esp. the one that check SQL statement.

Also, Fluent :)

test/LeanCode.IntegrationTests/docker/docker-compose.yml
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env -S docker compose -f
```
jakubfijalkowski, line 1: Works, but isn't really how the docker-compose should be used. Let's remove that.

test/LeanCode.IntegrationTests/TimestampTzTests.cs
```diff
@@ -0,0 +1,67 @@
...
+        dbContext.Meetings.Add(meeting1);
+        dbContext.Meetings.Add(meeting2);
+
+        await dbContext.SaveChangesAsync();
+
+        dbContext.ChangeTracker.QueryTrackingBehavior = QueryTrackingBehavior.TrackAll;
+
+        var orderedByLocalQuery = dbContext.Meetings.OrderBy(m => m.StartTime.LocalTimestampWithoutOffset);
+
+        Assert.Contains(
```
marcin-dardzinski: I wonder if we can get inspired by other providers and test it in a better fashion. Technically, you don't even need a running postgres to check if the sql is generated correctly,

test/LeanCode.IntegrationTests/TimestampTzTests.cs
Saancreed: It indeed shouldn't matter (because on any given day the difference between Los Angeles and Tokyo local times should be greater than a few hours). But I can hardcode any arbitrary date here if you believe it will be an improvement.

test/LeanCode.IntegrationTests/docker/docker-compose.yml
Saancreed: `docker compose` is not supposed to be used via `docker compose -f path/to/docker-compose.yml`? Why not? :upside_down_face:

Jokes aside, this is _very_ conveniant for me to use because those files live in a separate directory. I'd prefer to keep this just because you're not at all required to use it (and we already have useless shebangs in other places where they make even less sense to be).

test/LeanCode.IntegrationTests/TimestampTzTests.cs
Saancreed: You are right, just like we don't need running database just to check if all migrations were generated, and yet we do run such checks as integration tests because it's simpler.

Should we prepare unit tests to reference `Npgsql` and build `DbContext` just to execute a single assert?

test/LeanCode.IntegrationTests/TimestampTzTests.cs
Saancreed: > We should split that into two tests. Performance, for now (and I presume - for later also), is not really a problem and we can favor readability instead.

:+1:

> Also, let's remove from sight what doesn't really matter, mainly the "arrange" part that adds data to the database.

I can extract this to a separate method. Cannot really hide it though, it still needs to happen here.

> And let's name the assertions, esp. the one that check SQL statement.

How does one name method calls? Should I add labels, known from `goto`? Or would extracting them to their own methods be acceptable?

> Also, Fluent :)

:(

jakubfijalkowski: > How does one name method calls? Should I add labels, known from goto? Or would extracting them to their own methods be acceptable?

Paraphrasing a classic, "Too much sarcasm will kill you" ;P

> I can extract this to a separate method. Cannot really hide it though, it still needs to happen here.

You will reduce it from tens (~20?) of lines to one by extracting it to a method, it will basically be hidden.

test/LeanCode.IntegrationTests/TimestampTzTests.cs
jakubfijalkowski: Most of the people are used to hardcoded data in tests (for the "normal" unit test reasons) and I think it would make understanding what is going on easier in that case - that's the only benefit I see. So basically your call which way we go.

Saancreed: I'll just hardcode it.

test/LeanCode.IntegrationTests/docker/docker-compose.yml
jakubfijalkowski: Hmm... That's a reasonable argument.

Food for thought - maybe it is worth to re-think how we approach dev env for testing here? This docker-compose approach was coined a loong time ago, when I haven't really had any major opinion about that. Maybe "just" moving it somewhere else, or keeping it as a long-lived env would be better? Or maybe using something like TestContainers?

Saancreed: I wouldn't mind alternatives but I'm okay with how they work currently. Compose is not too bad.

src/Domain/LeanCode.DomainModels/Model/TimestampTz.cs
Saancreed: Pls gib linter/analyzer for this :weary:

test/LeanCode.IntegrationTests/docker/docker-compose.yml
jakubfijalkowski: I'm not saying that we should ditch compose. I think that for this case there isn't really a better option (and for sure that shouldn't be full K8s cluster with TF & Tilt). But maybe there is a better way to structure things 
(e.g. use compose to run the databases, but run tests directly on the PC).

src/Domain/LeanCode.DomainModels.EF/TimestampTzExpressionInterceptor.cs
```diff
@@ -0,0 +1,69 @@
...
+
+namespace LeanCode.DomainModels.EF;
+
+public sealed class TimestampTzExpressionRewriter : ExpressionVisitor
+{
+    private static readonly Type TimestampTzType = typeof(TimestampTz);
+    private static readonly Type TimeZoneInfoType = typeof(TimeZoneInfo);
+    private static readonly Type DateTimeOffsetType = typeof(DateTimeOffset);
+
+    private static readonly PropertyInfo LocalTimestampWithoutOffsetProperty =
```
jakubfijalkowski: I would add a set of tests that check if these really access the properties that we meant, basically `ThePropertyInfo.Get(obj).Should().Be(obj.TheProperty)` for each and every property. Would prevent some nasty errors that refactorings might introduce.

test/LeanCode.IntegrationTests/TimestampTzTests.cs
marcin-dardzinski: Just thinking out loud - I just don't know how others do it.

jakubfijalkowski (approved):

src/Domain/LeanCode.DomainModels.EF/TimestampTzExpressionInterceptor.cs
Saancreed: Can do, even though we are indirectly verifying this in integration tests. I imagine I should make these `public` for testing?

jakubfijalkowski: Yeah, it is tested in integration tests, but I fear that if this is the problem, the integration test will result in a very cryptic failure.

Yup, I will be more than OK with making it public.

jakubfijalkowski (approved):

PR #558, opened 2023-09-11
Make migrator usable with postgres

michal12334:


Dragemil (commented):
Wouldn't here be value in providing a more specific base classes as well (inheriting from this base class) for our common cases - a SqlServer one, and a Postgres one?

src/Infrastructure/LeanCode.EFMigrator/BaseFactory.cs
```diff
@@ -15,31 +15,24 @@ public abstract class BaseFactory<TContext, TFactory> : IDesignTimeDbContextFact
         typeof(TFactory).Assembly.GetName().Name
         ?? throw new InvalidOperationException("This type is not supported on Assembly-less runtimes.");
 
-    protected virtual void UseAdditionalSqlServerDbContextOptions(SqlServerDbContextOptionsBuilder builder) { }
-
     protected virtual void UseAdditionalDbContextOptions(DbContextOptionsBuilder<TContext> builder) { }
```
marcin-dardzinski: Do we need this, can't you do the same in `UseDbProvider`?

marcin-dardzinski (approved):

jakubfijalkowski:
@Dragemil we would need separate assemblies for that, or reference both SqlServer and Postgres-specific packages in the migrator, no matter what db we use.

Saancreed:
A bit off-topic for this but have we considered using [migration bundles](https://learn.microsoft.com/en-us/ef/core/managing-schemas/migrations/applying?tabs=dotnet-core-cli#bundles)? Maybe it would allow us to reduce our `Main` to just applying the seed if needed, or even get rid of it altogether if we don't.

Saancreed (approved):

jakubfijalkowski:
I think we rejected it some time ago but I don't remember why we did that. I agree that we should look at it once again.

PR #570, opened 2023-09-26
Add initiator to identity trace attributes

lukaszgarstecki:
The reasoning behind the change is that we need to relay on the actual user (eg `System` in internal calls) but we want to preserve the first user who performed an action (called `initiator`).

### DIFF
--- a/src/Infrastructure/LeanCode.OpenTelemetry/IdentityTraceBaggageHelpers.cs
+++ b/src/Infrastructure/LeanCode.OpenTelemetry/IdentityTraceBaggageHelpers.cs
@@ -5,10 +5,10 @@ namespace LeanCode.OpenTelemetry;
 
 public static class IdentityTraceBaggageHelpers
 {
-    public const string UserIdKey = "enduser.id";
-    public const string UserRoleKey = "enduser.role";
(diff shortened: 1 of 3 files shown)


Dragemil (commented):
The implementations fine however I've realized we've got the span tags mixed up regarding to [OTel standard](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#general-identity-attributes).

src/Infrastructure/LeanCode.OpenTelemetry/IdentityTraceBaggageHelpers.cs
```diff
@@ -6,17 +6,19 @@ namespace LeanCode.OpenTelemetry;
 public static class IdentityTraceBaggageHelpers
 {
     public const string UserIdKey = "enduser.id";
-    public const string RoleKey = "enduser.role";
+    public const string UserRoleKey = "enduser.role";
```
jakubfijalkowski: Yeah, as Emil said, we probably should use `enduser` as "initiator" (but let's not rename them - let's keep `EndUser` name) and something like `current_user` for `User`.

jakubfijalkowski (approved):

PR #573, opened 2023-09-27
Make FCMClient public methods virtual

michal12334:


jakubfijalkowski (approved):

jakubfijalkowski (changes requested):
Let's make `virtual` only the methods that really do something, not the overloads - they are helpers and shouldn't behave differently.

jakubfijalkowski (approved):

PR #574, opened 2023-09-28
Make SendEmailAsync method of SendGridRazorClient class virtual

michal12334:


jakubfijalkowski (approved):

Saancreed:
This shouldn't be necessary, we can already override the one relevant method on `SendGridClient` itself (`RequestAsync`?).

Unless you specifically want to mock the rendering part, that is.

michal12334:
> This shouldn't be necessary, we can already override the one relevant method on `SendGridClient` itself (`RequestAsync`?).
> 
> Unless you specifically want to mock the rendering part, that is.

I want to mock whole process of sending email.

Saancreed (approved):

PR #575, opened 2023-09-29
Add helpers that allow loading tracked entites directly from context [COR-121]

jakubfijalkowski:
This adds two helper methods that try to streamline loading already-tracked entities in repositories. They basically try to replicate how [`FindAsync`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.dbset-1.findasync?view=efcore-7.0) works, but to also allow to use `Include`s.

It is very important that the `Include`s used in all calls stay the same (otherwise, you might get entity that looks different by mistake), that's why they are constrained to `EFRepository` - we already have this assumption there (i.e. we should load the whole aggregate, always).

Since this almost abuses EF internals, there is a lot of test for 4 lines of code. :)

Closes COR-121

### DIFF
--- a/src/Domain/LeanCode.DomainModels.EF/CachingEFRepository.cs
+++ b/src/Domain/LeanCode.DomainModels.EF/CachingEFRepository.cs
@@ -1,0 +1,67 @@
+using System.Diagnostics.CodeAnalysis;
+using LeanCode.DomainModels.Model;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Internal;
+
+namespace LeanCode.DomainModels.EF;
+
+public abstract class CachingEFRepository<TEntity, TIdentity, TContext> : EFRepository<TEntity, TIdentity, TContext>
+    where TEntity : class, IAggregateRootWithoutOptimisticConcurrency<TIdentity>
+    where TIdentity : notnull
+    where TContext : DbContext
+{
+    protected CachingEFRepository(TContext dbContext)
+        : base(dbContext) { }
+
+    /// <summary>
+    /// Finds an entity by primary key. If the entity with provided key is tracked by the underlying
+    /// <see cref="DbContext" />, this method will return the cached entity.
+    /// </summary>
+    /// <remarks>For implementers: the default implementation won't work for composite primary keys.</remarks>
+    /// <param name="id">The identifier that</param>
+    /// <param name="cancellationToken">A <see cref="CancellationToken" /> to observe while waiting for the task to complete.</param>
+    /// <returns>Task with the found entity, or -null- if not found.</returns>
+    public override async Task<TEntity?> FindAsync(TIdentity id, CancellationToken cancellationToken = default)
+    {
+        return await FindTrackedOrLoadNewAsync(
+            id,
+            _ => BaseQuery().AsTracking().FirstOrDefaultAsync(e => e.Id.Equals(id), cancellationToken)
+        );
+    }
+
+    [SuppressMessage(
+        "?",
+        "EF1001",
+        Justification = "This is basically the `EntityFinder.FindTracked` to mimic `FindAsync` snapshot re-use."
+    )]
+    protected TEntity? FindTracked(TIdentity id)
+    {
+        // Safety: aggregates are bound to have Id as a primary key by design.
+        var primaryKey = DbContext.Model.FindEntityType(typeof(TEntity))!.FindPrimaryKey()!;
+        return ((IDbContextDependencies)DbContext).StateManager!.TryGetEntryTyped(primaryKey, id)?.Entity as TEntity;
+    }
+
+    [SuppressMessage(
+        "?",
+        "EF1001",
+        Justification = "This is basically the `EntityFinder.FindTracked` to mimic `FindAsync` snapshot re-use."
+    )]
+    protected TEntity? FindTracked(params object[] id)
+    {
+        // Safety: aggregates are bound to have Id as a primary key by design.
+        var primaryKey = DbContext.Model.FindEntityType(typeof(TEntity))!.FindPrimaryKey()!;
+        return ((IDbContextDependencies)DbContext).StateManager!.TryGetEntryTyped(primaryKey, id)?.Entity as TEntity;
+    }
+
+    protected ValueTask<TEntity?> FindTrackedOrLoadNewAsync(TIdentity id, Func<DbSet<TEntity>, Task<TEntity?>> query)
+    {
+        var tracked = FindTracked(id);
+        return tracked is not null ? new(tracked) : new(query(DbSet));
+    }
+
+    protected ValueTask<TEntity?> FindTrackedOrLoadNewAsync(object[] id, Func<DbSet<TEntity>, Task<TEntity?>> query)
+    {
+        var tracked = FindTracked(id);
+        return tracked is not null ? new(tracked) : new(query(DbSet));
+    }
+}
--- a/src/Domain/LeanCode.DomainModels.EF/CachingEFRepository.cs
+++ b/src/Domain/LeanCode.DomainModels.EF/CachingEFRepository.cs
@@ -13,14 +13,11 @@ public abstract class CachingEFRepository<TEntity, TIdentity, TContext> : EFRepo
     protected CachingEFRepository(TContext dbContext)
         : base(dbContext) { }
 
+    /// <inheritdoc />
     /// <summary>
     /// Finds an entity by primary key. If the entity with provided key is tracked by the underlying
     /// <see cref="DbContext" />, this method will return the cached entity.
     /// </summary>
-    /// <remarks>For implementers: the default implementation won't work for composite primary keys.</remarks>
-    /// <param name="id">The identifier that</param>
-    /// <param name="cancellationToken">A <see cref="CancellationToken" /> to observe while waiting for the task to complete.</param>
-    /// <returns>Task with the found entity, or -null- if not found.</returns>
     public override async Task<TEntity?> FindAsync(TIdentity id, CancellationToken cancellationToken = default)
     {
         return await FindTrackedOrLoadNewAsync(
--- a/src/Domain/LeanCode.DomainModels.EF/EFRepository.cs
+++ b/src/Domain/LeanCode.DomainModels.EF/EFRepository.cs
@@ -1,7 +1,9 @@
+using System.Diagnostics.CodeAnalysis;
 using LeanCode.DomainModels.DataAccess;
 using LeanCode.DomainModels.Model;
 using LeanCode.TimeProvider;
 using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Internal;
 
 namespace LeanCode.DomainModels.EF;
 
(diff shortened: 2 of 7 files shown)


src/Domain/LeanCode.DomainModels.EF/EFRepository.cs
```diff
@@ -58,4 +60,28 @@ public virtual void Update(TEntity entity)
...
+
+    [SuppressMessage(
+        "?",
+        "EF1001",
+        Justification = "There basically resembles the EntityFinder.FindTracked to mimic `FindAsync` snapshot re-use."
+    )]
+    protected TEntity? FindTracked(TIdentity id)
+    {
+        // Safety: aggregates are bound to have Id as a primary key by design.
+        var primaryKey = DbContext.Model.FindEntityType(typeof(TEntity))!.FindPrimaryKey()!;
```
marcin-dardzinski: This wouldn't work for compound keys, would it?

src/Domain/LeanCode.DomainModels.EF/EFRepository.cs
```diff
@@ -58,4 +60,28 @@ public virtual void Update(TEntity entity)
...
+        return ((IDbContextDependencies)DbContext).StateManager!.TryGetEntryTyped(primaryKey, id)?.Entity as TEntity;
+    }
+
+    protected ValueTask<TEntity?> FindTrackedOrLoadNewAsync(
+        TIdentity id,
+        IQueryable<TEntity> withIncludes,
```
marcin-dardzinski: I find it a bit awkward that I can essentially put any queryable here, while in most cases I'm going to use plain `DbSet` (we tend to do owned types anyway plus there's a `Navigation().AutoInclude()` option).

Maybe 
```
    protected ValueTask<TEntity?> FindTrackedOrLoadNewAsync(
        TIdentity id,
        CancellationToken cancellationToken
        Func<IQueryable<TEntity>,IQueryable<TEntity>>? includes = null)
``` 
Placement of cancellation token is a bit awkward here, but in this context it's better than having two overloads, one with `includes` and one without.

src/Domain/LeanCode.DomainModels.EF/EFRepository.cs
jakubfijalkowski: As far as I know, there is still a single PK in the model, it just has multiple properties. But I'll add a test for that.

src/Domain/LeanCode.DomainModels.EF/EFRepository.cs
jakubfijalkowski: This "tend" with owned types is, IMO, exaggerated. It's not that common for sub-entities. Also, if you use owned types **and** the db context is tracking, then you don't need this mumbo jumbo - `FindAsync` will work (as this is basically FindAsync reimplemented).

We don't work on dbsets, unfortunately - when you do anything with it (include or AsTracking), you end up with IQueryable.

And, to be honest, having lambda instead of plain object is not that much different for me. I can change that (except the cancelation token, let's not deviate from the standard)

src/Domain/LeanCode.DomainModels.EF/EFRepository.cs
```diff
@@ -58,4 +60,28 @@ public virtual void Update(TEntity entity)
...
+    [SuppressMessage(
+        "?",
+        "EF1001",
+        Justification = "There basically resembles the EntityFinder.FindTracked to mimic `FindAsync` snapshot re-use."
+    )]
+    protected TEntity? FindTracked(TIdentity id)
+    {
+        // Safety: aggregates are bound to have Id as a primary key by design.
+        var primaryKey = DbContext.Model.FindEntityType(typeof(TEntity))!.FindPrimaryKey()!;
+        return ((IDbContextDependencies)DbContext).StateManager!.TryGetEntryTyped(primaryKey, id)?.Entity as TEntity;
```
Saancreed: `DbContext.GetService<IStateManager>()` could be more future-proof because it doesn't rely on the property in `IDbContextDependencies`. And it avoids an ugly cast.

src/Domain/LeanCode.DomainModels.EF/EFRepository.cs
Saancreed: Random thought, but considering

> It is very important that the `Include`s used in all calls stay the same (otherwise, you might get entity that looks different by mistake)

… maybe we could take this opportunity to rework this class a bit and instead of having the consumer/subclasses implement `FindAsync`, they could instead implement `protected abstract IQueryable<TEntity> QueryWithIncludes();` (or `PrepareQuery`) with the default implementation of `dbSet.AsQueryable()`?

Then we could use this new method to feed the same include chain to both `Find` methods' default implementations.

src/Domain/LeanCode.DomainModels.EF/EFRepository.cs
Saancreed: But `TIdentity id` can't represent multiple properties outside of some by-convention `(FooId FooId, BarId BarId)` which EFCore is unlikely to understand.

Saancreed (commented):
With new methods being `protected`, they won't be directly callable from application code by default (unless reimplemented and exposed as `public` in each subclass). Are they meant to be used only in the implementations of `FindAsync`? Is the decision to use them supposed to be made separately in each repository?

There is probably little value in repositories that expose entry points for both "find or fetch" and "always fetch" and let the caller decide which one to use but I could be missing something.

src/Domain/LeanCode.DomainModels.EF/EFRepository.cs
jakubfijalkowski: Yeah, forgot about the "magic" that maps tuple to composite keys and that EF interprets that as an array. :(

src/Domain/LeanCode.DomainModels.EF/EFRepository.cs
jakubfijalkowski: Will think about it.

test/Domain/LeanCode.DomainModels.EF.Tests/EFRepositoryTrackingTests.cs
```diff
@@ -0,0 +1,190 @@
...
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Storage;
+using Xunit;
+
+namespace LeanCode.DomainModels.EF.Tests;
+
+public class EFRepositoryTrackingTests
+{
+    private readonly InMemoryDatabaseRoot DbRoot = new();
+    private readonly Guid DbId = Guid.NewGuid();
```
Saancreed: > Naming rule violation: The first word must begin with a lower case character (IDE1006)

:stuck_out_tongue:

src/Domain/LeanCode.DomainModels.EF/EFRepository.cs
jakubfijalkowski: We trade readability (which would be bad no matter what way we take IMO) for potential "one lookup less" if the `StateManager` property is not implemented through the `GetRequiredService` call.

test/Domain/LeanCode.DomainModels.EF.Tests/EFRepositoryTrackingTests.cs
jakubfijalkowski: Eh, this is much less visible in Rider than I'm used to.

src/Domain/LeanCode.DomainModels.EF/EFRepository.cs
jakubfijalkowski: Did that, plus changed how the method looks like to better support composite keys (thus - arrays).

src/Domain/LeanCode.DomainModels.EF/CachingEFRepository.cs
```diff
@@ -0,0 +1,67 @@
...
+{
+    protected CachingEFRepository(TContext dbContext)
+        : base(dbContext) { }
+
+    /// <summary>
+    /// Finds an entity by primary key. If the entity with provided key is tracked by the underlying
+    /// <see cref="DbContext" />, this method will return the cached entity.
+    /// </summary>
+    /// <remarks>For implementers: the default implementation won't work for composite primary keys.</remarks>
+    /// <param name="id">The identifier that</param>
```
Dragemil: Seems incomplete.

src/Domain/LeanCode.DomainModels.EF/CachingEFRepository.cs
```diff
@@ -0,0 +1,67 @@
...
+
+public abstract class CachingEFRepository<TEntity, TIdentity, TContext> : EFRepository<TEntity, TIdentity, TContext>
+    where TEntity : class, IAggregateRootWithoutOptimisticConcurrency<TIdentity>
+    where TIdentity : notnull
+    where TContext : DbContext
+{
+    protected CachingEFRepository(TContext dbContext)
+        : base(dbContext) { }
+
+    /// <summary>
```
Dragemil, line 17: Consider `<inheritdoc>`ing the docs from the `FindAsync` of the `EFRepository` and overridding necessary stuff.

jakubfijalkowski: Ah yes, forgot about that. Will do!

src/Domain/LeanCode.DomainModels.EF/CachingEFRepository.cs
jakubfijalkowski: All hail multitasking.

Will fix

test/LeanCode.IntegrationTests/EFRepositoryTests.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        var foundEntity = await repository.FindAsync(entity.Id);
+        foundEntity.Should().BeNull();
+    }
+
+    public Task InitializeAsync() => app.InitializeAsync();
+
+    public Task DisposeAsync() => app.DisposeAsync().AsTask();
+
+    class EntityRepository : EFRepository<Entity, Guid, TestDbContext>
```
marcin-dardzinski: Nit: I'd seal it as it asks. I don't like these warnings getting stacked everywhere.

src/Domain/LeanCode.DomainModels.EF/CachingEFRepository.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// Finds an entity by primary key. If the entity with provided key is tracked by the underlying
+    /// <see cref="DbContext" />, this method will return the cached entity.
+    /// </summary>
+    /// <remarks>For implementers: the default implementation won't work for composite primary keys.</remarks>
+    /// <param name="id">The identifier that</param>
+    /// <param name="cancellationToken">A <see cref="CancellationToken" /> to observe while waiting for the task to complete.</param>
+    /// <returns>Task with the found entity, or -null- if not found.</returns>
+    public override async Task<TEntity?> FindAsync(TIdentity id, CancellationToken cancellationToken = default)
+    {
+        return await FindTrackedOrLoadNewAsync(
```
marcin-dardzinski, line 23: Nit: why `async`/`await`? `AsTask` should be enough.

test/LeanCode.IntegrationTests/EFRepositoryTests.cs
jakubfijalkowski: Okay, I'm going with the nuclear option (treat warnings as errors) soon probably, otherwise these errors are freakishly invisible.

src/Domain/LeanCode.DomainModels.EF/CachingEFRepository.cs
jakubfijalkowski: Stacktrace. `FindTrackedOrLoadNewAsync` is internal, so if someone forgets to override it for composite key (or anything bad happens), they will  be left in the dark on what happened.

Saancreed (approved):

marcin-dardzinski (approved):

PR #576, opened 2023-09-29
FCM push notification token rework

Saancreed:
* Make `UserId` generic
* Use `{ UserId, Token }` as primary key instead of separate `Id` column
* Use shared, DB-agnostic store implementation
* Use EF ExecuteDelete / ExecuteSql APIs to be able to drop Dapper
* Move store tests to integration tests


src/Infrastructure/LeanCode.Firebase.FCM/PushNotificationTokenStore.cs
```diff
@@ -0,0 +1,201 @@
...
+                    when not matched then
+                        insert ({{userIdColumn}}, {{tokenColumn}}, {{dateCreatedColumn}})
+                        values ("nt"."UserId", "nt"."Token", "nt"."DateCreated");
+                    """,
+                    userId,
+                    newToken,
+                    TimeProvider.Time.UtcNow
+                ),
+                cancellationToken
+            );
```
michal12334, line 91: Does it work with both sql server and postgres?

src/Infrastructure/LeanCode.Firebase.FCM/ModelBuilderExtensions.cs
```diff
@@ -0,0 +1,36 @@
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
+
+namespace LeanCode.Firebase.FCM;
+
+public static class ModelBuilderExtensions
+{
+    public static void ConfigurePushNotificationTokenEntity<TUserId>(
```
jakubfijalkowski, line 8: We should not force users to know how long the token might be. This should be always handled by us.

src/Infrastructure/LeanCode.Firebase.FCM/PushNotificationTokenEntity.cs
```diff
@@ -0,0 +1,24 @@
+using LeanCode.TimeProvider;
+
+namespace LeanCode.Firebase.FCM;
+
+public class PushNotificationTokenEntity<TUserId>
```
jakubfijalkowski: Can we change this to a record?

src/Infrastructure/LeanCode.Firebase.FCM/ModelBuilderExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+        where TUserId : notnull, IEquatable<TUserId>
+    {
+        builder.ConfigurePushNotificationTokenEntity<TUserId>(
+            c => c.Property(e => e.Token).HasMaxLength(tokenColumnMaxLength ?? 512)
+        );
+    }
+
+    public static void ConfigurePushNotificationTokenEntity<TUserId>(
+        this ModelBuilder builder,
+        Action<EntityTypeBuilder<PushNotificationTokenEntity<TUserId>>>? callback = null
```
jakubfijalkowski: I don't really think we should allow this kind of callback here. If user wants to change _something_, they might define the entity themselves.

src/Infrastructure/LeanCode.Firebase.FCM/PushNotificationTokenStore.cs
jakubfijalkowski: Our integration tests, since recently, are run on both SQL Server & Postgres.

src/Infrastructure/LeanCode.Firebase.FCM/PushNotificationTokenStore.cs
```diff
@@ -0,0 +1,201 @@
...
+            var tokensTable = GetTokensTableName();
+            var userIdColumn = GetTokensColumnName(nameof(PushNotificationTokenEntity<TUserId>.UserId));
+            var tokenColumn = GetTokensColumnName(nameof(PushNotificationTokenEntity<TUserId>.Token));
+            var dateCreatedColumn = GetTokensColumnName(nameof(PushNotificationTokenEntity<TUserId>.DateCreated));
+
+            // perform an upsert using SQL merge, ensuring that given token is assigned to our user and no-one else
+            await dbContext.Database.ExecuteSqlInterpolatedAsync(
+                FormattableStringFactory.Create(
+                    $$"""
+                    merge into {{tokensTable}} "pt"
```
jakubfijalkowski: I will be the one - SQL SHOULD BE UPPERCASE.

src/Infrastructure/LeanCode.Firebase.FCM/ModelBuilderExtensions.cs
Saancreed: Okay, good point. I've noticed that tokens I'm getting are ~160 chars long but our column length was 512, is the length of those token documented anywhere to be possibly this long or was this an educated guess plus some safety buffer?

src/Infrastructure/LeanCode.Firebase.FCM/PushNotificationTokenEntity.cs
Saancreed: We could, but what's the point of structural equality for those tokens? `DateCreated` makes comparing instances tricky anyway and we don't actually ever instanciate those entities in memory.

Plus, now that we are actually updating token's users instead of replacing them, perhaps renaming this column to `DateModified` would be more appropriate?

src/Infrastructure/LeanCode.Firebase.FCM/ModelBuilderExtensions.cs
Saancreed: Hmm, okay. I can see nonzero value in someone wanting to, e.g. configure custom table/schema name but yeah, they can always to that separately.

src/Infrastructure/LeanCode.Firebase.FCM/ModelBuilderExtensions.cs
jakubfijalkowski: As far as I remember, there was a note in FCM docs that stated that the token length is unspecified, but it will never go above 512 chars or something like that. Will try to dig that out.

src/Infrastructure/LeanCode.Firebase.FCM/PushNotificationTokenEntity.cs
jakubfijalkowski: In my mind the change was not really related to the structural equality but only to the syntax of this class - will be shorter (and w/o public setters).

Saancreed: If we're to do this in name of shorter syntax, sure, I can try that.

What about `DateCreated` though, should it stay exactly as is?

src/Infrastructure/LeanCode.Firebase.FCM/ModelBuilderExtensions.cs
Saancreed: That would be great, if you find it then we could leave a comment explaining this choice with a link to those docs.

src/Infrastructure/LeanCode.Firebase.FCM/ModelBuilderExtensions.cs
jakubfijalkowski: I haven't really seen `callback` pattern used in EFCore models, hence my comment.

src/Infrastructure/LeanCode.Firebase.FCM/PushNotificationTokenEntity.cs
jakubfijalkowski: Your call. For me, `DateCreated` is perfectly OK. Although we modify the record, the _meaning_ stays the same - when the token itself has been created. `DateModified` would relate more to the record than the token, which also makes sense.

src/Infrastructure/LeanCode.Firebase.FCM/ModelBuilderExtensions.cs
Saancreed: I might have been inspired by MassTransit here: https://github.com/MassTransit/MassTransit/blob/master/src/Persistence/MassTransit.EntityFrameworkCoreIntegration/Configuration/EntityFrameworkOutboxConfigurationExtensions.cs#L145

jakubfijalkowski: TIL

src/Infrastructure/LeanCode.Firebase.FCM/ModelBuilderExtensions.cs
jakubfijalkowski: Okay, can't find that. The only "credible" source that I found is this [SO question](https://stackoverflow.com/questions/39959417/what-is-the-maximum-length-of-an-fcm-registration-id-token). The longest token that they found is now slightly more than 300 chars. I also might mistake this exact SO thread with docs now, as my previous search was a couple of years back.

src/Infrastructure/LeanCode.Firebase.FCM/PushNotificationTokenStore.cs
Saancreed: SQL being case insensitive was a mistake :weary:

jakubfijalkowski (approved):

michal12334 (approved):

docs/guides/advanced/v8_migration_guide.md
```diff
@@ -260,7 +279,10 @@ public void ConfigureServices(IServiceCollection services)
     {
         busCfg.AddEntityFrameworkOutbox<MyDbContext>(outboxCfg =>
         {
-            outboxCfg.UseSqlServer(); // Use your database flavor
+            // outboxCfg.UseSqlServer(); // Use your database flavor
+            // NOTE: Mass Transit's lock statement providers are currently incompatible with EF8, so for now use this instead:
+            outboxCfg.LockStatementProvider =
```
marcin-dardzinski: Just add a note that this is 'termporary'

Saancreed: I was hoping that `for now` would be clear enough? We can revert this once MT fixes it and we remove custom lock providers from our code (naturally forcing replacement).

Saancreed: Or I can just… not touch this and hopefully let it resolve itself in a few weeks.

Saancreed: Okay, whatever, I decided to just revert this change.

PR #580, opened 2023-10-06
Allow defining authorizers using only `ClaimsPrincipal`s

Dragemil:
They aren't in many cases limiting, yet they'll allow to be used in all kinds of topics in LeanPipe. Using authorizers basing on `HttpContext` is still possible.

### DIFF
--- a/src/CQRS/LeanCode.CQRS.Security/CustomAuthorizer.cs
+++ b/src/CQRS/LeanCode.CQRS.Security/CustomAuthorizer.cs
@@ -19,7 +19,7 @@ public interface ICustomAuthorizer : IHttpContextCustomAuthorizer
     ) => CheckIfAuthorizedAsync(context.User, obj, customData);
 }
 
-public abstract class CustomHttpContextAuthorizer<TObject> : IHttpContextCustomAuthorizer
+public abstract class HttpContextCustomAuthorizer<TObject> : IHttpContextCustomAuthorizer
 {
     public Task<bool> CheckIfAuthorizedAsync(HttpContext context, object obj, object? customData) =>
         CheckIfAuthorizedAsync(context, (TObject)obj);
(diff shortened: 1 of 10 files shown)


jakubfijalkowski (commented):
Tests.
We need to add the missing `CustomAuthorizer` tests plus now we should test the automatic DI registration (as it is no longer simple scan).

Dragemil:
`CustomAuthorizer` is as tested as it was before. Maybe `HttpContextCustomAuthorizer` was less tested, but now I've included it in CQRS integration tests.

As for that:
> we should test the automatic DI registration (as it is no longer simple scan)

How come, how adding a one inherited interface changes things?

jakubfijalkowski:
>  is as tested as it was before

I mean, we should use this opportunity to test things better. :)

> How come, how adding a one inherited interface changes things?

Because now, something that is ~`IHttpContextCustomAuthorizer` should also be registered as `ICustomAuthorizer`~ `ICustomAuthorizer` should also be registered as `IHttpContextCustomAuthorizer` automatically, but not the other way around. And everything should resolve.

Dragemil:
> > is as tested as it was before
> 
> I mean, we should use this opportunity to test things better. :)
> 
> > How come, how adding a one inherited interface changes things?
> 
> Because now, something that is `IHttpContextCustomAuthorizer` should also be registered as `ICustomAuthorizer` automatically, but not the other way around. And everything should resolve.

Right, actually exactly only the other way around, but right.

src/CQRS/LeanCode.CQRS.Security/CustomAuthorizer.cs
```diff
@@ -1,29 +1,70 @@
...
 {
-    public Task<bool> CheckIfAuthorizedAsync(HttpContext httpContext, object obj, object? customData) =>
-        CheckIfAuthorizedAsync(httpContext, (TObject)obj);
+    public Task<bool> CheckIfAuthorizedAsync(ClaimsPrincipal user, object obj, object? customData) =>
+        CheckIfAuthorizedAsync(user, (TObject)obj);
+
+    protected abstract Task<bool> CheckIfAuthorizedAsync(ClaimsPrincipal user, TObject obj);
+}
+
+public abstract class HttpContextCustomAuthorizer<TObject, TCustomData> : IHttpContextCustomAuthorizer
```
jakubfijalkowski, line 38: Overall, looks good, but can we get a set of (somewhat dummy) unit tests that check if the data is cast properly (and extracted properly)?

jakubfijalkowski (approved):

src/CQRS/LeanCode.CQRS.Security/CustomAuthorizer.cs
Dragemil: Possibly, yes.

jakubfijalkowski (approved):

PR #582, opened 2023-10-09
New docs

wojtek2288:
* New documentation. If you want to check out how does it look visit: https://leancode-corelibrary--582.org.readthedocs.build/ (I would recommend reading it there).
* This also uses old `General`  section from `README.md` as new github `README.md` (with some small overview addition).
* Removed `Check links` step from markdown lint check as it exceeded  nuget rate limits and failed.


docs/cqrs/authorization/index.md
```diff
@@ -0,0 +1,64 @@
+# Authorization
+
+Each command and query has to be authorized or must explicitly opt-out of authorization (it's enforced using Roslyn analyzers). You can specify which authorizer to use using the `AuthorizeWhen` attribute and custom `ICustomAuthorizer`. Opting-out is done using the `AllowUnauthorized` attribute. There is a predefined authorizer that uses role- and permission-based authorization. You can specify which permissions to enforce using `AuthorizeWhenHasAnyOf` and configure the role-to-permission relationship using `IRoleRegistrations`.
```
tomaszlaskowskiLC, line 3: command, query _and operation_?
I wonder do we have a clear term for all three concepts. Sometimes commands and queries are jointly called CQRS operations, but this would be misleading in our case. :thinking: 

If we pick one, it should be used consistently in all contexts. I would prefer a separate term instead of enumerating "comands, queries and operations". Unfortunately, I don't have an idea, maybe someone else would fill in. :smile:

docs/cqrs/authorization/index.md
```diff
@@ -0,0 +1,64 @@
...
+        HttpContext context,
+        IProjectRelated obj)
+    {
+        var project = await Access.FindAsync(new(obj.ProjectId), context.RequestAborted);
+        return project.OwnerId == context.GetEmployeeId();
+    }
+}
+```
+
+Both queries, commands and operations can (and should!) be behind authorization. By default, authorization is run before validation so the object that the command/query/operation is pointing at might not exist.
```
tomaszlaskowskiLC: Nitpick: "both" refers to two items, I guess the correct equivalent should be "all".

docs/cqrs/authorization/index.md
```diff
@@ -0,0 +1,64 @@
...
+
+If multiple `AuthorizeWhen` attributes are specified, **all** authorization rules must pass.
+
+An authorizer is a class that implements the `ICustomAuthorizer` interface or derives from one of the `CustomAuthorizer` base classes. It has access to both context and command/query. Command/query type doesn't need to be exact, it just has to be coercible to the specified type (`CustomAuthorizer` casts objects to the types internally). Therefore, if you want to use the same authorizer for many commands/queries, you can use base classes or interfaces and implement the authorizer for them.
+
+Example authorizer, along with the (not required, but convenient) plumbing:
```
tomaszlaskowskiLC: Examples for `AuthorizeWhenHasAnyOf` and `AllowUnauthorized` might also be useful.
[Task](https://linear.app/leancodepl/issue/COR-142/cqrs-authorization-additional-examples)

docs/cqrs/authorization/index.md
tomaszlaskowskiLC: As for " By default, authorization is run before validation". It might be good to add a snippet of code with chaining middlewares and example where custom authorizer do not fail if related object does not exist.
[Task](https://linear.app/leancodepl/issue/COR-143/cqrs-authorization-authorization-before-validation)

docs/cqrs/command/index.md
```diff
@@ -0,0 +1,76 @@
+# Command
+
+Command is just an class that implements the `ICommand` interface. Commands are used to execute an action that modifies data. Commands are validated, and if they pass validation, they should succeed. Commands do not return any value. This makes them quite constrained, yet reasoning is much easier.
```
tomaszlaskowskiLC: **a** class

docs/cqrs/command/index.md
```diff
@@ -0,0 +1,76 @@
...
+    public static class ErrorCodes
+    {
+        public const int InvalidName = 1;
+        public const int ProjectIdInvalid = 2;
+        public const int ProjectAlreadyExists = 3;
+    }
+}
+```
+
+that creates a new project. Caller of the command is required to have the `CreateProject` permission.
```
tomaszlaskowskiLC: I would rather not split paragraphs to include code snippet (if it's not one-liner), but that's a personal preference, so I'm open for the feedback from the rest.

docs/cqrs/authorization/index.md
```diff
@@ -0,0 +1,64 @@
...
+    public string ProjectId { get; set; }
+    public string Name { get; set; }
+}
+
+public class ProjectIsOwnedAuthorizer
+    : CustomAuthorizer<IProjectRelated>, IProjectIsOwned
+{
+    private readonly IRepository<Project, ProjectId> Projects;
+
+    public ProjectIsOwnedAuthorizer(IRepository<Project, ProjectId> Access)
```
tomaszlaskowskiLC: Can we just name both argument and field `projects`?

docs/cqrs/command/index.md
```diff
@@ -0,0 +1,76 @@
...
+            ProjectId.Parse(command.ProjectId),
+            command.Name,
+            context.EmployeeId());
+
+        // We only notify the repository that this is new entity,
+        // we let other part of the code commit the database transaction
+        projects.Add(project);
+
+        // `Execute` operations are async by nature, but here we don't need it
+        return Task.CompletedTask;
```
tomaszlaskowskiLC: I would prefer having a more typical async method e.g. fetching an aggregate and then performing the operation, but we can also do both. :slightly_smiling_face:

docs/cqrs/command/index.md
```diff
@@ -0,0 +1,76 @@
...
+    {
+        // context.EmployeeId() is an extension method defined elsewhere
+        var project = Project.Create(
+            ProjectId.Parse(command.ProjectId),
+            command.Name,
+            context.EmployeeId());
+
+        // We only notify the repository that this is new entity,
+        // we let other part of the code commit the database transaction
+        projects.Add(project);
```
tomaszlaskowskiLC: Regarding the comment: as stated some time ago I feel it needs more explanation and maybe just a separate subsection.

docs/cqrs/command/index.md
```diff
@@ -0,0 +1,76 @@
...
+
+1. Keep them simple and testable, do not try to model whole flows with a single command.
+2. Commands should rely on aggregates to gather the data (try not to use queries inside command handlers).
+3. Commands should modify just a single aggregate (try to `Add`/`Update`/`Delete` at most once).
+4. If the business process requires to modify multiple aggregates, try to use [events] (but don't over-engineer).
+5. If that does not help, modify/add/delete multiple aggregates.
+6. Do not throw exceptions from inside commands. The client will receive generic error (`500 Internal Server Error`). Do it only as a last resort.
+7. Database transaction will be commited at the end of the [pipeline] (assuming [CommitTransaction] pipeline element was added), so it's not recommended to commit it inside query handler as it may make serialized [events] inconsistent with the entity.
```
tomaszlaskowskiLC: 7. should be epmhasized, here is a separate [task](https://linear.app/leancodepl/issue/COR-144) for it.
Other points might also use examples -- [task](https://linear.app/leancodepl/issue/COR-144), although let's treat it as a low priority.

docs/cqrs/index.md
```diff
@@ -0,0 +1,18 @@
+# CQRS
+
+**Command-Query Responsibility Segregation** is a pattern that promotes hard split between parts of code (classes/objects) that read data ([query]) and the ones that modify and write the data ([command]). It also separates the payload (e.g. object id, user name) from the code that executes the query/command. This library went a step further and used commands and queries as client-facing API. It leverages `HttpContext` to define the environment in which the command or query handler operates. This design choice enables data retrieval from side channels such as cookies or HTTP headers.
```
tomaszlaskowskiLC: Let's split paragraph before "This library..." sentence. It introduces a separate important piece of information.

docs/cqrs/index.md
```diff
@@ -0,0 +1,18 @@
...
+- **Improved security:** Security measures can be applied more granularly. For instance, you might have different authorization mechanisms for read and write operations, providing an additional layer of control.
+- **Validation on the write side:** With CQRS, the write side typically handles commands that change the state of the system. This is where validation of incoming data and business rules can be enforced. By centralizing validation logic on the write side, you ensure that all changes to the system go through a consistent validation process.
+- **Flexibility in storage and models:** CQRS allows to use different data storage mechanisms for the read and write sides. For example, you might use a relational database for the write side and a NoSQL database for the read side. This flexibility enables you to choose the right tool for each job.
+
+> **Tip:** You can find more about how CQRS allows us to define the API in the [ContractsGenerator documentation](https://github.com/leancodepl/contractsgenerator/blob/main/docs/index.md).
```
tomaszlaskowskiLC: I'm not entirely sure that ContractsGenerator docs fill this promise. :smile:

docs/cqrs/operation/index.md
tomaszlaskowskiLC, line 1: I really like starting with potential use cases!

docs/cqrs/operation/index.md
```diff
@@ -0,0 +1,60 @@
...
+## Handler
+
+Operation handlers execute complex operations. They should not contain logic themselves, instead they should orchestrate commands, queries (and potentially other services) via `ICommandExecutor`, `IQueryExecutor` interfaces.
+
+For the above operation, you can have handler like this:
+
+```csharp
+public class PayForAccessOH : IOperationHandler<PayForAccess, PaymentTokenDTO>
+{
+    private readonly IPaymentsService payments;
```
tomaszlaskowskiLC: Ending service names with `Service` might be another topic for our freak fight weekly. :clown_face:

docs/cqrs/query/index.md
```diff
@@ -0,0 +1,58 @@
+# Query
+
+Query is just an class that implements the `IQuery<TResult>` interface (there's also non-generic `IQuery` interface but it shouldn't be used directly). The only generic parameter specifies the type that the query returns when executed. It should be a DTO (because most of the time it will be serialized). Queries get the data from the system but don't modify it.
```
tomaszlaskowskiLC: **a** class

docs/cqrs/query/index.md
```diff
@@ -0,0 +1,58 @@
...
+        HttpContext context,
+        FindProjectsMatchingName query)
+    {
+        // Here, we use raw SQL but you are free
+        // to use other mechanisms to get the data
+        var filter = $"%{query.NameFilter.ToLower()}%";
+        var results = await dbContext.QueryAsync<ProjectInfoDTO>(@"
+            SELECT ""Id"" AS ""ProjectId"", ""Name"" FROM ""Projects""
+            WHERE ""Name"" LIKE @filter",
+            new { filter });
```
tomaszlaskowskiLC: This seems a bit too obscure for the base example. Maybe let's just change to getting some object by Id and simple EF+Linq?

docs/cqrs/query/index.md
```diff
@@ -0,0 +1,58 @@
...
+
+```csharp
+public class ProjectInfoDTO
+{
+    public string ProjectId { get; set; }
+    public string Name { get; set; }
+}
+
+[AllowUnauthorized]
+public class FindProjectsMatchingName : IQuery<List<ProjectInfoDTO>>
```
tomaszlaskowskiLC: Query's name should describe the result and means of getting it e.g. `ProjectNameById`, emphasizing the read-side of CQRS. `Find` shouldn't be used as we would like to set an example here.

It would be also worth to add a short paragraph describing naming convention.

docs/external_integrations/messaging_masstransit/index.md
```diff
@@ -0,0 +1,171 @@
...
+
+The integration of CoreLibrary and [MassTransit] relies on the implementation of the [MassTransit] Inbox & Outbox pattern. To learn more about this pattern, you can refer to [the documentation](https://masstransit.io/documentation/patterns/transactional-outbox). In addition to adding this pattern within `AddCQRSMassTransitIntegration` in your configuration, it also needs to be incorporated within `OnModelCreating` when using Entity Framework, as demonstrated below:
+
+```csharp
+protected override void OnModelCreating(ModelBuilder builder)
+{
+    . . .
```
wojtek2288: TODO: change everywhere to `// . . .` to not fail syntax

wojtek2288: In fact, it has no impact on syntax highlighting whatsoever

docs/cqrs/pipeline/avoid_commiting_transactions_in_handlers.md
```diff
@@ -0,0 +1,132 @@
+# Avoid committing transactions in handlers
+
+> **Tip:** Before delving into this section, it's highly recommended to explore the [CQRS], [Domain] and [MassTransit] sections.
+
+Directly commiting transactions in [command]/[operation] handlers poses a challenge, potentially causing inconsistencies between [events] and the associated entities. Let's examine the pipeline configuration below:
```
tomaszlaskowskiLC: At this point I don't know if it's obvious for the reader, where this problem stems from. Maybe let's start with stating that transaction is committed in event publishing part of pipeline. Then present an example CH with commit transaction and state what problem it poses.

The advice however, can be put into quote (or highlighted in other way) at the top of the page.

docs/cqrs/query/index.md
```diff
@@ -0,0 +1,74 @@
...
+            {
+                Id = p.Id,
+                Name = p.Name
+            })
+            .ToListAsync(context.RequestAborted);
+    }
+}
+```
+
+## Naming conventions
```
tomaszlaskowskiLC: I would opt for moving this subsection up (under Contract).
Also it would be good to make it clear that namespace is also a part of contract name (same with `Command`). Similarly to what MT does [here](https://masstransit.io/documentation/concepts/messages) (apart from that MT shouldn't be a role model for us :stuck_out_tongue: ).

docs/cqrs/validation/index.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Packages
+
+| Package | Link | Application in section |
+| --- | ----------- | ----------- |
+| FluentValidation | [![NuGet version (LeanCode.CQRS.Validation.Fluent)](https://img.shields.io/nuget/vpre/FluentValidation.svg?style=flat-square)](https://www.nuget.org/packages/FluentValidation/11.8.0/) | Validation |
+| LeanCode.CQRS.Validation | [![NuGet version (LeanCode.CQRS.Validation)](https://img.shields.io/nuget/vpre/LeanCode.CQRS.Validation.svg?style=flat-square)](https://www.nuget.org/packages/LeanCode.CQRS.Validation/8.0.2260-preview/) | `ICommandValidator` |
+| LeanCode.CQRS.Validation.Fluent | [![NuGet version (LeanCode.CQRS.Validation.Fluent)](https://img.shields.io/nuget/vpre/LeanCode.CQRS.Validation.Fluent.svg?style=flat-square)](https://www.nuget.org/packages/LeanCode.CQRS.Validation.Fluent/8.0.2260-preview/) | `AbstractValidator` |
+
+To validate example command introduced in [command] section, you can use something like:
```
tomaszlaskowskiLC: It's better to say e.g. "following error codes and validator" than "something like".

tomaszlaskowskiLC: Or even "you can add following code". :stuck_out_tongue:

docs/cqrs/validation/index.md
```diff
@@ -0,0 +1,81 @@
...
+                UpdateProjectName.ErrorCodes.ProjectDoesNotExist
+            );
+        }
+    }
+}
+```
+
+> **Tip:** More on ids can be found [here](../../domain/id/index.md).
+
+If you need complex validation logic that needs to access external state, use `MustAsync`/`CustomAsync` validators. For `CustomAsync` validators, you can use `AddValidationError` helper to specify the error code.
```
tomaszlaskowskiLC: It would be nice to add an example of async validation (checking for existence with dbContext should suffice).

docs/domain/domain_event/index.md
```diff
@@ -0,0 +1,73 @@
...
+        EmployeeId employeeId)
+    {
+        Id = Guid.NewGuid();
+        DateOccurred = Time.UtcNow;
+
+        AssignmentId = assignmentId;
+        EmployeeId = employeeId;
+    }
+
+    [JsonConstructor]
```
tomaszlaskowskiLC, line 55: It might be worth mentioning why this additional constructor is necessary.

docs/cqrs/validation/index.md
wojtek2288: It was there above this, moved this text above async validator to make it more clear.

docs/external_integrations/authorization_ory_kratos/handling_webhooks.md
```diff
@@ -0,0 +1,183 @@
+# Handling webhooks
+
+To effectively manage incoming webhooks from Ory Kratos, you can employ the [KratosWebHookHandlerBase] class. The following example demonstrates how to synchronize identity data received from Kratos through webhooks with your database using [MassTransit]. This assumes that your Kratos instance is properly configured to send webhooks to your application.
```
Saancreed: > This assumes that your Kratos instance is properly configured to send webhooks to your application.

Consider showing a fragment of `kratos.yaml` and related Jsonnet snippet that actually shows how to do this. E.g. it might not be obvious that we expect the shared secret to be passed in `X-Api-Key` header by default but we allow it to be overriden per webhook handler.

docs/external_integrations/authorization_ory_kratos/handling_webhooks.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Packages
+
+| Package | Link | Application in section |
+| --- | ----------- | ----------- |
+| LeanCode.Kratos | [![NuGet version (LeanCode.Kratos)](https://img.shields.io/nuget/vpre/LeanCode.Kratos.svg?style=flat-square&logo=nuget)](https://www.nuget.org/packages/LeanCode.Kratos/8.0.2260-preview/) | Webhook handling |
```
Saancreed, line 9: Do we need to specifically link to `8.0.2260-preview` version? Maybe omitting this part of the path will just redirect the user to latest version?

docs/external_integrations/authorization_ory_kratos/handling_webhooks.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Mapping the Endpoint in Application Configuration
+
+First, you should map the `KratosIdentitySyncHandler` class to the `/kratos/sync-identity` endpoint in your application's configuration:
```
Saancreed: The exact path is not important as long as it matches relevant webhook configuration in `kratos.yaml` (and that part is not set in stone).

docs/external_integrations/authorization_ory_kratos/handling_webhooks.md
```diff
@@ -0,0 +1,183 @@
...
+    public KratosIdentitySyncHandler(
+        KratosWebHookHandlerConfig config,
+        IBus bus)
+        : base(config)
+    {
+        this.bus = bus;
+    }
+
+    protected override async Task HandleCoreAsync(HttpContext ctx)
+    {
```
Saancreed, line 97: Might be worth mentioning that at this point we have already handled API key check in the base class so it's not required to do this here again.

docs/external_integrations/authorization_ory_kratos/handling_webhooks.md
```diff
@@ -0,0 +1,183 @@
...
+            keyValues: new[] { (object)identityId },
+            context.CancellationToken
+        );
+
+        // Database transaction will be commited at the end of the pipeline
+        // assuming `CommitDatabaseTransactionMiddleware` was added
+        if (dbIdentity is null)
+        {
+            dbIdentity = new(kratosIdentity);
+            dbContext.KratosIdentities.Add(dbIdentity);
```
Saancreed, line 211: TODO for later: link to example implementation of this entity?

docs/external_integrations/authorization_ory_kratos/handling_webhooks.md
wojtek2288: It will redirect user to latest version which is not in preview (or latest one if all versions are in preview), so it could be removed after corelib v8.0 is out of preview.

docs/domain/domain_event/index.md
Dragemil, line 1: A section about unit testing whether domain event is raised could be useful.

docs/cqrs/authorization/index.md
Dragemil: We could adapt a bit the _CQRS objects_ terminology. Maybe go back to the roots calling them **R**CQRS objects (Remote), however we have abandoned this terminology I'm afraid. Otherwise, we can get a little bit silly and call them CQRS+ objects, where _+_ would mean the operation.

docs/features/audit_logs/index.md
Dragemil, line 1: Do we want to also mention the features not present in this repo? Like LeanPipe, app rating and probably something else. We could have a list somewhere with just one sentence long descriptions and links to the docs (LeanPipe e.g. has some docs in its repo)

docs/domain/domain_event/index.md
wojtek2288: Added to backlog

docs/features/audit_logs/index.md
wojtek2288: Yes, we could do it, but I can see that only app rating has public repo, leanpipe and notification center are private and I don't think that we should mention features that might not be available to user outside leancode.

jakubfijalkowski: Let's put that in the backlog - these repo will be open rather soon.

README.md
```diff
@@ -1,8 +1,164 @@
...
+
+Or you can just modify the following project template (most of the projects use this):
+
+```xml
+<Project Sdk="Microsoft.NET.Sdk">
+
+</Project>
+```
+
+[/Directory.Build.props]: https://github.com/leancodepl/corelibrary/blob/v8.0-preview/Directory.Build.props
```
marcin-dardzinski, line 162: This will be broken very soon. I wonder if we can make it a link to a default branch no matter what it is.

docs/cqrs/authorization/index.md
```diff
@@ -0,0 +1,213 @@
+# Authorization
+
+Each command and query has to be authorized or must explicitly opt-out of authorization (it's enforced using Roslyn analyzers). You can specify which authorizer to use using the `AuthorizeWhen` attribute and custom `ICustomAuthorizer`. Opting-out is done using the `AllowUnauthorized` attribute. There is a predefined authorizer that uses role- and permission-based authorization. You can specify which permissions to enforce using `AuthorizeWhenHasAnyOf` and configure the role-to-permission relationship using `IRoleRegistrations`.
```
marcin-dardzinski, line 3: > (it's enforced using Roslyn analyzers)

Do we have any entry in the docs covering this topic?

docs/cqrs/authorization/index.md
```diff
@@ -0,0 +1,213 @@
...
+| Package | Link | Application in section |
+| --- | ----------- | ----------- |
+| LeanCode.Contracts | [![NuGet version (LeanCode.Contracts)](https://img.shields.io/nuget/vpre/LeanCode.Contracts.svg?style=flat-square&logo=nuget)](https://www.nuget.org/packages/LeanCode.Contracts/2.0.0-preview.3/) | Default authorizers |
+| LeanCode.CQRS.Security | [![NuGet version (LeanCode.CQRS.Security)](https://img.shields.io/nuget/vpre/LeanCode.CQRS.Security.svg?style=flat-square&logo=nuget)](https://www.nuget.org/packages/LeanCode.CQRS.Security/8.0.2260-preview/) | Configuration, custom authorizers |
+
+## AuthorizeWhenHasAnyOf
+
+The `AuthorizeWhenHasAnyOf` attribute, found in `LeanCode.Contracts.Security`, has default authorization implementation. Upon its application, the `CheckIfAuthorizedAsync` method from the `DefaultPermissionAuthorizer` class is invoked to check whether the user possesses adequate permissions:
+
+```csharp
```
marcin-dardzinski: I think the implementation details are not needed here. On the other hand I don't see any example of how to use the attribute. `AppRoles` with the sample is good - maybe extend it with some sample contracts.

docs/cqrs/authorization/index.md
```diff
@@ -0,0 +1,213 @@
...
+Other than `AuthorizeWhenHasAnyOf` and `AllowUnauthorized` attributes which have default implementations custom authorizers can be defined. Here is an example along with the (not required, but convenient) plumbing:
+
+```csharp
+// Object that use `ProjectIsOwned` attribute must implement this interface
+public interface IProjectRelated
+{
+    string ProjectId { get; }
+}
+
+// Authorizer marker
```
marcin-dardzinski: I'd leave a backstory why it's needed and why it has to be implemented by the authorizer, i.e. DI resolving.

docs/cqrs/pipeline/index.md
```diff
@@ -0,0 +1,138 @@
...
+## Packages
+
+| Package | Link | Application in section |
+| --- | ----------- | ----------- |
+| LeanCode.CQRS.AspNetCore | [![NuGet version (LeanCode.CQRS.AspNetCore)](https://img.shields.io/nuget/vpre/LeanCode.CQRS.AspNetCore.svg?style=flat-square&logo=nuget)](https://www.nuget.org/packages/LeanCode.CQRS.AspNetCore/8.0.2260-preview/) | Configuration |
+| LeanCode.CQRS.MassTransitRelay | [![NuGet version (LeanCode.CQRS.MassTransitRelay)](https://img.shields.io/nuget/vpre/LeanCode.CQRS.MassTransitRelay.svg?style=flat-square&logo=nuget)](https://www.nuget.org/packages/LeanCode.CQRS.MassTransitRelay/8.0.2260-preview/) | MassTransit related middlewares |
+
+## Configuration
+
+To establish LeanCode CoreLibrary pipelines, the [MapRemoteCqrs(...)] method from `LeanCode.CQRS.AspNetCore` package needs to be invoked after calling the `UseEndpoints(...)` method on the `IApplicationBuilder`. In the following example, [MapRemoteCqrs(...)] is configured to handle:
```
marcin-dardzinski: This sentence feels awkward. Maybe sth like (rough scratch, nothing final):

CQRS objects can only be registered in the ASP.NET request pipeline via endpoint routing. To register use `IEndpointRouteBuilder.MapRemoteCqrs` extension methods (note that it's not the same as `UseEndpoints` as with the minimal startup you can call it directly). In `MapRemoteCqrs` you can configure the inner cqrs requests pipeline.

Also, I'd extend this sample to also include `IServceCollection.AddCQRS` method to make it self contained.

docs/cqrs/pipeline/adding_custom_middlewares.md
```diff
@@ -0,0 +1,74 @@
+
+# Adding custom middlewares
+
+Custom middlewares can further augment the pipeline by employing the `UseMiddleware` method on `IApplicationBuilder`. To illustrate, consider the following basic example, which introduces a check if employee is blocked into the pipeline:
```
marcin-dardzinski: Maybe it's worth to mention `CQRSRequestPayload` and `CQRSEndpointMetadata` objects that might be useful when writing middleware.

docs/domain/aggregate/index.md
```diff
@@ -0,0 +1,145 @@
...
+        {
+            Id = projectId,
+            Name = name,
+            OwnerId = ownerId,
+        };
+    }
+}
+```
+
+As you can see, the class implements `IAggregateRoot` interface - it marks the class as being the root of an aggregate. Moreover the `Id` field of the class is of type `ProjectId` - it is a special source-generated type present in the CoreLibrary. You can read more about `Id` types [here](../id/index.md). In this case, the Id of the project will look somewhat like `project_45a8f39f-9df0-4a23-b781-2a46de22fac1`.
```
marcin-dardzinski:
```suggestion
As you can see, the class implements `IAggregateRoot` interface - it marks the class as being the root of an aggregate. Moreover the `Id` field of the class is of type `ProjectId` - it is a special source-generated type present in the CoreLibrary. You can read more about `Id` types [here](../id/index.md). In this case, the Id of the project will look somewhat like `project_45a8f39f9df04a23b7812a46de22fac1`.
```

docs/tests/faking_time/index.md
```diff
@@ -0,0 +1,35 @@
...
+    TestTimeProvider.ActivateFake(expectedTime, timeZoneInfo: null);
+
+    var project = Project.Create(
+        ProjectId.New(),
+        "Project name",
+        EmployeeId.New());
+
+    // Assert that the `DateCreated` time matches
+    // the expected time in UTC.
+    Assert.Equal(expectedTime.UtcDateTime, project.DateCreated);
```
marcin-dardzinski, line 33: Nit: FluentAssertions

README.md
wojtek2288: We could point to the HEAD but we probably do not want docs just to point to newest commit. When v8 is released some other link also need to be updated. Will probably update it then.

docs/cqrs/authorization/index.md
wojtek2288: Not really, I will add this to backlog and reword it

docs/tests/faking_time/index.md
wojtek2288: Hm not sure about that, I think it should be as simple as possible and using FluentAssertions in docs might add some unnecessary complication (as it is not as popular and recognized XUnit's `Assert`).

docs/cqrs/pipeline/index.md
```diff
@@ -11,7 +11,7 @@ The LeanCode CoreLibrary utilizes ASP.NET middlewares to create customized pipel
 
 ## Configuration
 
-To establish LeanCode CoreLibrary pipelines, the [MapRemoteCqrs(...)] method from `LeanCode.CQRS.AspNetCore` package needs to be invoked after calling the `UseEndpoints(...)` method on the `IApplicationBuilder`. In the following example, [MapRemoteCqrs(...)] is configured to handle:
+CQRS objects can only be registered in the ASP.NET request pipeline via endpoint routing. To register use `IEndpointRouteBuilder.MapRemoteCqrs(...)` extension method (this differs from `UseEndpoints(...)`, as within minimal startup, you can directly invoke it). In `MapRemoteCqrs(...)` you can configure the inner cqrs requests pipeline. In the following example, app is configured to handle:
```
marcin-dardzinski: `(this differs from UseEndpoints(...), as within minimal startup, you can directly invoke it)` oh - this one was just a side note, shouldn't make it to the final docs.

docs/quick_start/index.md
```diff
@@ -0,0 +1,118 @@
...
+!!! Prerequisites info
+    Prior to proceeding with this tutorial, ensure you have the following requirements installed: [kubectl](https://kubernetes.io/docs/tasks/tools/),
+    [K3D](https://k3d.io/#installation), [Helm](https://helm.sh/docs/intro/quickstart/), [Tilt](https://docs.tilt.dev/install.html), [Terraform](https://www.terraform.io/), [dotnet](https://dotnet.microsoft.com/en-us/download), [git](https://git-scm.com/).
+
+## Setting up the template
+
+To get started, begin by cloning the repository containing `lncdproj` template:
+
+```sh
+git clone git@github.com:leancodepl/exampleapp.git
```
wojtek2288: Right now we also need to checkout to template branch but I think it will be merged by next wednesday.

jakubfijalkowski: I think the base will be merged even sooner, so we can assume working on master.

docs/README.md
```diff
@@ -1,29 +1,21 @@
-# LeanCode Core Library
+# Overview
 
-LeanCode Core Library, or CoreLib for short, is a set of helper libraries developed at [our company](https://leancode.pl) that aids our day-to-day development. It also provides a general guidelines on how we should build our apps.
+[![LeanCode Core Library](https://img.shields.io/badge/Repository-CoreLibrary-green?logo=github)](https://github.com/leancodepl/corelibrary)
+[![Nuget](https://img.shields.io/nuget/vpre/LeanCode.Components?logo=nuget)](https://www.nuget.org/packages/LeanCode.Components/8.0.2260-preview)
+[![License](https://img.shields.io/badge/License-Apache_2.0-green.svg)](https://www.apache.org/licenses/LICENSE-2.0)
 
-Our aim is to provide an opinionated framework for .NET Core app development. As for now we have, more-or-less, standardized:
+The LeanCode Core Library is a set of helper libraries developed at [our company](https://leancode.co/) that aids our day-to-day development. Not only does it serve as a facilitator in our day-to-day coding activities, but it also encapsulates comprehensive guidelines, gathers our collective knowledge on application architecture and development best practices.
```
jakubfijalkowski, line 7: Not now probably, but for future - we need to re-word this and put more focus on "CoreLib being a enterprise-ready framework that structures how you create the applications".

jakubfijalkowski: Sth like the second paragraph.

docs/quick_start/index.md
```diff
@@ -0,0 +1,118 @@
+# Quick start
+
+LeanCode CoreLibrary offers flexibility to suit different needs. Whether you want quick results or a detailed dive into features, it's here for you. In this tutorial, we'll take the DIY route, building, setting up CoreLibrary based project from scratch and learning how to deploy it on a local Kubernetes cluster.
+
+!!! Prerequisites info
+    Prior to proceeding with this tutorial, ensure you have the following requirements installed: [kubectl](https://kubernetes.io/docs/tasks/tools/),
+    [K3D](https://k3d.io/#installation), [Helm](https://helm.sh/docs/intro/quickstart/), [Tilt](https://docs.tilt.dev/install.html), [Terraform](https://www.terraform.io/), [dotnet](https://dotnet.microsoft.com/en-us/download), [git](https://git-scm.com/).
```
jakubfijalkowski, line 7: ~Do we really require `Helm`? Is that ever used as a binary?~

Okay, it is used.

docs/quick_start/index.md
```diff
@@ -0,0 +1,118 @@
...
+git clone git@github.com:leancodepl/exampleapp.git
+```
+
+Once the repository is cloned, navigate to the `exampleapp` directory. Here, install the `lncdproj` .NET template:
+
+```sh
+dotnet new install .
+```
+
+Then, you can run it in your preferred directory using the `dotnet new lncdproj` command. The `lncdproj` template requires two arguments: `--project-name`, specifying the name of the newly created project, and `--context`, indicating the name of the first bounded context used throughout the project (additional `--allow-scripts` flag is used to not display user prompts):
```
jakubfijalkowski: Listing arguments works better as a list. Making it a free-text requires a lot of "boilerplate".

docs/quick_start/index.md
```diff
@@ -0,0 +1,118 @@
...
+dotnet new lncdproj --allow-scripts Yes --project-name ProjectName --context ContextName
+```
+
+This template generates 3 directories:
+
+- `backend`: Contains the dotnet solution with template files.
+- `dev-cluster`: Holds the cluster configuration for local development.
+- `infrastructure`: Contains the configuration for infrastructure in production environments.
+
+## Setting up the local cluster
```
jakubfijalkowski: Let's include a short explanation _why_ we went this way.

docs/quick_start/index.md
```diff
@@ -0,0 +1,118 @@
+# Quick start
```
jakubfijalkowski, line 1: We also need a section describing what's in the template, because e.g. CoreLib does not depend on PostgreSQL nor Kratos directly, but the template forces that on you.

docs/quick_start/index.md
```diff
@@ -0,0 +1,118 @@
...
+- **Blob Storage**: A service for storing large amounts of unstructured data, like images or videos.
+- **RabbitMQ**: A message broker that enables applications to communicate with each other asynchronously.
+- **Kratos**: An identity management system that handles user accounts, authentication, and authorization.
+- **Seq**: Log system designed for searching, analysis, and visualization of structured log data.
+- **Jaeger**: An open-source tool for tracing requests.
+- **Traefik**: A cloud-native application proxy that routes incoming requests to the appropriate microservices.
+- **OpenTelemetry**: A set of APIs, SDKs, and tools to generate, collect, and export telemetry data.
+- **Metabase**: Tool for visualizing and sharing data insights.
+
+To configure the local cluster, start by navigating to the `dev-cluster` directory. Once there, ensure that you've set the `sendgrid_api_key` in the `terraform.tfvars` file. Without this api key, the functionality for sending emails won't be available:
```
jakubfijalkowski: I wouldn't consider e-mails a vital part of this description. There are things that deserve the attention first.

docs/quick_start/index.md
```diff
@@ -0,0 +1,118 @@
...
+
+```terraform
+sendgrid_api_key = "my_sendgrid_api_key"
+```
+
+Then, execute the following script to set up the local cluster:
+
+```sh
+./deploy.sh
```
jakubfijalkowski: Let's list the contents directly. `deploy.sh` is not x-plat.

docs/quick_start/index.md
```diff
@@ -0,0 +1,118 @@
...
+```
+
+## Setting up the application
```
jakubfijalkowski: I think that dev-cluster description should be _after_ the app description. In the end, app is more important here. :)

docs/quick_start/index.md
jakubfijalkowski: Also, this won't work if we don't have access to LeanCode's Azure account with `az` CLI. :(

docs/quick_start/index.md
```diff
@@ -0,0 +1,118 @@
...
+
+After successfully setting up the local cluster, your next step is to initiate your application. Begin by navigating to the `backend` directory. This directory is the heart of your application, containing various subdirectories and files, each serving a specific purpose:
+
+```txt
```
jakubfijalkowski, line 43: This will change soon, so let's remember to fix it after the fact.

docs/quick_start/index.md
```diff
@@ -0,0 +1,118 @@
...
+├── tests
+│   ├── ContextName
+│   │   ├── ProjectName.ContextName.Domain.Tests - Tests for business logic.
+│   │   └── ProjectName.ContextName.Services.Tests - Tests for service layer.
+│   ├── ProjectName.IntegrationTests - Comprehensive application-wide tests.
+```
+
+### Generating initial migrations
+
+Let's start with generating initial migrations. Begin by navigating, to the `backend/src/Apps/ProjectName.Migrations` directory. Generate initial migrations using the `dotnet ef` tool (install it via `dotnet tool install --global dotnet-ef`):
```
jakubfijalkowski: I talked with Krzysiek about this one, this tool IMO should be put in the `dotnet-tools.json`. /cc: @Saancreed

docs/quick_start/index.md
```diff
@@ -0,0 +1,118 @@
...
+Once initial migrations have been generated we can start the application. First of all ensure that you've selected the `kubectl` context corresponding to the newly created cluster:
+
+```sh
+kubectl config use-context k3d-projectname
+```
+
+Then you can run the migrations and the API using the following command:
+
+```sh
+tilt up migrations api
```
jakubfijalkowski: This will run API only, it will not start migrations.

docs/cqrs/index.md
```diff
@@ -0,0 +1,21 @@
+# CQRS
+
+**Command-Query Responsibility Segregation** is a pattern that promotes hard split between parts of code (classes/objects) that read data ([query]) and the ones that modify and write the data ([command]). It also separates the payload (e.g. object id, user name) from the code that executes the query/command.
+
+This library went a step further and used commands and queries as client-facing API. It leverages `HttpContext` to define the environment in which the command or query handler operates. This design choice enables data retrieval from side channels such as cookies or HTTP headers.
+
+The strict separation of [command] and [query] has sometimes led to awkwardness for clients consuming the API. So [operation] was introduced to add more flexibility. Operations change the state of the system, but also allow to return some result.
```
jakubfijalkowski: Either `some data` or `a result`.

docs/cqrs/index.md
```diff
@@ -0,0 +1,21 @@
+# CQRS
+
+**Command-Query Responsibility Segregation** is a pattern that promotes hard split between parts of code (classes/objects) that read data ([query]) and the ones that modify and write the data ([command]). It also separates the payload (e.g. object id, user name) from the code that executes the query/command.
+
+This library went a step further and used commands and queries as client-facing API. It leverages `HttpContext` to define the environment in which the command or query handler operates. This design choice enables data retrieval from side channels such as cookies or HTTP headers.
```
jakubfijalkowski: Mentioning `HttpContext` here is too early - it should describe the `why`, not the code-level details.

docs/cqrs/index.md
```diff
@@ -0,0 +1,21 @@
...
+
+The strict separation of [command] and [query] has sometimes led to awkwardness for clients consuming the API. So [operation] was introduced to add more flexibility. Operations change the state of the system, but also allow to return some result.
+
+## Why CQRS?
+
+- **Separation of concerns:** CQRS enforces a clear separation of concerns between the write and read operations. This leads to a more maintainable and modular codebase, making it easier to understand, test, and develop.
```
jakubfijalkowski: Let's try not to use `operation`, as it has some other meaning.

docs/cqrs/index.md
```diff
@@ -0,0 +1,21 @@
...
+- **Improved security:** Security measures can be applied more granularly. For instance, you might have different authorization mechanisms for read and write operations, providing an additional layer of control.
+- **Validation on the write side:** With CQRS, the write side typically handles commands that change the state of the system. This is where validation of incoming data and business rules can be enforced. By centralizing validation logic on the write side, you ensure that all changes to the system go through a consistent validation process.
```
jakubfijalkowski: I don't get the `typically` here - what does it refer to?

jakubfijalkowski: And I don't get the point fully. Could you paraphrase it to me?

docs/cqrs/index.md
```diff
@@ -0,0 +1,21 @@
...
+- **Flexibility in storage and models:** CQRS allows to use different data storage mechanisms for the read and write sides. For example, you might use a relational database for the write side and a NoSQL database for the read side. This flexibility enables you to choose the right tool for each job.
```
jakubfijalkowski: Let's emphasize that you _can_ do it but you _don't need to_. There is this misunderstanding of CQRS that you _have to_ have two datastores, one for writes, one for reads.

docs/cqrs/index.md
jakubfijalkowski, line 1: I would add some links that elaborate more on the CQRS topic:
1. https://martinfowler.com/bliki/CQRS.html
2. https://learn.microsoft.com/en-us/azure/architecture/patterns/cqrs

docs/cqrs/pipeline/index.md
```diff
@@ -0,0 +1,138 @@
+# Pipeline
+
+The LeanCode CoreLibrary utilizes ASP.NET middlewares to create customized pipelines for handling requests and responses. This section intends to showcase the setup of a basic pipeline and explore its inner workings.
```
jakubfijalkowski: We're in context of the CQRS pipelines, not arbitrary pipelines. I would not use "requests" and "responses" here.

docs/cqrs/pipeline/index.md
```diff
@@ -0,0 +1,138 @@
...
+## Packages
+
+| Package | Link | Application in section |
+| --- | ----------- | ----------- |
+| LeanCode.CQRS.AspNetCore | [![NuGet version (LeanCode.CQRS.AspNetCore)](https://img.shields.io/nuget/vpre/LeanCode.CQRS.AspNetCore.svg?style=flat-square&logo=nuget)](https://www.nuget.org/packages/LeanCode.CQRS.AspNetCore/8.0.2260-preview/) | Configuration |
+| LeanCode.CQRS.MassTransitRelay | [![NuGet version (LeanCode.CQRS.MassTransitRelay)](https://img.shields.io/nuget/vpre/LeanCode.CQRS.MassTransitRelay.svg?style=flat-square&logo=nuget)](https://www.nuget.org/packages/LeanCode.CQRS.MassTransitRelay/8.0.2260-preview/) | MassTransit related middlewares |
+
+## Configuration
+
+CQRS objects can only be registered in the ASP.NET request pipeline via endpoint routing. To register use `IEndpointRouteBuilder.MapRemoteCqrs(...)` extension method. In `MapRemoteCqrs(...)` you can configure the inner cqrs requests pipeline. In the following example, app is configured to handle:
```
jakubfijalkowski:
```suggestion
CQRS objects can only be registered in the ASP.NET request pipeline via endpoint routing. To register use `IEndpointRouteBuilder.MapRemoteCqrs(...)` extension method. In `MapRemoteCqrs(...)` you can configure the inner CQRS pipeline. In the following example, app is configured to handle:
```

docs/cqrs/index.md
```diff
@@ -0,0 +1,21 @@
...
+
+!!! tip
```
jakubfijalkowski, line 21: Let's incorporate ContractsGenerator docs in CoreLib. We won't have enough manpower to maintain both.

docs/cqrs/pipeline/index.md
```diff
@@ -0,0 +1,138 @@
...
+```
+
+In this code snippet, you can specify which middlewares to use for handling commands, queries, and operations. Several middlewares are added in the example:
+
+| Method                           | Middleware                             | Responsibility              |
+|------------------------------- |--------------------------------------- |---------------------------- |
+| [CQRSTrace()]                  | [CQRSTracingMiddleware]                | Tracing                     |
+| [Secure()]                     | [CQRSSecurityMiddleware]               | Authorization               |
+| [Validate()]                   | [CQRSValidationMiddleware]             | Validation                  |
+| [CommitTransaction&lt;T&gt;()]  | [CommitDatabaseTransactionMiddleware]  | Saving changes on database  |
```
jakubfijalkowski:
```suggestion
| [CommitTransaction&lt;T&gt;()]  | [CommitDatabaseTransactionMiddleware]  | Saves changes to the database  |
```

docs/cqrs/pipeline/index.md
```diff
@@ -0,0 +1,138 @@
...
+| [PublishEvents()]              | [EventsPublisherMiddleware]            | Publishing domain events    |
```
jakubfijalkowski:
```suggestion
| [PublishEvents()]              | [EventsPublisherMiddleware]            | Publishing domain events to MassTransit    |
```

docs/cqrs/index.md
jakubfijalkowski: Also, I would propose to incorporate the notion of events here. This section basically describes "how to communicate with the application/how to define an API" and events are part of the API. Having at least a short description on how they work/would be used would be good.

jakubfijalkowski: When it comes to structure - I don't think starting with Pipeline is the way to go. Let the user understand what is a command/query/operation first and then go into the details.

jakubfijalkowski: /cc: @tomaszlaskowskiLC

docs/cqrs/pipeline/index.md
```diff
@@ -0,0 +1,138 @@
...
+
+The order in which these middlewares are added determines the sequence of execution. Additionally, there are a few other middlewares provided by library that can be incorporated into CQRS pipeline, although they are not covered in this basic example:
```
jakubfijalkowski, line 65:
```suggestion
The order in which these middlewares are added determines the sequence of execution.

### Other middlewares
Additionally, there are a few other middlewares provided by library that can be incorporated into CQRS pipeline, although they are not covered in this basic example:
```

jakubfijalkowski: Plus, it would be good to describe how the middlewares are run - saying "sequence" is too vague. We can point to ASP.NET Core documentation on middlewares as an explanation TBH.

docs/cqrs/pipeline/adding_custom_middlewares.md
```diff
@@ -0,0 +1,80 @@
...
```
jakubfijalkowski: We should point here to the ASP.NET Core docs on middlewares so that the users are aware, that this is a _normal_ middleware.

docs/cqrs/pipeline/adding_custom_middlewares.md
```diff
@@ -0,0 +1,80 @@
...
+            this IApplicationBuilder builder)
+        {
+            return builder.UseMiddleware<EmployeeBlockerMiddleware>();
+        }
+    }
+```
+
+Moreover, CoreLibrary provides extension methods for `HttpContext`, which can be useful when creating middlewares:
+
+* `GetCQRSEndpoint()`: This method allows access to metadata about the endpoint, providing details such as object types, result types, and handler types involved in the request.
```
jakubfijalkowski: We should link to what these two return.

docs/cqrs/pipeline/adding_custom_middlewares.md
```diff
@@ -0,0 +1,80 @@
...
+                    "/api",
+                    cqrs =>
+                    {
+                        cqrs.Commands = c =>
+                            c.CQRSTrace()
+                            .BlockEmployees()
+                            .Secure()
+                            .Validate()
+                            .CommitTransaction<CoreDbContext>()
+                            .PublishEvents();
```
jakubfijalkowski:
```suggestion
                        cqrs.Commands = c => c
                            .CQRSTrace()
                            .BlockEmployees()
                            .Secure()
                            .Validate()
                            .CommitTransaction<CoreDbContext>()
                            .PublishEvents();
```

docs/cqrs/pipeline/avoid_commiting_transactions_in_handlers.md
```diff
@@ -0,0 +1,137 @@
...
+        projects.Update(project);
+
+        // Directly committing the transaction in the command handler,
+        // which should be avoided.
+        await dbContext.SaveChangesAsync(context.RequestAborted)
+    }
+}
+```
+
+In our pipeline configuration, [events] are published after the [command] handler is executed. This implies that changes on `CoreDbContext` will be committed before [events] are published. In LeanCode Corelibrary, we utilize [MassTransit] for message broker interaction, employing the [transactional outbox](https://masstransit.io/documentation/patterns/transactional-outbox) concept. This ensures that messages are committed to the database before being available to message brokers after publication and the invocation of the `SaveChangesAsync` method on CoreDbContext.
```
jakubfijalkowski:
```suggestion
In our pipeline configuration, [events] are published after the [command] handler is executed. This implies that changes on `CoreDbContext` will be committed before [events] are published. In LeanCode CoreLibrary, we utilize [MassTransit] for message broker interaction, employing the [transactional outbox](https://masstransit.io/documentation/patterns/transactional-outbox) concept. This ensures that messages are committed to the database before being available to message brokers after publication and the invocation of the `SaveChangesAsync` method on CoreDbContext.
```

Also, the idea behind transactional outbox is not to wait for changes to the data before publishing, but to ensure the messages will be published **if and only if** the changes were made, so that the messages are neither lost, nor published even if save failed.

jakubfijalkowski: You describe that later, but this paragraph somewhat misleads the user - too much of a mental shortcut.

docs/cqrs/pipeline/avoid_commiting_transactions_in_handlers.md
```diff
@@ -0,0 +1,137 @@
...
+
+In our scenario, if the database fails after successfully saving changes to the project, the `EmployeeAssignedToAssignment` message won't be saved, leading to it being unavailable to message brokers and not sent.
+
+Conversely, removing `SaveChangesAsync` from the [command] handler would result in both messages and project changes being committed in a single transaction. In the event of a database failure, neither project changes nor messages would be saved or sent, providing clients with information about the request failure without unintended side effects. This is why it's not recommended to not commit transactions directly in [command]/[operation] handlers.
+
+!!! tip
+    To read more about LeanCode Corelibrary MassTransit integtation visit [here](../../external_integrations/messaging_masstransit/index.md).
```
jakubfijalkowski:
```suggestion
    To read more about LeanCode CoreLibrary MassTransit integtation visit [here](../../external_integrations/messaging_masstransit/index.md).
```

docs/cqrs/pipeline/avoid_commiting_transactions_in_handlers.md
```diff
@@ -0,0 +1,137 @@
...
+
+In our scenario, if the database fails after successfully saving changes to the project, the `EmployeeAssignedToAssignment` message won't be saved, leading to it being unavailable to message brokers and not sent.
+
+Conversely, removing `SaveChangesAsync` from the [command] handler would result in both messages and project changes being committed in a single transaction. In the event of a database failure, neither project changes nor messages would be saved or sent, providing clients with information about the request failure without unintended side effects. This is why it's not recommended to not commit transactions directly in [command]/[operation] handlers.
```
jakubfijalkowski:
```suggestion
Conversely, removing `SaveChangesAsync` from the [command] handler would result in both messages and project changes being committed in a single transaction. In the event of a database failure, neither project changes nor messages would be saved or sent, providing clients with information about the request failure without unintended side effects. This is why it's not recommended to commit transactions directly in [command]/[operation] handlers, nor outside the `CommitDatabaseTransactionMiddleware`.
```

docs/cqrs/command/index.md
```diff
@@ -0,0 +1,98 @@
+# Command
+
+Command is just a class that implements the `ICommand` interface. Commands are used to execute an action that modifies data. Commands are validated, and if they pass validation, they should succeed. Commands do not return any value. This makes them quite constrained, yet reasoning is much easier.
```
jakubfijalkowski:
```suggestion
Command is a class that implements the `ICommand` interface. Commands are used to execute an action that modifies data. Commands are validated, and if they pass validation, they should succeed. Commands do not return any value. This makes them quite constrained, yet reasoning is much easier.
```

docs/cqrs/command/index.md
```diff
@@ -0,0 +1,98 @@
...
+    public string Name { get; set; }
+
+    public static class ErrorCodes
+    {
+        public const int InvalidName = 1;
+        public const int ProjectIdInvalid = 2;
+        public const int ProjectDoesNotExist = 3;
+    }
+}
+```
```
jakubfijalkowski, line 31: We should then explain step-by-step what the user has seen - the intent is easy to understand if you know what we mean by command, but it might not be obvious to someone that was exposed to CQRS recently.

docs/cqrs/command/index.md
```diff
@@ -0,0 +1,98 @@
...
+
+!!! tip
+    More on authorization and permissions can be found [here](../authorization/index.md) and on error codes and validation [here](../validation/index.md).
+
+## Naming conventions
+
+Commands in CQRS are responsible for modifying the state of the system, and therefore their names should reflect the action they trigger. A common convention is to use imperative verbs that clearly describe the intent, including the namespace as part of the contract, such as:
```
jakubfijalkowski, line 37: I don't get the `including the namespace as part of the contract`. What did you have in mind here?

docs/cqrs/command/index.md
```diff
@@ -0,0 +1,98 @@
...
+
+* `ExampleApp.Contracts.Projects.UpdateProjectName`
+* `ExampleApp.Contracts.Projects.DeleteProject`
+* `ExampleApp.Contracts.Employees.AssignEmployeeToAssignment`
+
+Handlers should share the identical name as the associated contract, appended with the `CH` suffix. This practice enhances ease of identification and ensures a straightforward correlation between contracts and their corresponding handlers.
```
jakubfijalkowski:
```suggestion
Handlers should share the identical name as the associated contract, with the `CH` suffix appended. This practice enhances ease of identification and ensures a straightforward correlation between contracts and their corresponding handlers.
```

docs/cqrs/command/index.md
```diff
@@ -0,0 +1,98 @@
...
+}
+```
+
+!!! tip
+    More on ids can be found [here](../../domain/id/index.md) and on validation [here](../validation/index.md).
+
+As you can see, the command handler is really simple - it just finds project with specified `ProjectId` and updates it's name. That does not mean this is the only responsibility of the handlers (it's just an example), but there are some guidelines related to them:
+
+1. Keep them simple and testable, do not try to model whole flows with a single command.
+2. Commands should rely on [aggregates] to gather the data (try not to use queries inside command handlers).
```
jakubfijalkowski:
```suggestion
2. Commands should rely on [aggregates] to gather the data.
```

Going into "queries" here is slightly too much if you don't know the original intent.

docs/cqrs/query/index.md
```diff
@@ -0,0 +1,84 @@
+# Query
+
+Query is just a class that implements the `IQuery<TResult>` interface (there's also non-generic `IQuery` interface but it shouldn't be used directly). The only generic parameter specifies the type that the query returns when executed. It should be a DTO (because most of the time it will be serialized). Queries get the data from the system but don't modify it.
```
jakubfijalkowski:
```suggestion
Query is a class that implements the `IQuery<TResult>` interface. The only generic parameter specifies the type that the query returns when executed. It should be a DTO (because most of the time it will be serialized). Queries get the data from the system but don't modify it.
```

> (there's also non-generic `IQuery` interface but it shouldn't be used directly)
This should be a remark somewhere down below. This is too detailed for intro paragraph.

docs/cqrs/query/index.md
```diff
@@ -0,0 +1,84 @@
...
+## Packages
+
+| Package | Link | Application in section |
+| --- | ----------- | ----------- |
+| LeanCode.Contracts | [![NuGet version (LeanCode.Contracts)](https://img.shields.io/nuget/vpre/LeanCode.Contracts.svg?style=flat-square&logo=nuget)](https://www.nuget.org/packages/LeanCode.Contracts/2.0.0-preview.3/) | `IQuery` |
+| LeanCode.CQRS.Execution | [![NuGet version (LeanCode.CQRS.Execution)](https://img.shields.io/nuget/vpre/LeanCode.CQRS.Execution.svg?style=flat-square&logo=nuget)](https://www.nuget.org/packages/LeanCode.CQRS.Execution/8.0.2260-preview/) | `IQueryHandler` |
+
+## Contract
+
+Consider the query that finds all projects that match the name filter. It may be called anonymously and returns a list of `ProjectDTO`s (we use a `List` instead of a `IList` or `IReadOnlyList` because of the DTO constraint; `List` is more DTO-ish than any interface):
```
jakubfijalkowski:
```suggestion
Consider the query that finds all projects that match the name filter. It may be called anonymously and returns a list of `ProjectDTO`s:
```

> (we use a `List` instead of a `IList` or `IReadOnlyList` because of the DTO constraint; `List` is more DTO-ish than any interface):

Same as above - this probably should be a note somewhere below, not here directly.

docs/cqrs/query/index.md
```diff
@@ -0,0 +1,84 @@
...
+
+* `ExampleApp.Contracts.Projects.AllProjects`
+* `ExampleApp.Contracts.Projects.ProjectById`
+* `ExampleApp.Contracts.Employees.EmployeesInAssignment`
+
+Query handlers should similarly be named in alignment with the associated query, appending the QH suffix within the respective namespace structure.
+
+## Handler
+
+Query handlers execute queries. They should not have any side effects but can return data back to the client. Since they can return data to the client, they don't need separate [validation] (handler can do it internally). In query handlers you don't need to operate on aggregate level (as this is read-side and is relatively DDD-free) and are allowed to perform arbitrary SQL queries.
```
jakubfijalkowski:
```suggestion
Query handlers execute queries. They should not have any side effects but can return data back to the client. Since they can return data to the client, they don't need separate [validation] (it is rare to need validation; if it is needed, handlers can do it internally). In query handlers you don't need to operate on aggregate level (as this is read-side and is relatively DDD-free) and are allowed to perform arbitrary SQL queries.
```

docs/quick_start/index.md
wojtek2288: I think it's worth at least to mention that it could be optionally set, as otherwise user will get few annoying prompts for   `sendgrid_api_key` during `terraform apply`.

jakubfijalkowski: I wonder if we should set defaults there in the first place.

docs/quick_start/index.md
wojtek2288: This is what we always had in `README` if I remember correctly it did start the migrations when I ran it. Anyway I separated those commands now.

docs/quick_start/index.md
wojtek2288: I will update this when it will be in `dotnet-tools.json`

docs/quick_start/index.md
```diff
@@ -8,6 +8,8 @@ LeanCode CoreLibrary offers flexibility to suit different needs. Whether you wan
 
 ## Setting up the template
 
+The `lncdproj` template, provided by LeanCode CoreLibrary, is a comprehensive template designed for modern .NET application development, featuring CQRS as client-facing APIs, PostgreSQL for database management, and Kratos for identity management. It integrates MassTransit with RabbitMQ for efficient communication and includes tools to aid testing. Additionally, the template incorporates LeanPipe for real-time SignalR notifications, Blob Storage for managing large datasets, and resources like Seq for logging, Jaeger for tracing, and Traefik as an application proxy, collectively laying a solid foundation for building efficient .NET applications.
```
jakubfijalkowski: And most of the things explained here are dev-only. ;P

docs/quick_start/index.md
```diff
@@ -82,37 +60,84 @@ backend
...
+```
+
+Follow these commands to establish your local cluster, or alternatively, use the `deploy.sh` script for an automated setup:
+
+```sh
+# Remove existing cluster and its registry to ensure a clean state.
+k3d cluster delete projectname
+k3d registry delete k3d-projectname-registry.local.lncd.pl
+rm *.tfstate*
+
```
jakubfijalkowski: We can remove these lines from this document - for the first run, they are totally unnecessary.

docs/quick_start/index.md
```diff
@@ -82,37 +60,84 @@ backend
...
+# Log in to Azure Container Registry and pull the Traefik proxy image.
```
jakubfijalkowski: And, I think, we should also remove this section or at least add a comment that this requires access to LeanCode's ACR and will be fixed in future version.s

docs/quick_start/index.md
```diff
@@ -8,7 +8,7 @@ LeanCode CoreLibrary offers flexibility to suit different needs. Whether you wan
...
+The `lncdproj` template, provided by LeanCode CoreLibrary, is a comprehensive template designed for modern .NET application development, featuring CQRS as client-facing APIs, PostgreSQL for database management, and Kratos for identity management. It integrates MassTransit with RabbitMQ for efficient communication and includes tools to aid testing. Additionally, the template incorporates LeanPipe for real-time SignalR notifications, Blob Storage for managing large data, and resources like Seq for logging, Jaeger for tracing, and Traefik as an application proxy, collectively laying a solid foundation for building efficient .NET applications in a development setting.
```
jakubfijalkowski: This does not really solve the problem. What is "in a development setting"? LeanPipe? Blob Storage? We should focus here on the template from the perspective of deployed application, leaving dev setting to the section below that describes it well already.

wojtek2288: Okay, included here only things that are common for both dev and production setting.

docs/quick_start/index.md
```diff
@@ -0,0 +1,134 @@
+# Quick start
+
+LeanCode CoreLibrary offers flexibility to suit different needs. Whether you want quick results or a detailed dive into features, it's here for you. In this tutorial, we'll take the DIY route, building, setting up CoreLibrary based project from scratch and learning how to deploy it on a local Kubernetes cluster.
```
jakubfijalkowski:
```suggestion
LeanCode CoreLibrary offers flexibility to suit different needs. Whether you want quick results or a detailed dive into features, it's here for you. In this tutorial, we'll take the DIY route, setting up CoreLibrary based project from scratch and learning how to deploy it on a local Kubernetes cluster.
```

docs/quick_start/index.md
```diff
@@ -0,0 +1,134 @@
...
+
+!!! Prerequisites info
+    Prior to proceeding with this tutorial, ensure you have the following requirements installed: [kubectl](https://kubernetes.io/docs/tasks/tools/),
+    [K3D](https://k3d.io/#installation), [Helm](https://helm.sh/docs/intro/quickstart/), [Tilt](https://docs.tilt.dev/install.html), [Terraform](https://www.terraform.io/), [dotnet](https://dotnet.microsoft.com/en-us/download), [git](https://git-scm.com/).
+
+## Setting up the template
+
+The `lncdproj` template, provided by LeanCode CoreLibrary, is a comprehensive template designed for modern .NET application development, featuring CQRS as client-facing APIs, PostgreSQL for database management, and Kratos for identity management. It integrates with MassTransit for efficient communication and includes tools to aid testing. Additionally, the template incorporates LeanPipe for real-time SignalR notifications, and Blob Storage for managing large data, collectively laying a solid foundation for building efficient .NET applications.
```
jakubfijalkowski: Let's add link to the:
1. PostgreSQL,
2. Kratos,
3. MassTransit,
4. LeanPipe repo,
5. SignalR,
6. Blob Storage (and let's call it Azure Blob Storage).

docs/quick_start/index.md
```diff
@@ -0,0 +1,134 @@
...
+
+To get started, begin by cloning the repository containing `lncdproj` template:
+
+```sh
+git clone git@github.com:leancodepl/exampleapp.git
```
jakubfijalkowski: Let's switch to HTTPS repo - if user does not have SSH keys (which is quite common, although I don't know why), they will have problems here. HTTPS will work for all.

docs/quick_start/index.md
jakubfijalkowski: [Sooner than I anticipated](https://github.com/leancodepl/exampleapp/pull/61)

jakubfijalkowski (commented):
Last couple of nits when it comes to quick start.

docs/domain/index.md
jakubfijalkowski, line 1: Maybe let's also add links that would help people get a grip on DDD?
1. [Evans](https://www.amazon.com/Domain-Driven-Design-Tackling-Complexity-Software/dp/0321125215),
2. [Vernon 1](https://www.amazon.com/Implementing-Domain-Driven-Design-Vaughn-Vernon-ebook/dp/B00BCLEBN8),
3. [Vernon 2](https://www.amazon.com/Domain-Driven-Design-Distilled-Vaughn-Vernon/dp/0134434420),
Maybe some more?

docs/domain/aggregate/index.md
```diff
@@ -0,0 +1,145 @@
+# Aggregate
+
+Aggregates, as defined by Domain Driven Design, are clusters of related objects which can be treated as a single domain entity. One of these objects is distinguished as an aggregate root. Every object of an aggregate must be accessed through the aggregate root. Transactions should not cross aggregate boundaries.
```
jakubfijalkowski, line 3: This somewhat assumes that user knows what are entities. If they do not, they will miss one of the most important properties of aggregates, that they are identifiable.

docs/domain/aggregate/index.md
jakubfijalkowski, line 1: Not for now, but we should probably start adding "best practices" for aggregates. One is already in the text:
1. Don't expect failure (i.e. don't validate something that should be validated outside) but ensure correctness.

docs/domain/identifiable/index.md
jakubfijalkowski, line 1: Let's name this section `Entity`, not `Identifiable`. The `Identifiable` name was (and still is) wrong.

docs/domain/domain_event/index.md
jakubfijalkowski, line 1: We should write down the rules of events and how they are handled, i.e.:
1. They are published transactionally,
2. They are handled outside of the "current" transaction,
3. They are not published outside of a bounded context.
And probably some more.

Also, we probably need a deep-dive into how they work underneath, because there is a healthy dose of "magic" there.

docs/domain/domain_event/index.md
```diff
@@ -0,0 +1,76 @@
...
+## Packages
+
+| Package | Link | Application in section |
+| --- | ----------- | ----------- |
+| LeanCode.DomainModels | [![NuGet version (LeanCode.DomainModels)](https://img.shields.io/nuget/vpre/LeanCode.DomainModels.svg?style=flat-square&logo=nuget)](https://www.nuget.org/packages/LeanCode.DomainModels/8.0.2260-preview/) | `IAggregateRoot`, `IDomainEvent` |
+| LeanCode.DomainModels.Generators | [![NuGet version (LeanCode.DomainModels.Generators)](https://img.shields.io/nuget/vpre/LeanCode.DomainModels.Generators.svg?style=flat-square&logo=nuget)](https://www.nuget.org/packages/LeanCode.DomainModels.Generators/8.0.2260-preview/) | Ids |
+
+## Example
+
+Let's imagine that we want to perform some action after a employee has been assigned to a assignment. We will modify `AssignEmployeeToAssignment` method from the `Project` class:
```
jakubfijalkowski, line 14: I would be somewhat wary of wording it like that. This somewhat states that the events should be created as a response to the need of doing something. This mental model suggests that you should raise an event whenever you need to do something, which might get twisted into events like `SendEmailToEmployee`.

docs/domain/id/index.md
```diff
@@ -1,31 +1,25 @@
-# IDs
+# Id
 
-CoreLib supports three different flavors of IDs:
+[Aggregate] requires you to specify identity type. Every aggregate has an `Id` property of the specified type. CoreLib supports three different flavors of IDs:
```
jakubfijalkowski, line 3: TBH, it's entities (And aggregates follow) that require identity ;P

docs/domain/id/index.md
```diff
@@ -1,31 +1,25 @@
...
-The domain part of the library supports a set of generic IDs:
+## Source generated IDs
 
-1. [Id&lt;T&gt;],
-2. [IId&lt;T&gt;],
-3. [LId&lt;T&gt;],
-4. [SId&lt;T&gt;].
-
-All the types give you type safety when passing the IDs, without introducing penalty (they basically work as `newtype`s). Unfortunately, they require an entity to be defined beforehand - it works as a generic parameter. This means you can't use it without a corresponding entity type. This poses a problem if you want to use the ID outside the parent domain. It is also quite hard to use - you need to know the exact ID format before you reference it (you need to choose between the four types when you just want to reference other entity).
+Source generated IDs leverage [Source Generators](https://learn.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/source-generators-overview) to generate fully functional ID structs that can work as IDs for entities. They are specialized for a particular entity type, but don't use the generic mechanisms of the language. They support the same feature set as aforementioned IDs.
```
jakubfijalkowski, line 20: This paragraph assumes that someone used `SId` (et al.) earlier - otherwise, they don't really know what generics are you talking about.

docs/domain/time_provider/index.md
```diff
@@ -0,0 +1,42 @@
+# Time provider
+
+`LeanCode.TimeProvider` is utilized to centralize and manage time-related operations. It enables the domain to encapsulate temporal logic, ensuring consistency and testability. Employing a dedicated time provider, facilitates easier maintenance and testing of time-dependent domain logic.
```
jakubfijalkowski, line 3: That's a very extensive description of the feature that boils down to "you can't mock DateTime.Now so now we have this". :D

docs/cqrs/index.md
wojtek2288: I disagree about starting with command/query/operation. Reader has some basic understanding about commands/queries/operations after reading initial CQRS section. I think it worth to explain at the beginning how requests in CoreLib are being handled otherwise user will just get contract and  handler classes not knowing anything about how does it happen that request goes into some handler and a lot of questions will arise e.g. How does contract connect with handler? Do I need to define my controllers? Do I create this command object and then somehow pass it into handler? This would be explained later in pipeline but when reading command/query/operation sections it will be hard to grasp how does it work.

jakubfijalkowski: I disagree that you have basic understanding on what a C/Q/O is from the initial section. That section describes all three in a basically single paragraph. You only get a very brief description of why we use that (r/w separation), then a single sentence that the payload (what the heck is payload here) is separated from execution.

When reading `Pipelines`, you read about connecting things that you have never seen yet. And the questions you will have then is "what is query", "what is command", "what is execution", which are much more basic.

docs/cqrs/index.md
wojtek2288: I think `typically` is unnecessary here and can be removed/

The point is that we do not need to include validation on our read side. I will reword it to make it more clear.

docs/cqrs/command/index.md
wojtek2288: This was @tomaszlaskowskiLC suggestion somewhere above. Namespace might sometimes provide additional information about the contract, and I think it should be treated as a part of it so that we do not create some weird folder structure which might confuse the clients.

docs/cqrs/pipeline/index.md
wojtek2288: It's responsibility section so `Saving`

tomaszlaskowskiLC (approved):
:clap: :checkered_flag:

PR #586, opened 2023-10-11
Pass cancellation token in CustomAuthorizer

Dragemil:


src/CQRS/LeanCode.CQRS.Security/CustomAuthorizer.cs
```diff
@@ -10,13 +10,18 @@ public interface IHttpContextCustomAuthorizer
 
 public interface ICustomAuthorizer : IHttpContextCustomAuthorizer
 {
-    Task<bool> CheckIfAuthorizedAsync(ClaimsPrincipal user, object obj, object? customData);
+    Task<bool> CheckIfAuthorizedAsync(
+        ClaimsPrincipal user,
+        object obj,
+        object? customData,
+        CancellationToken ct = default
```
jakubfijalkowski: Can we use the full form (`cancellationToken`) here, please? Rest of the code uses full name (_most_ of the time) and we should stick with it.

jakubfijalkowski (changes requested):

jakubfijalkowski (approved):

PR #587, opened 2023-10-11
Introduce LeanCode.AuditLogs project

lukaszgarstecki:


cloud-tests-infrastructure/storage.tf
```diff
@@ -0,0 +1,27 @@
+module "storage" {
+  source = "git@github.com:leancodepl/terraform-common-modules.git//azure_blob_storage?ref=v0.1.0"
```
jakubfijalkowski: The `leancodepl/terraform-common-modules` repo is private. Can we not reference it here and do most of the things manually?

cloud-tests-infrastructure/readme.md
```diff
@@ -1,5 +1,5 @@
 # Cloud tests infrastructure
 
-Cloud resources used in [LeanCode.Azure.Tests](test/Infrastructure/LeanCode.Azure.Tests/LeanCode.Azure.Tests.csproj)
+Cloud resources used in [LeanCode.Azure.Tests](../test/Infrastructure/LeanCode.Azure.Tests/LeanCode.Azure.Tests.csproj) and [LeanCode.AuditLogs.Tests](../test/Infrastructure/LeanCode.AuditLogs.Tests/LeanCode.AuditLogs.Tests.csproj)
```
jakubfijalkowski: I'm not sure we should have a separate project for these tests. `LeanCode.Azure.Tests` seems sufficient to contain both of them.

src/Infrastructure/LeanCode.AuditLogs/AuditLogsFilter.cs
```diff
@@ -0,0 +1,78 @@
...
+    public void Probe(ProbeContext context) { }
+
+    public async Task Send(
+        ConsumerConsumeContext<TConsumer, TMessage> context,
+        IPipe<ConsumerConsumeContext<TConsumer, TMessage>> next
+    )
+    {
+        await next.Send(context);
+
+        var entitiesChanged = ChangedEntitiesExtractor.Extract(dbContext);
```
jakubfijalkowski: This code is shared across filter & middleware. Can we share it?

src/Infrastructure/LeanCode.AuditLogs/AzureBlobAuditLogStorage.cs
```diff
@@ -0,0 +1,129 @@
...
+            }
+            else
+            {
+                throw;
+            }
+        }
+    }
+
+    private async Task AppendLogToBlobAsync(AuditLogMessage auditLogMessage, CancellationToken cancellationToken)
+    {
```
jakubfijalkowski, line 56: Let's refactor this (simply extracting methods) into:
1. `CreateBlobAsync` (that gets the `BlobClient` and calls `CreateIfNotExistAsync`)
2. Serialize (that serializes the data to memory stream that it then returns)
3. Append & log (just like it is now)

Basically let's make this a 4-line method.

src/Infrastructure/LeanCode.AuditLogs/ChangedEntitiesExtractor.cs
```diff
@@ -0,0 +1,58 @@
...
+
+    public static IReadOnlyList<EntityData> Extract(DbContext dbContext)
+    {
+        return dbContext.ChangeTracker
+            .Entries()
+            .Where(e => e.State != EntityState.Unchanged && e.State != EntityState.Detached)
+            .Select(
+                e =>
+                    new EntityData(
+                        // TODO: FIXME, I fail with owned entities
```
jakubfijalkowski: Fixed, hopefully?

src/Infrastructure/LeanCode.AuditLogs/ChangedEntitiesExtractor.cs
```diff
@@ -0,0 +1,58 @@
...
+                        e.Metadata
+                            .FindPrimaryKey()!
+                            .Properties.Select(
+                                p =>
+                                    // This may lose some info comparing to JsonSerializer.Serialize , but we don't get
+                                    // values in unnecessary "". We accept this tradeoff
+                                    p.PropertyInfo?.GetMethod?.Invoke(e.Entity, null)?.ToString()
```
jakubfijalkowski: https://learn.microsoft.com/en-us/dotnet/api/system.reflection.propertyinfo.getvalue?view=net-7.0

src/Infrastructure/LeanCode.AuditLogs/README.md
```diff
@@ -0,0 +1,151 @@
+# LeanCode.AuditLogs
```
jakubfijalkowski: And now, the big question - should it be here, or should it be in the docs?

test/Infrastructure/LeanCode.AuditLogs.Tests/TestDbContext.cs
```diff
@@ -0,0 +1,139 @@
...
+
+    public IncludedEntity(TestEntity testEntity, int someInt, string someString)
+    {
+        TestEntity = testEntity;
+        TestEntityId = testEntity.Id;
+        SomeInt = someInt;
+        SomeString = someString;
+    }
+
+    public void Update(string someString)
```
jakubfijalkowski: Unnecessary method. Set the value directly, you won't los readability.

test/Infrastructure/LeanCode.AuditLogs.Tests/TestDbContext.cs
```diff
@@ -0,0 +1,139 @@
...
+
+    private OwnedEntity() { }
+
+    public OwnedEntity(int someInt, string someString)
+    {
+        SomeInt = someInt;
+        SomeString = someString;
+    }
+
+    public void Update(string someString)
```
jakubfijalkowski: Ditto.

test/Infrastructure/LeanCode.AuditLogs.Tests/TestDbContext.cs
```diff
@@ -0,0 +1,139 @@
...
+    public void UpdateIncludedEntities()
+    {
+        foreach (var e in includedEntities)
+        {
+            e.Update((e.SomeInt + 100).ToString());
+        }
+    }
+}
+
+public class OwnedEntity
```
jakubfijalkowski, line 105: For the purpose of this task, this can be (even) a record.

test/Infrastructure/LeanCode.AuditLogs.Tests/LeanCode.AuditLogs.Tests.csproj
```diff
@@ -0,0 +1,24 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <Nullable>enable</Nullable>
+    <!-- Necessary for LeanCode.TimeProvider.TestHelpers to restore -->
```
jakubfijalkowski: Not really. They will restore happily but will _only_ leave warnings. Let's not introduce random `NoWarn`s.

test/Infrastructure/LeanCode.AuditLogs.Tests/ChangedEntititesExtractorTests.cs
```diff
@@ -0,0 +1,248 @@
...
+                    Type = typeof(TestEntity).FullName,
+                    Changes = JsonSerializer.SerializeToDocument(testEntity, Options),
+                    EntityState = "Deleted",
+                },
+                opt => opt.ComparingByMembers<JsonElement>()
+            );
+    }
+
+    [Fact]
+    public void Check_if_finds_multiple_changed_entities()
```
jakubfijalkowski, line 108: Can we split this test into more smaller tests? These are basically unit tests so there is no need to keep it in a single test case.

test/Infrastructure/LeanCode.AuditLogs.Tests/ChangedEntititesExtractorTests.cs
```diff
@@ -0,0 +1,248 @@
...
+
+    [Fact]
+    public void Check_if_updated_entity_is_extracted()
+    {
+        dbContext.TestEntities.Add(TestEntity.Create(SomeId));
+
+        dbContext.SaveChanges();
+
+        var testEntity = dbContext.TestEntities.Find(SomeId);
+        testEntity.Should().NotBeNull();
```
jakubfijalkowski: What does this test?

test/Infrastructure/LeanCode.AuditLogs.Tests/AzureBlobAuditLogStorageTests.cs
```diff
@@ -0,0 +1,209 @@
...
+using Azure.Data.Tables;
+using Azure.Storage.Blobs;
+using Azure.Storage.Blobs.Models;
+using FluentAssertions;
+using LeanCode.TimeProvider;
+using Xunit;
+
+namespace LeanCode.AuditLogs.Tests;
+
+public class AzureBlobAuditLogStorageTests
```
jakubfijalkowski: Can we refactor this and make the test more readable? Let's wrap what is "not important" into named methods - it will be much easier to read.

test/Infrastructure/LeanCode.AuditLogs.Tests/AuditLogsFilterTests.cs
```diff
@@ -0,0 +1,175 @@
...
+    public class DefaultConsumerDefinition<TConsumer> : ConsumerDefinition<TConsumer>
+        where TConsumer : class, IConsumer
+    {
+        protected override void ConfigureConsumer(
+            IReceiveEndpointConfigurator endpointConfigurator,
+            IConsumerConfigurator<TConsumer> consumerConfigurator,
+            IRegistrationContext context
+        )
+        {
+            endpointConfigurator.UseMessageRetry(
```
jakubfijalkowski: Why we need it in tests at all?

test/Infrastructure/LeanCode.AuditLogs.Tests/AuditLogsFilterTests.cs
```diff
@@ -0,0 +1,175 @@
...
+            .Select<AuditLogMessage>()
+            .Should()
+            .ContainSingle()
+            .Which.Context.Message.Should()
+            .BeEquivalentTo(
+                new
+                {
+                    ActionName = typeof(Consumer).FullName,
+                    ActorId,
+                    DateOccurred = Time.NowWithOffset,
```
jakubfijalkowski:
```suggestion
                    ActorId,
```

The test suggests testing only `ActorId` so let's test only that (everything else will be tested in a different place).

test/Infrastructure/LeanCode.AuditLogs.Tests/AuditLogsFilterTests.cs
```diff
@@ -0,0 +1,175 @@
...
+                        Changes = JsonSerializer.SerializeToDocument(TestEntity, Options),
+                    },
+                    ActionName = typeof(Consumer).FullName,
+                    ActorId = null as string,
+                    DateOccurred = Time.NowWithOffset,
+                },
+                opt => opt.ComparingByMembers<JsonElement>()
+            )
+            .And.Subject.Should()
+            .Match((s) => s.As<AuditLogMessage>().SpanId != null && s.As<AuditLogMessage>().TraceId != null);
```
jakubfijalkowski:
```suggestion
            .Match(s => s.As<AuditLogMessage>().SpanId is not null && s.As<AuditLogMessage>().TraceId is not null);
```

Although i'm not sure about the `is not` change (might be related to this being an `Expression`, which I don't know if it is).

test/Infrastructure/LeanCode.AuditLogs.Tests/AuditLogsFilterTests.cs
```diff
@@ -0,0 +1,175 @@
...
+            .Select<AuditLogMessage>()
+            .Should()
+            .ContainSingle()
+            .Which.Context.Message.Should()
+            .BeEquivalentTo(
+                new
+                {
+                    EntityChanged = new
+                    {
+                        Ids = new string[] { SomeId },
```
jakubfijalkowski:
```suggestion
                        Ids = new[] { SomeId },
```

test/Infrastructure/LeanCode.AuditLogs.Tests/AuditLogsFilterTests.cs
```diff
@@ -0,0 +1,175 @@
...
+                        Type = typeof(TestEntity).FullName,
+                        EntityState = "Added",
+                        Changes = JsonSerializer.SerializeToDocument(TestEntity, Options),
+                    },
+                    ActionName = typeof(Consumer).FullName,
+                    ActorId = null as string,
```
jakubfijalkowski: `(string)null!`

test/Infrastructure/LeanCode.AuditLogs.Tests/AuditLogsFilterTests.cs
```diff
@@ -0,0 +1,175 @@
...
+        collection.AddTransient<IAuditLogStorage, StubAuditLogStorage>();
+        collection.AddMassTransitTestHarness(ConfigureMassTransit);
+
+        serviceProvider = collection.BuildServiceProvider();
+        harness = serviceProvider.GetRequiredService<ITestHarness>();
+    }
+
+    private static void ConfigureMassTransit(IBusRegistrationConfigurator cfg)
+    {
+        cfg.AddConsumersWithDefaultConfiguration(
```
jakubfijalkowski: Shouldn't we have an extension method for this one?

test/Infrastructure/LeanCode.AuditLogs.Tests/AuditLogsFilterTests.cs
```diff
@@ -0,0 +1,175 @@
...
+    private readonly ITestHarness harness;
+    private static readonly TestEntity TestEntity = new() { Id = SomeId };
+
+    public AuditLogsFilterTests()
+    {
+        TestTimeProvider.ActivateFake(new DateTimeOffset(2023, 10, 6, 11, 0, 3, 0, TimeSpan.Zero));
+        var collection = new ServiceCollection();
+        collection
+            .AddOpenTelemetry()
+            .ConfigureResource(r => r.AddService("AuditLogsFilterTests", serviceInstanceId: Environment.MachineName))
```
jakubfijalkowski:
```suggestion
            .ConfigureResource(r => r.AddService("AuditLogsFilterTests"))
```

Plus, TBH, I'm not sure if `ConfigureResource` call is even needed.

test/Infrastructure/LeanCode.AuditLogs.Tests/AuditLogsFilterTests.cs
```diff
@@ -0,0 +1,175 @@
...
+            .WithTracing(builder =>
+            {
+                builder
+                    .AddProcessor<IdentityTraceAttributesFromBaggageProcessor>()
+                    .AddSource("MassTransit")
+                    .AddLeanCodeTelemetry();
```
jakubfijalkowski: Is this needed?

test/Infrastructure/LeanCode.AuditLogs.Tests/AuditLogsFilterTests.cs
```diff
@@ -0,0 +1,175 @@
...
+    {
+        TestTimeProvider.ActivateFake(new DateTimeOffset(2023, 10, 6, 11, 0, 3, 0, TimeSpan.Zero));
+        var collection = new ServiceCollection();
+        collection
+            .AddOpenTelemetry()
+            .ConfigureResource(r => r.AddService("AuditLogsFilterTests", serviceInstanceId: Environment.MachineName))
+            .WithTracing(builder =>
+            {
+                builder
+                    .AddProcessor<IdentityTraceAttributesFromBaggageProcessor>()
```
jakubfijalkowski: And this one?

test/Infrastructure/LeanCode.AuditLogs.Tests/AuditLogsFilterTests.cs
```diff
@@ -0,0 +1,175 @@
...
+                busCfg.ConfigureEndpoints(ctx, new DefaultEndpointNameFormatter("InMemory"));
+                busCfg.ConnectBusObservers(ctx);
+            }
+        );
+    }
+
+    [Fact]
+    public async Task Publishes_entity_changes_to_bus()
+    {
+        await harness.Start();
```
jakubfijalkowski: Let's move this call to the `StartAsync` of `IAsyncLifetime`.

jakubfijalkowski: And probably the `consumerHarness` can be shared across tests (i.e. set it in ctor as a class field).

test/Infrastructure/LeanCode.AuditLogs.Tests/AuditLogsMiddlewareTests.cs
```diff
@@ -0,0 +1,204 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using OpenTelemetry.Resources;
+using OpenTelemetry.Trace;
+using Xunit;
+using static LeanCode.AuditLogs.Tests.AuditLogsFilterTests;
+
+namespace LeanCode.AuditLogs.Tests;
+
+public sealed class AuditLogsMiddlewareTests : IAsyncLifetime, IDisposable
```
jakubfijalkowski: I totally don't get why we need MassTransit here. We're testing `AuditLogsMiddleware` that is unrelated to MT - it only publishes messages (and we can assert that differently - we can have mocks for that).

I also don't think we need WebHost here at all. We don't really rely on HTTP in the middleware, so there is no point in putting everything just to test two lines of code.

jakubfijalkowski: Also, if you extract the "gather what changed and publish message" logic into a separate class, then test it independently of MT/WebHost, then testing both filter & middleware assuming that it works will be _much_ simpler.

jakubfijalkowski (changes requested):

src/Infrastructure/LeanCode.AuditLogs/ChangedEntitiesExtractor.cs
lukaszgarstecki: Well, I guess `fail` is not a right word here. The key name is not extracted for owned entities

src/Infrastructure/LeanCode.AuditLogs/README.md
lukaszgarstecki: I believe it should be in the docs, as it is something that describes how to configure a feature that this package provides. On the other hand this is strictly connected to this package and it's references, so here is a good place for it as well.
@wojtek2288 do you have any thoughts on this?

test/Infrastructure/LeanCode.AuditLogs.Tests/AuditLogsFilterTests.cs
lukaszgarstecki:
```suggestion
                        Ids = [SomeId],
```
:see_no_evil:

test/Infrastructure/LeanCode.AuditLogs.Tests/AuditLogsFilterTests.cs
lukaszgarstecki: is this some kind of preference?

test/Infrastructure/LeanCode.AuditLogs.Tests/AuditLogsFilterTests.cs
lukaszgarstecki: It is. I tried to use `is not null` here but unfortunately it is not supported :disappointed:

test/Infrastructure/LeanCode.AuditLogs.Tests/ChangedEntititesExtractorTests.cs
lukaszgarstecki: well, it tests `SaveChanges` which I guess is not what we are interested in within this test...

test/Infrastructure/LeanCode.AuditLogs.Tests/AuditLogsFilterTests.cs
jakubfijalkowski: There is a difference in the behavior - `as` is "safer" as it does the cast _if it can_ (otherwise it will return `null`), normal cast will just fail. It's safer to use normal casts if you assume you are sure about the source type (i.e. most of the time).

src/Infrastructure/LeanCode.AuditLogs/AuditLogsFilter.cs
```diff
@@ -0,0 +1,57 @@
...
+
+    public async Task Send(
+        ConsumerConsumeContext<TConsumer, TMessage> context,
+        IPipe<ConsumerConsumeContext<TConsumer, TMessage>> next
+    )
+    {
+        await next.Send(context);
+        await AuditLogsPublisher.ExtractAndPublishAsync(
+            dbContext,
+            bus,
```
jakubfijalkowski: I _think_ we should use `context` to publish messages.

src/Infrastructure/LeanCode.AuditLogs/AuditLogsPublisher.cs
```diff
@@ -0,0 +1,44 @@
+using System.Diagnostics;
+using LeanCode.OpenTelemetry;
+using LeanCode.TimeProvider;
+using MassTransit;
+using Microsoft.EntityFrameworkCore;
+
+namespace LeanCode.AuditLogs;
+
+public static class AuditLogsPublisher
```
jakubfijalkowski: Let's make this a proper, non-static class. Will be easier to unit-test the filter & middleware (and will follow OOP principles more).

src/Infrastructure/LeanCode.AuditLogs/AuditLogsPublisher.cs
```diff
@@ -0,0 +1,44 @@
...
+        CancellationToken cancellationToken
+    )
+    {
+        var entitiesChanged = ChangedEntitiesExtractor.Extract(dbContext);
+        if (entitiesChanged.Count != 0)
+        {
+            var actorId = Activity.Current?.GetBaggageItem(IdentityTraceBaggageHelpers.CurrentUserIdKey);
+            var now = Time.NowWithOffset;
+
+            return Task.WhenAll(
```
jakubfijalkowski: Let's await this task to persist the stacktrace.

src/Infrastructure/LeanCode.AuditLogs/AzureBlobAuditLogStorage.cs
```diff
@@ -0,0 +1,144 @@
...
+            if (e.ErrorCode == BlobErrorCode.BlockCountExceedsLimit)
+            {
+                await BumpSuffixInTableAsync(auditLogMessage, cancellationToken);
+                await AppendLogToBlobAsync(auditLogMessage, cancellationToken);
+            }
+            else
+            {
+                throw;
+            }
+        }
```
jakubfijalkowski:
```suggestion
        catch (RequestFailedException e) when (e.ErrorCode == BlobErrorCode.BlockCountExceedsLimit)
        {
            await BumpSuffixInTableAsync(auditLogMessage, cancellationToken);
            await AppendLogToBlobAsync(auditLogMessage, cancellationToken);
        }
```

src/Infrastructure/LeanCode.AuditLogs/ChangedEntitiesExtractor.cs
jakubfijalkowski: So, we still need to fix something before releasing, right?

lukaszgarstecki: Considering our goal here, I'd say no. We discussed it some time ago and we want to focus on the aggregates logging here (which works fine, they get logged with owned entities for which we cannot extract key). There is no meaningful log for those entities that are owned by some other entities, which I don't find being a problem, since they do not have a meaning without root entity

jakubfijalkowski: Okay, so let's get rid of the TODO then? :D

test/Infrastructure/LeanCode.AuditLogs.Tests/TestDbContext.cs
lukaszgarstecki: I'll leave it as a class, since I want to modify it without changing an entity reference

src/Infrastructure/LeanCode.AuditLogs/ChangedEntitiesExtractor.cs
lukaszgarstecki: Yeah, I'll replace it with more suitable comment

test/Infrastructure/LeanCode.AuditLogs.Tests/AuditLogsFilterTests.cs
lukaszgarstecki: Well, I'll need to stick to your version :/
image

test/Infrastructure/LeanCode.AuditLogs.Tests/ChangedEntititesExtractorTests.cs
lukaszgarstecki: Well, we do have those cases unit tested above. I guess we can simplify this test and finish assertions on 
`.Should().HaveCount(2)` ?

jakubfijalkowski: Yeah, it will be OK that way.

src/Infrastructure/LeanCode.AuditLogs/README.md
jakubfijalkowski: I also think this should be in docs - we don't yet have a sensible approach to READMEs, so let's keep everything in a single/correct place.

cloud-tests-infrastructure/storage.tf
```diff
@@ -0,0 +1,45 @@
...
+  role_definition_name = "Storage Blob Data Owner"
+  principal_id         = azuread_service_principal.tests.object_id
+}
+
+resource "azurerm_storage_table" "example" {
+  name                 = local.table_storage_table_name
+  storage_account_name = azurerm_storage_account.storage.name
+}
+
+locals {
```
jakubfijalkowski, line 42: Do we benefit from having these as locals?

src/Infrastructure/LeanCode.AuditLogs/AuditLogsPublisher.cs
```diff
@@ -0,0 +1,42 @@
...
+using LeanCode.OpenTelemetry;
+using LeanCode.TimeProvider;
+using MassTransit;
+using Microsoft.EntityFrameworkCore;
+
+namespace LeanCode.AuditLogs;
+
+public class AuditLogsPublisher
+{
+    public async Task ExtractAndPublishAsync(
```
jakubfijalkowski: `virtual`

src/Infrastructure/LeanCode.AuditLogs/AzureBlobAuditLogStorage.cs
```diff
@@ -0,0 +1,137 @@
...
+            cancellationToken: cancellationToken
+        );
+        if (!res.HasValue)
+        {
+            var entity = new TableEntity(
+                auditLogMessage.EntityChanged.Type,
+                string.Join("", auditLogMessage.EntityChanged.Ids)
+            )
+            {
+                ["Suffix"] = 0,
```
jakubfijalkowski: I think we should extract the `"Suffix"` string to a `const`.

src/Infrastructure/LeanCode.AuditLogs/StubAuditLogStorage.cs
```diff
@@ -0,0 +1,24 @@
+using Serilog;
+
+namespace LeanCode.AuditLogs;
+
+public class StubAuditLogStorage : IAuditLogStorage
```
jakubfijalkowski: I wonder if we should ever provide this implementation - it's somewhat useless for real projects.

test/Infrastructure/LeanCode.AuditLogs.Tests/AuditLogsFilterTests.cs
```diff
@@ -0,0 +1,37 @@
+using MassTransit;
+using NSubstitute;
+using Xunit;
+
+namespace LeanCode.AuditLogs.Tests;
+
+public sealed class AuditLogsFilterTests
+{
+    public const string ConsumerName = "Test.Consumer.Name";
```
jakubfijalkowski:
```suggestion
    private const string ConsumerName = "Test.Consumer.Name";
```

test/Infrastructure/LeanCode.AuditLogs.Tests/AuditLogsFilterTests.cs
```diff
@@ -0,0 +1,37 @@
...
+
+    [Fact]
+    public async void Extracts_changes_after_pipeline_execution()
+    {
+        var dbContext = new TestDbContext();
+        var publisher = Substitute.For<AuditLogsPublisher>();
```
jakubfijalkowski, line 15: This will result in the "real" logic of `AuditLogsPublisher` being run. IIRC NSubstitute won't by default stub any method of a class that isn't virtual, thus all the calls to `publisher.ExtractAndPublishAsync` will run (and you don't want that here).

Adding the missing `virtual` will fix the problem (and will probably fix the problem with randomly failing tests).

test/Infrastructure/LeanCode.AuditLogs.Tests/AuditLogsIntegrationTests.cs
```diff
@@ -0,0 +1,200 @@
...
+                webHost
+                    .UseTestServer()
+                    .ConfigureServices(cfg =>
+                    {
+                        cfg.AddOpenTelemetry()
+                            .WithTracing(builder =>
+                            {
+                                builder
+                                    .AddAspNetCoreInstrumentation(
+                                        opts => opts.Filter = ctx => !ctx.Request.Path.StartsWithSegments("/live")
```
jakubfijalkowski: This is unnecessary

test/Infrastructure/LeanCode.AuditLogs.Tests/AuditLogsIntegrationTests.cs
```diff
@@ -0,0 +1,200 @@
...
+                                    )
+                                    .AddProcessor<IdentityTraceAttributesFromBaggageProcessor>()
```
jakubfijalkowski: Is this necessary? Is the next line necessary?

test/Infrastructure/LeanCode.AuditLogs.Tests/AuditLogsIntegrationTests.cs
```diff
@@ -0,0 +1,200 @@
...
+                        cfg.AddDbContext<TestDbContext>();
+                        cfg.AddTransient<IAuditLogStorage, StubAuditLogStorage>();
+                        cfg.AddTransient<AuditLogsPublisher>();
+                        cfg.AddMassTransitTestHarness(ConfigureMassTransit);
+                        cfg.AddRouting();
+                    })
+                    .Configure(app =>
+                    {
+                        app.Audit<TestDbContext>()
+                            .UseIdentityTraceAttributes()
```
jakubfijalkowski: Does this add anything meaningful to the test?

test/Infrastructure/LeanCode.AuditLogs.Tests/AuditLogsMiddlewareTests.cs
```diff
@@ -0,0 +1,26 @@
...
+using Xunit;
+
+namespace LeanCode.AuditLogs.Tests;
+
+public sealed class AuditLogsMiddlewareTests
+{
+    [Fact]
+    public async void Extracts_changes_after_pipeline_execution()
+    {
+        var requestPath = "/test.request.path";
```
jakubfijalkowski:
```suggestion
        const string RequestPath = "/test.request.path";
```

test/Infrastructure/LeanCode.AuditLogs.Tests/AuditLogsMiddlewareTests.cs
```diff
@@ -0,0 +1,26 @@
...
+        var dbContext = new TestDbContext();
+        var bus = Substitute.For<IBus>();
+        var publisher = Substitute.For<AuditLogsPublisher>();
+        var requestDelegate = Substitute.For<RequestDelegate>();
+        var middleware = new AuditLogsMiddleware<TestDbContext>(requestDelegate);
```
jakubfijalkowski:
```suggestion
        var middleware = new AuditLogsMiddleware<TestDbContext>(c => Task.CompletedTask);
```

test/Infrastructure/LeanCode.AuditLogs.Tests/AuditLogsMiddlewareTests.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        var httpContext = Substitute.For<HttpContext>();
+        httpContext.Request.Path.Returns(PathString.FromUriComponent(requestPath));
+        await middleware.InvokeAsync(httpContext, dbContext, bus, publisher);
```
jakubfijalkowski:
```suggestion
        var httpContext = Substitute.For<HttpContext>();
        httpContext.Request.Path.Returns(PathString.FromUriComponent(requestPath));
        
        await middleware.InvokeAsync(httpContext, dbContext, bus, publisher);
```

Setting `HttpContext` is part of the arrange.

test/Infrastructure/LeanCode.AuditLogs.Tests/LeanCode.AuditLogs.Tests.csproj
```diff
@@ -0,0 +1,21 @@
...
+    <ProjectReference Include="../../../src/Testing/LeanCode.TimeProvider.TestHelpers/LeanCode.TimeProvider.TestHelpers.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+
+    <PackageReference Include="Azure.Identity" />
+    <PackageReference Include="MassTransit" />
+    <PackageReference Include="Microsoft.AspNetCore.TestHost" />
+    <PackageReference Include="OpenTelemetry.Extensions.Hosting" Version="1.5.1" />
+    <PackageReference Include="OpenTelemetry.Instrumentation.AspNetCore" Version="1.5.1-beta.1" />
```
jakubfijalkowski:
```suggestion
    <PackageReference Include="OpenTelemetry.Extensions.Hosting" />
    <PackageReference Include="OpenTelemetry.Instrumentation.AspNetCore" />
```

test/Infrastructure/LeanCode.AuditLogs.Tests/LeanCode.AuditLogs.Tests.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
```
jakubfijalkowski, line 3:
```suggestion

  <PropertyGroup>
```

test/Infrastructure/LeanCode.Azure.Tests/AzureStorageAuditLogIntegrationTests.cs
```diff
@@ -0,0 +1,181 @@
...
+        await foreach (var b in containerClient.GetBlobsAsync(BlobTraits.All, BlobStates.All, $"{type}/{id}"))
+        {
+            var blockBlob = containerClient.GetBlobClient(b.Name);
+            using var stream = new MemoryStream();
+            await blockBlob.DownloadToAsync(stream);
+
+            stream.Position = 0;
+            using var reader = new StreamReader(stream, Encoding.UTF8);
+
+            linesFound += (await reader.ReadToEndAsync()).Count(c => c == '\n');
```
jakubfijalkowski:
```suggestion
            linesFound += stream.ToArray().Count(c => c == '\n');
```

src/Infrastructure/LeanCode.AuditLogs/StubAuditLogStorage.cs
lukaszgarstecki: I was thinking about this, I believe this implementation can be beneficial when used in integration tests (since it doesn't require any external services to be configured)

src/Infrastructure/LeanCode.AuditLogs/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,25 @@
+using Microsoft.Extensions.DependencyInjection;
+
+namespace LeanCode.AuditLogs;
+
+public static class ServiceCollectionExtensions
+{
+    public static IServiceCollection AddAuditLogs<TAuditLogStorage>(this IServiceCollection services)
```
jakubfijalkowski: `AddAuditLogs<AzureBlobAuditLogStorage>` won't work. We should prevent that on API level (i.e. remove this one).

src/Infrastructure/LeanCode.AuditLogs/MassTransintRegistrationConfigurationExtensions.cs
```diff
@@ -0,0 +1,11 @@
+using MassTransit;
+
+namespace LeanCode.AuditLogs;
+
+public static class MassTransitRegistrationConfigurationExtensions
+{
+    public static void AddAuditLogsConsumer(this IRegistrationConfigurator configurator)
+    {
+        configurator.AddConsumer(typeof(AuditLogsConsumer));
```
jakubfijalkowski: What about consumer configuration?

lukaszgarstecki: I was wondering about this. I believe there is no need for custom configuration (except retries?). We don't need any extra actions here

cloud-tests-infrastructure/storage.tf
lukaszgarstecki: This is something that we need to configure on our own, that is why I wanted to separate those values from the resource definitions. 

That said, I wonder if we should provide those values using variables :thinking:

src/Infrastructure/LeanCode.AuditLogs/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,17 @@
+using Microsoft.Extensions.DependencyInjection;
+
+namespace LeanCode.AuditLogs;
```
jakubfijalkowski: I forgot about that earlier, but we should probably go with [this approach](https://github.com/leancodepl/corelibrary/pull/596#discussion_r1363841717).

src/Infrastructure/LeanCode.AuditLogs/MassTransintRegistrationConfigurationExtensions.cs
```diff
@@ -0,0 +1,25 @@
+using MassTransit;
+
+namespace LeanCode.AuditLogs;
+
+public static class MassTransitRegistrationConfigurationExtensions
+{
+    public static void AddAuditLogsConsumer(this IRegistrationConfigurator configurator)
```
jakubfijalkowski, line 7: Let's add an override with generic parameter that allows to specify the consumer definition.

test/Infrastructure/LeanCode.AuditLogs.Tests/AuditLogsMiddlewareTests.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace LeanCode.AuditLogs.Tests;
+
+public sealed class AuditLogsMiddlewareTests
+{
+    [Fact]
+    public async void Extracts_changes_after_pipeline_execution()
+    {
+        const string RequestPath = "/test.request.path";
+        var dbContext = new TestDbContext();
```
jakubfijalkowski:
```suggestion
        using var dbContext = new TestDbContext();
```

jakubfijalkowski (approved):
Final minor comments, after that - :shipit: .

jakubfijalkowski (approved):
Can we also fix the IDE/CA warnings in tests sometime in the future?

PR #593, opened 2023-10-16
Add PN token store method to remove all tokens for given user

Saancreed:
This will be a breaking change unless we provide a default implementation, but I'm not sure if we should.


jakubfijalkowski:
Considering that `IPushNotificationTokenStore` is basically an internal interface, I wouldn't consider that a breaking change.

Nevertheless, we'll be changing major version soon, so we're OK with that.

jakubfijalkowski (approved):

PR #596, opened 2023-10-18
Add LeanCode.ConfigCat package

Saancreed:
The only changes from previous iteration concern the logger adapter: messages are now always formatted using `InvariantCulture` (let me know if this isn't what we want) and `ArgumentOutOfRangeException` is now thrown instead of `IndexOutOfRangeException` which is apparently runtime-reserved type.

### DIFF
--- a/src/Infrastructure/LeanCode.ConfigCat/ConfigCatExtensions.cs
+++ b/src/Infrastructure/LeanCode.ConfigCat/ConfigCatExtensions.cs
@@ -1,0 +1,128 @@
+using System.Text.Json;
+using ConfigCat.Client;
+using LeanCode.ConfigCat;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+[assembly: System.Diagnostics.CodeAnalysis.SuppressMessage(
+    "?",
+    "IDE0130:NamespaceDoesNotMatchFolderStructure",
+    Justification = "Extensions on IServiceCollection by convention live in Microsoft.Extensions.DependencyInjection namespace.",
+    Scope = "namespace",
+    Target = "~N:Microsoft.Extensions.DependencyInjection"
+)]
+
+namespace Microsoft.Extensions.DependencyInjection;
+
+public static class ConfigCatExtensions
+{
+    public static IServiceCollection AddConfigCat(
+        this IServiceCollection services,
+        string? sdkKey,
+        string? flagOverridesFilePath,
+        bool autoReload = true,
+        OverrideBehaviour overrideBehaviour = OverrideBehaviour.LocalOnly,
+        bool initializeOnStartup = false
+    )
+    {
+        return services.AddConfigCat(
+            sdkKey,
+            flagOverridesFilePath is null
+                ? null
+                : FlagOverrides.LocalFile(flagOverridesFilePath, autoReload, overrideBehaviour),
+            initializeOnStartup
+        );
+    }
+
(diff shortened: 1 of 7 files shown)


src/Infrastructure/LeanCode.ConfigCat/ConfigCatExtensions.cs
```diff
@@ -0,0 +1,128 @@
+using System.Text.Json;
+using ConfigCat.Client;
+using LeanCode.ConfigCat;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+[assembly: System.Diagnostics.CodeAnalysis.SuppressMessage(
```
jakubfijalkowski, line 8: I wonder if we should move (not now, but sometime soon) all the extensions to `IServiceCollection` to this namespace.

src/Infrastructure/LeanCode.ConfigCat/ConfigCatInitializer.cs
```diff
@@ -0,0 +1,28 @@
...
+        "CA1031:DoNotCatchGeneralExceptionTypes",
+        Justification = "We don't want any exceptions to be propagated."
+    )]
+    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+        try
+        {
+            await configCatClient.GetAllKeysAsync(stoppingToken);
+        }
+        catch { }
```
jakubfijalkowski, line 26: Logged, maybe?

jakubfijalkowski (approved):

src/Infrastructure/LeanCode.ConfigCat/ConfigCatExtensions.cs
Saancreed: I believe we should :slightly_smiling_face:

src/Infrastructure/LeanCode.ConfigCat/ConfigCatInitializer.cs
Saancreed: I'd expect the ConfigCat's client to take care of logging any failures here but sure, I can add something myself.

jakubfijalkowski: Nono, if it is logged already, we don't need it to log again.

Saancreed: Actually, hmm, looks like [all the exceptions except `OperationCanceledException` are swallowed anyway](https://github.com/configcat/.net-sdk/blob/57f08f27508e6f98c6e4db5ca305dfa4ade36b0f/src/ConfigCatClient/ConfigCatClient.cs#L395-L416). But yes, they will be logged through the adapter anyway.

marcin-dardzinski: So, do we even need it or not? Last time I though we're dumping it.

Saancreed: It will stay as an opt-in. Defaults provided in DI extensions don't register this but if you want to warm up the cache and be told earlier that something's wrong then you can use it.

Saancreed: Oh, and the flavor based on `BackgroundService` shouldn't block the host from starting even if ConfigCat takes a long time to get ready.

PR #602, opened 2023-10-23
Fixup semantic logging in ConfigCat's logger adapter

Saancreed:


src/Infrastructure/LeanCode.ConfigCat/ConfigCatToMSLoggerAdapter.cs
```diff
@@ -88,5 +91,20 @@ public LogValues(ref FormattableLogMessage message)
...
 
         public override string ToString() => Message.InvariantFormattedMessage;
+
+        private string RenderTemplate()
+        {
+            // Message.Format contains template in a format: 'A sample error message with parameter value: {0}'
+            // Serilog does not treat it as valid argument, and logs this as is.
+            // So instead we render the format to get: 'A sample error message with parameter value: {Value}'
+            // Which can be consumed correctly later
+            return Message.ArgNames is not { } argNames
```
marcin-dardzinski: Client is nullable aware, so I think we can spare this null check, also the `is not` form seems extra confusing for me.

src/Infrastructure/LeanCode.ConfigCat/ConfigCatToMSLoggerAdapter.cs
```diff
@@ -70,8 +73,8 @@ public LogValues(ref FormattableLogMessage message)
                 ArgumentOutOfRangeException.ThrowIfGreaterThanOrEqual(index, Count);
 
                 return index == Count - 1
-                    ? new KeyValuePair<string, object?>("{OriginalFormat}", Message.Format)
-                    : new KeyValuePair<string, object?>(Message.ArgNames![index], Message.ArgValues![index]);
+                    ? new("{OriginalFormat}", templateFormat ??= RenderTemplate())
+                    : new(Message.ArgNames![index], Message.ArgValues![index]);
```
marcin-dardzinski: Nit: since you're touching it, `!`s are not necessary

marcin-dardzinski (approved):

src/Infrastructure/LeanCode.ConfigCat/ConfigCatToMSLoggerAdapter.cs
Saancreed: `Message` is a `struct` so whenever someone does `= default;` it immediately breaks nullables :disappointed:

I'd prefer to keep this because I'm a bit too paranoid but I could invert the check to hopefully make it more readable.

PR #603, opened 2023-10-25
Add cancellation token naming analyzer + code fix

wojtek2288:


src/Infrastructure/LeanCode.PeriodicService/IPeriodicAction.cs
```diff
@@ -8,5 +8,5 @@ public interface IPeriodicAction
 {
     CronExpression When { get; }
     bool SkipFirstExecution { get; }
-    Task ExecuteAsync(CancellationToken stoppingToken);
+    Task ExecuteAsync(CancellationToken cancellationToken);
```
lukaszgarstecki: stoppingToken here was correct

src/Infrastructure/LeanCode.PeriodicService/PeriodicHostedService.cs
```diff
@@ -34,6 +34,7 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
         "CA1031",
         Justification = "The method is an exception boundary."
     )]
+#pragma warning disable LNCD0006 // For consistency with `ExecuteAsync`
```
lukaszgarstecki: can we use `System.Diagnostics.CodeAnalysis.SuppressMessage` here?

src/Tools/LeanCode.CodeAnalysis/Analyzers/CancellationTokensShouldFollowNamingConvention.cs
```diff
@@ -0,0 +1,66 @@
...
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace LeanCode.CodeAnalysis.Analyzers;
+
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public class CancellationTokensShouldFollowNamingConvention : DiagnosticAnalyzer
+{
+    internal const string IdentifierName = "CancellationToken";
+    internal const string ExpectedName = "cancellationToken";
```
lukaszgarstecki:
```suggestion
    internal const string TypeName = "CancellationToken";
    internal const string ParameterName = "cancellationToken";
```
What do you think? Are the names you used relate to CodeAnalysis API?

src/Infrastructure/LeanCode.PeriodicService/IPeriodicAction.cs
wojtek2288: Hm here I'm not so sure, do we stick with `BackgroundService`s `stoppingToken` or with ours `cancellationToken`? I'm leaning towards renaming this to `cancellationToken` as this will be used in our apps and will be more consistent with our naming.

jakubfijalkowski: This `CancellationToken` is much different than "normal" `CancellationToken`. It does not represent cancelling the operation - it notifies you about stopping the app. We should keep the original name.

Also, this is not the only place where `CancellationToken` does not represent `cancellation`.

jakubfijalkowski (approved):

lukaszgarstecki (approved):

PR #604, opened 2023-10-26
Collect CQRS metrics

marcin-dardzinski:
I went with success and tagged failure metrics, but I can convert them to separate metric per failure. Didn't think much of the naming, any comments welcome.  
I didn't do any tests, I don't really see a sane way to do it, apart from connecting to some app.

Plus, `"{@Object} executed successfully"` log message after each command was somehow dropped when rewriting cqrs, bringing it back now.

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -102,6 +102,7 @@
     <PackageReference Update="Microsoft.Extensions.Options" Version="$(ExtensionsVersion)" />
     <PackageReference Update="Microsoft.Extensions.Options.ConfigurationExtensions" Version="$(ExtensionsVersion)" />
 
+    <PackageReference Update="Microsoft.Extensions.Telemetry.Testing" Version="8.0.0-rc.2.23510.2" />
     <PackageReference Update="Microsoft.Extensions.TimeProvider.Testing" Version="8.0.0-rc.2.23510.2" />
 
     <PackageReference Update="Microsoft.Extensions.WebEncoders" Version="$(AspNetCoreVersion)" />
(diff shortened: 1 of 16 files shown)


jakubfijalkowski:
You can wire-up some fake metrics listener to test this, I think.

jakubfijalkowski (commented):
Also, I wonder, maybe we should switch to [DI-based metrics](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/metrics-instrumentation#get-a-meter-via-dependency-injection) ?

marcin-dardzinski:
> Also, I wonder, maybe we should switch to [DI-based metrics](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/metrics-instrumentation#get-a-meter-via-dependency-injection) ?

Sounds good, seems to be easier to test, as what I tried is a nightmare. Bonus side effect - DI will throw if you don't add any meter :see_no_evil: 

On the other hand, I think it's safe to add `IServiceProvider.AddMetrics()` to `AddCQRS` registration method.

jakubfijalkowski:
> Bonus side effect - DI will throw if you don't add any meter 🙈

> Starting in .NET 8, hosts will automatically register [IMeterFactory](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics.imeterfactory) in the service container

:)

marcin-dardzinski:
Switched to DI, rest of the tests will have to be re-written

Dragemil (approved):

jakubfijalkowski (approved):

PR #608, opened 2023-11-08
CQRS handlers name analyzer + code action

wojtek2288:


src/Tools/LeanCode.CodeAnalysis/CodeActions/FixCQRSHandlerNamingCodeAction.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        if (token.IsKind(SyntaxKind.IdentifierToken) && token.Parent is ClassDeclarationSyntax classDeclaration)
+        {
+            var expectedHandlerName = GetExpectedCQRSHandlerName(classDeclaration, model);
+            var classSymbol = model.GetDeclaredSymbol(classDeclaration, cancellationToken)!;
+
+            return await Renamer.RenameSymbolAsync(
+                solution,
+                classSymbol,
+                new(RenameFile: true),
```
wojtek2288, line 46: thank god (or roslyn) for this option

src/Tools/LeanCode.CodeAnalysis/Analyzers/EnsureCQRSHandlersFollowNamingConvention.cs
```diff
@@ -0,0 +1,184 @@
...
+        int implementationCount
+    )
+    {
+        var expectedSuffix = GetSuffix(handlerType);
+
+        if (implementationCount > 1)
+        {
+            return type.Name.EndsWith(expectedSuffix, StringComparison.InvariantCulture)
+                ? type.Name
+                : type.Name + expectedSuffix;
```
Dragemil, line 96: What happens here? It is when there is one handler for a few CQRS objects? How would this blow up if one would have the same handler for a few different CQRS objects (like the same handler for a command and a query) - just a thought experiment.

Dragemil (commented):
Will command validators also be handled in a similar way?

src/Tools/LeanCode.CodeAnalysis/Analyzers/EnsureCQRSHandlersFollowNamingConvention.cs
wojtek2288: Yes it's for a case when handler implements for example multiple commands. If we mix up commands, queries, etc no diagnostic will be raised as `var handlerType = GetCQRSHandlerType(type, out var implementationCount);` will return null and analyzer will return without reporting anything.

wojtek2288:
@Dragemil We did not discuss adding similar analyzer for command validators but now when you mention it I think it can be added to the backlog. Btw there is already quick action which will add command validator, but in our case we often copy code from other commands so rename might be also useful. cc: @lukaszgarstecki

lukaszgarstecki:
I'd say a rename + quickfix can also be helpful for CV. The current analyzer tackles only the case when it's missing, I believe having similar validator as for CH, QH and OH would be nice. Will that be a lot of work @wojtek2288 ?

lukaszgarstecki (commented):
LGTM. I'd say that we can add CV rule and fix in this PR if this is not too much trouble

wojtek2288:
Check formatting step needs to be investigated here as when `csharpier` released `26.0` version CI check formatting step started failing (with generic error: `Error: Process completed with exit code 1.`) when `csharpier` version on github actions was different then new released one (maybe due to newest version being installed globally somehow?).

lukaszgarstecki (approved):

wojtek2288:
Need to wait for https://github.com/leancodepl/corelibrary/pull/618. As bumping dotnet to 8.0.100 fixes this issue: https://github.com/dotnet/sdk/issues/36434

PR #610, opened 2023-11-10
Enable registering specific handlers

michal12334:


lukaszgarstecki (approved):

wojtek2288:
Hm, CI checks did not run. Please check if you are facing: https://github.com/leancodepl/corelibrary/pull/608#issuecomment-1805425593 (check formatting step started failing after `csharpier` 26.0 release) on CI, as it might fail the build.

src/CQRS/LeanCode.CQRS.AspNetCore/ServiceCollectionCQRSExtensions.cs
```diff
@@ -31,6 +32,24 @@ TypesCatalog handlersCatalog
 
         return new CQRSServicesBuilder(serviceCollection, objectsSource);
     }
+
+    public static CQRSServicesBuilder AddCQRS(
```
jakubfijalkowski: Let's make the first `AddCQRS` use this one underneath.

src/CQRS/LeanCode.CQRS.AspNetCore/ServiceCollectionCQRSExtensions.cs
```diff
@@ -55,4 +74,10 @@ public CQRSServicesBuilder AddCQRSObjects(TypesCatalog contractsCatalog, TypesCa
         objectsSource.AddCQRSObjects(contractsCatalog, handlersCatalog);
         return this;
     }
+
+    public CQRSServicesBuilder AddCQRSObjects(IEnumerable<TypeInfo> possibleContracts, IEnumerable<TypeInfo> possibleHandlers)
```
Dragemil: What does `possible` prefix means here and in the other place?

src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSObjectsRegistrationSource.cs
```diff
@@ -18,16 +18,12 @@ public CQRSObjectsRegistrationSource(IServiceCollection services)
         this.services = services;
     }
 
-    public void AddCQRSObjects(TypesCatalog contractsCatalog, TypesCatalog handlersCatalog)
+    public void AddCQRSObjects(IEnumerable<TypeInfo> possibleContracts, IEnumerable<TypeInfo> possibleHandlers)
```
jakubfijalkowski: I'm against this change - if we want to allow registering limited number of types, then the `TypesCatalog` should change, not every application of it.

jakubfijalkowski (changes requested):

src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSObjectsRegistrationSource.cs
marcin-dardzinski: My take here is given the limited applicability of manual registration, it's better to have explicit
`AddCommand<TCommand,THandler>()` etc. methods.

By the way, the goal of this change was to facilitate open generic handlers (with user id type specified by the application), doing it via `TypesCatalog` would be awkward, as you'd need to find types you'd need to close (thus explicit manual registration).

jakubfijalkowski: You can't have open generics here - you won't have enough data to close them. Handlers must, in the end, be closed from the get go to be instantiatable by simple `GetService<ICommandHandler<XYZ>>()` calls - the call doesn't have the necessary types specified to close anything.

In this particular case, you _know_ what type the handler must close on, because it is constant for the project, i.e. the handler is `RegisterTokenCH<TUserId>`, but you will register only `RegisterTokenCH<Guid>` or `RegisterTokenCH<string>`.

You will basically end up with `new TypesCatalog().Add<RegisterTokenCH<Guid>>()`. It won't be that different from `AddCommand` that you propose, and TBH I'm indifferent here. `AddCommand`/`AddQuery` is also OK for me. I just don't want to have an option to add a set of queries/commands by passing list.

src/CQRS/LeanCode.CQRS.AspNetCore/ServiceCollectionCQRSExtensions.cs
michal12334: It mean that if someone passes type that cannot be registered as CQRS object for any reason.

But now it's outdated anyway.

src/CQRS/LeanCode.CQRS.AspNetCore/ServiceCollectionCQRSExtensions.cs
```diff
@@ -55,4 +57,32 @@ public CQRSServicesBuilder AddCQRSObjects(TypesCatalog contractsCatalog, TypesCa
         objectsSource.AddCQRSObjects(contractsCatalog, handlersCatalog);
         return this;
     }
+
+    public CQRSServicesBuilder AddQuery<TQuery, TResult, THandler>()
```
jakubfijalkowski, line 61: Tests pls

jakubfijalkowski (approved):

PR #618, opened 2023-11-16
Switch to .net8

marcin-dardzinski:
+ address some low hanging warnings


global.json
```diff
@@ -1,7 +1,7 @@
 {
   "sdk": {
-    "version": "8.0.0-rc.2.23479.6",
-    "allowPrerelease": true,
+    "version": "8.0.0",
```
Saancreed: That's incorrect, should be `8.0.100`.

`8.0.0` is the runtime version.

marcin-dardzinski: Huh, was it incorrect all the time?

Saancreed (approved):

global.json
Saancreed: Technically yes but 100 > 0 so it worked anyway :shrug:

Dragemil (approved):
Also there's MT 8.1.2 now.

marcin-dardzinski:
> Also there's MT 8.1.2 now.

It introduced some breaking change + we still need to fix _non low hanging_ MT warnings, let's do it separately.

PR #619, opened 2023-11-16
Fix csharpier check formatting step

wojtek2288:
Well this is ugly.


wojtek2288:
Actually this might not be necessary now as in PR with net8.0 this passes so maybe .net8.0 fixes it. Will wait for merge and check it then.

wojtek2288:
Actually this seems to be fixed in 8.0.100 (PR with bump passes). Will wait for merge. https://github.com/dotnet/sdk/issues/36434

PR #625, opened 2023-11-27
Fix code coverage calculation [resolves COR-169]

Dragemil:
Globbing should work?

### DIFF
--- a/test/LeanCode.TestCoverage.Tests/TestCoverageTests.cs
+++ b/test/LeanCode.TestCoverage.Tests/TestCoverageTests.cs
@@ -1,0 +1,9 @@
+using Xunit;
+
+namespace LeanCode.TestCoverage.Tests;
+
+public class TestCoverageTests
+{
+    [Fact]
+    public void Include_all_source_projects_for_test_coverage() { }
+}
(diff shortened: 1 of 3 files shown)


Dragemil:
Seems like it does, a ~4% coverage drop is noticed.

test/LeanCode.TestCoverage.Tests/TestCoverageTests.cs
```diff
@@ -0,0 +1,9 @@
+using Xunit;
+
+namespace LeanCode.TestCoverage.Tests;
+
+public class TestCoverageTests
+{
+    [Fact]
+    public void Include_all_source_projects_for_test_coverage() { }
```
jakubfijalkowski, line 8: Can we check that all the projects are really included in some easy way?

jakubfijalkowski (approved):

test/LeanCode.TestCoverage.Tests/TestCoverageTests.cs
Dragemil: Well, I can [fetch referenced assemblies](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.getreferencedassemblies?view=net-8.0), but how can I get expected projects to be included? One way is surely some file system hacks like globbing all `csproj`s in the `src` dir, but that I wouldn't call easy, very hacky rather.

PR #627, opened 2023-12-05
[COR-166] Make `IUserIdExtractor` common class (and implementation)

wojtek2288:


src/CQRS/LeanCode.CQRS.Security/UserIdExtractor.cs
jakubfijalkowski, line 1: I don't think we should put it here. Packages that use it don't need anything else apart from this one.

src/CQRS/LeanCode.CQRS.Security/UserIdExtractor.cs
```diff
@@ -0,0 +1,30 @@
+using System.Security.Claims;
+
+namespace LeanCode.CQRS.Security;
+
+public interface IUserIdExtractor<TUserId> : IUserIdExtractor
+    where TUserId : notnull, IEquatable<TUserId>
+{
+    new TUserId Extract(ClaimsPrincipal user);
```
jakubfijalkowski: Let's not use shadowing. These are both valid methods and both should be exposed.

test/CQRS/LeanCode.CQRS.Security.Tests/ClaimsPrinicipalTests.cs
```diff
@@ -12,6 +12,11 @@ namespace LeanCode.CQRS.Default.Tests.Security;
 
 public class ClaimsPrincipalTests
```
jakubfijalkowski: Why the tests ended up here?

src/CQRS/LeanCode.CQRS.Security/UserIdExtractor.cs
```diff
@@ -0,0 +1,30 @@
...
+}
+
+public interface IUserIdExtractor
+{
+    string Extract(ClaimsPrincipal user);
+}
+
+public class UserIdExtractor : IUserIdExtractor<Guid>
```
jakubfijalkowski: The name is too generic compared to the implementation. This handles `Guid`s only, and we use also `string`s, and typed IDs.

We should have a ready-made implementation for all three types of user IDs that we see in our projects (so - `Guid`, `string` and source-generated typed ID, which will be tricky). Plus we should have an extension method that registers it in the DI container.

jakubfijalkowski (changes requested):

src/Helpers/LeanCode.UserIdExtractors/Extractors/GuidUserIdExtractor.cs
```diff
@@ -0,0 +1,14 @@
+using System.Security.Claims;
+
+namespace LeanCode.UserIdExtractors.Extractors;
+
+internal class GuidUserIdExtractor : StringUserIdExtractor, IUserIdExtractor<Guid>
```
jakubfijalkowski: `public sealed` IMO (and others). We gain nothing from hiding it.

src/CQRS/LeanCode.CQRS.Security/UserIdExtractor.cs
wojtek2288: I thought it was a bit related that is why I put it here, moved it to separate project

src/Helpers/LeanCode.UserIdExtractors/Extractors/GuidUserIdExtractor.cs
jakubfijalkowski: Plus, I don't think we should inherit the `StirngUserIdExtractor`. There isn't really any connection between the two.

test/CQRS/LeanCode.CQRS.Security.Tests/ClaimsPrinicipalTests.cs
wojtek2288: It kinda was related to `ClaimsPrincipal` as we extract user id from it, anyway moved it to separate project.

src/Helpers/LeanCode.UserIdExtractors/Extractors/RawTypedUserIdExtractor.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Security.Claims;
+using LeanCode.DomainModels.Ids;
+
+namespace LeanCode.UserIdExtractors.Extractors;
+
+internal class RawTypedUserIdExtractor<TBacking, TId> : StringUserIdExtractor, IUserIdExtractor<TId>
+    where TBacking : struct
+    where TId : struct, IRawTypedId<TBacking, TId>
+{
+    private static readonly Dictionary<Type, Func<string, object>> Parsers = new Dictionary<Type, Func<string, object>>
```
jakubfijalkowski: Let's go with [`ReadOnlyDictionary`](https://learn.microsoft.com/en-us/dotnet/api/system.collections.objectmodel.readonlydictionary-2?view=net-8.0).

On the other hand, I would say that we should add `Parse` to the `IRawTypedId` so that we don't need the dictionary at all.

src/Helpers/LeanCode.UserIdExtractors/ServiceProviderExtensions.cs
```diff
@@ -0,0 +1,87 @@
+using System.Reflection;
+using LeanCode.DomainModels.Ids;
+using LeanCode.UserIdExtractors.Extractors;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace LeanCode.UserIdExtractors;
+
+public static class UserServiceProviderExtensions
```
jakubfijalkowski, line 7: Oh boi. You did a lot of work, except that, I think, it won't ever be used.
Let's estimate how many "user ids" we have in a project. I would say, that typically, there is only one user id - you either use `Guid` or use one, single typed id like `UserId`. You then use it across the whole system. This means that you need exactly one type `IUserIdExtractor`. What is more, you don't need any assembly scanning at all, because you know the type up-front.

src/Helpers/LeanCode.UserIdExtractors/ServiceProviderExtensions.cs
```diff
@@ -0,0 +1,87 @@
...
+        services.AddRawTypedUserIdExtractors(idAssemblies, userIdClaim);
+
+        return services;
+    }
+
+    private static void AddBasicUserIdExtractors(this IServiceCollection services, string userIdClaim)
+    {
+        var stringInstance = Activator.CreateInstance(typeof(StringUserIdExtractor), userIdClaim)!;
+        var genericStringInstance = Activator.CreateInstance(typeof(GenericStringUserIdExtractor), userIdClaim)!;
+        var guidInstance = Activator.CreateInstance(typeof(GuidUserIdExtractor), userIdClaim)!;
```
jakubfijalkowski:
```suggestion
        var stringInstance =new StringUserIdExtractor(userIdClaim)!;
        var genericStringInstance = new GenericStringUserIdExtractor(userIdClaim)!;
        var guidInstance = new GuidUserIdExtractor(userIdClaim)!;
```
;P

jakubfijalkowski (changes requested):

src/Helpers/LeanCode.UserIdExtractors/Extractors/RawTypedUserIdExtractor.cs
Saancreed: I'd say we should go a step further and use `FrozenDictionary` for such cases.

jakubfijalkowski: Ah! That's the one I had in mind TBH, but mistook the name.

src/Helpers/LeanCode.UserIdExtractors/LeanCode.UserIdExtractors.csproj
```diff
@@ -0,0 +1,6 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
```
Saancreed: Do we need to pull the entire AspNetCore here? This looks like this package could get away with just some `DependencyInjection.Abstractions` reference.

test/Helpers/LeanCode.UserIdExtractors.Tests/UserIdExtractorsTests.cs
```diff
@@ -0,0 +1,105 @@
...
+{
+    private const string UserIdClaim = "sub";
+
+    private readonly IServiceProvider serviceProvider;
+
+    public UserIdExtractorsTests()
+    {
+        var services = new ServiceCollection();
+
+        services.AddUserIdExtractors([ Assembly.GetExecutingAssembly() ], UserIdClaim);
```
Saancreed: Nitpicking (something for another PR most likely) but there should be no spaces between `[` and the first element (and between the last element and `]`), this is a CSharpier bug that has been fixed in some recent release. So we will need a CSharpier bump, probably sooner than later.

src/Helpers/LeanCode.UserIdExtractors/IUserIdExtractor.cs
```diff
@@ -0,0 +1,14 @@
...
+
+public interface IUserIdExtractor<TUserId> : IUserIdExtractor
+    where TUserId : notnull, IEquatable<TUserId>
+{
+    TUserId ExtractId(ClaimsPrincipal user);
+}
+
+public interface IUserIdExtractor
+{
+    string Extract(ClaimsPrincipal user);
```
Saancreed: Having the names be different feels weird. Is this about having to hide the base method otherwise? Because I'd prefer this to follow the same pattern as `IEnumerable` / `IEnumerable<T>`.

wojtek2288: Yea it's hiding base method otherwise, we can either keep both methods with different names or do not inherit `IUserIdExtractor` in `IUserIdExtractor<TUserId>` at all.

Saancreed: Ah, I see. That's unfortunate, and frankly I'm not sure which approach is better. I feel like we should have a larger discussion about this (on Slack?) but it's probably not important enough to warrant one.

test/Helpers/LeanCode.UserIdExtractors.Tests/UserIdExtractorsTests.cs
wojtek2288: Will bump it, probably not in this PR though

src/CQRS/LeanCode.CQRS.Security/UserIdExtractor.cs
wojtek2288: https://leancode.slack.com/archives/CH27S36HZ/p1701867170839379 - discussion about it

src/Helpers/LeanCode.UserIdExtractors/Extractors/RawTypedUserIdExtractor.cs
wojtek2288: Will add `Parse` method to `IRawTypedId` but I think in other PR, keeping frozen dict for now

src/Helpers/LeanCode.UserIdExtractors/Extractors/RawTypedUserIdExtractor.cs
```diff
@@ -0,0 +1,47 @@
...
+
+    private static TBacking GetBacking(string value)
+    {
+        var backingType = typeof(TBacking);
+        var parsedValue = Parsers[backingType].Invoke(value);
+
+        return (TBacking)Convert.ChangeType(parsedValue, typeof(TBacking), CultureInfo.InvariantCulture);
+    }
+
+    string IUserIdExtractor.Extract(ClaimsPrincipal user) => Extract(user).ToString()!;
```
Saancreed: How about we moved this to the interface itself so it becomes the default implementation?

src/Helpers/LeanCode.UserIdExtractors/LeanCode.UserIdExtractors.csproj
```diff
@@ -0,0 +1,6 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection" />
```
Saancreed: `.Abstractions`? Or do you need anything from this package in particular?

src/Helpers/LeanCode.UserIdExtractors/Extractors/RawTypedUserIdExtractor.cs
```diff
@@ -0,0 +1,45 @@
...
+        var backing = GetBacking(claim);
+        return TId.Parse(backing);
+    }
+
+    private static TBacking GetBacking(string value)
+    {
+        var backingType = typeof(TBacking);
+        var parsedValue = Parsers[backingType].Invoke(value);
+
+        return (TBacking)Convert.ChangeType(parsedValue, typeof(TBacking), CultureInfo.InvariantCulture);
```
Saancreed: Isn't `parsedValue` already of type `TBacking`? This looks like you could get away with just `return (TBacking)parsedValue;` and avoid the call completely.

src/Helpers/LeanCode.UserIdExtractors/ServiceProviderExtensions.cs
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        return services;
+    }
+
+    private static bool TryAddRawTypedUserIdExtractor(IServiceCollection services, Type userIdType, string userIdClaim)
+    {
+        var rawTypedId = GetImplementedIdTypes(userIdType, typeof(IRawTypedId<,>)).FirstOrDefault();
+
+        if (rawTypedId != null)
```
Saancreed:
```suggestion
        if (rawTypedId is not null)
```

I wonder if there's an analyzer for this somewhere :thinking:

src/Helpers/LeanCode.UserIdExtractors/ServiceProviderExtensions.cs
```diff
@@ -0,0 +1,89 @@
...
+
+            return true;
+        }
+        else
+        {
+            return false;
+        }
+    }
+
+    private static IEnumerable<Type> GetImplementedIdTypes(Type type, Type idKind)
```
Saancreed: You only ever use this function to find the first interface type (or check if one exists). Consider simplifying this to something like `Type? TryGetImplementedGenericType(Type type, Type genericTypeDefinition)` and check the result for `null` at call sites.

Saancreed (approved):
Last few nits and otherwise LGTM, even if the feature itself is still a bit controversial.

Now let's see if others approve as well :see_no_evil:

test/Helpers/LeanCode.UserIdExtractors.Tests/UserIdExtractorsTests.cs
```diff
@@ -0,0 +1,127 @@
+using System.Security.Claims;
```
jakubfijalkowski, line 1: This needs to be split - you have here unrelated tests. Tests for extractors should be separated from tests for DI.

src/Helpers/LeanCode.UserIdExtractors/ServiceProviderExtensions.cs
```diff
@@ -0,0 +1,91 @@
+using LeanCode.DomainModels.Ids;
+using LeanCode.UserIdExtractors.Extractors;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace LeanCode.UserIdExtractors;
+
+public static class UserServiceProviderExtensions
```
jakubfijalkowski, line 7: This class is IMO slightly over engineered. Why not provide three different methods, one for Guid, one for string, and one for typed ids? You could then get rid of the reflection.

wojtek2288: To get rid of the reflection we would probably need 4 methods: guid, string, raw and prefixed because `IRawTypedId` has 2 generic arguments: `TBacking` and `TSelf` and `TBacking` would also have to be provided as generic argument to the extension method.

So we could:
* Keep it as it is,
* Have 3 methods: non-generic string/guid, generic typedId for prefixed and raw (but uses reflection),
* Have 4 methods: non-generic string/guid, generic prefixed (with 1 arg and proper generic constraints), generic raw (with 2 args and proper generic constraints).

jakubfijalkowski: I'm all. For less reflection, so last option works the best for me (but we can consult what the people think).

jakubfijalkowski (approved):

PR #629, opened 2023-12-19
Rename `IIdentifiable` to `IEntity`

jakubfijalkowski:
Okay folks, I finally got fed up with that one (when reading the new docs).


marcin-dardzinski:
I'd at least keep obsoleted `IIdentifiable<TId> : IEntity<TId>` for some time - will give some air to breathe in migrations.

jakubfijalkowski:
I disagree to be honest. The only migration necessary here is a grand "search and replace" of `IIdentifiable` to `IEntity`. Not much TBH.

marcin-dardzinski:
Well, my latest experience is that 'just one more search'n'replace' quickly tends to snowball beyond readability.

jakubfijalkowski:
I somewhat don't get the readability argument (you can execute everything step-by-step, you don't have to make everything in a single step) and I don't think this adds anymore to the complexity of upgrading, but nevertheless, `Obsolete`d interface that we will get rid soon is OK with me.

Dragemil (approved):

michal12334 (approved):

PR #630, opened 2024-01-06
Local execution of CQRS

jakubfijalkowski:
Okay, now, this is somewhat ready - local execution, with separate pipeline and "custom" `HttpContext` that is mocked.

---
## Old comment

Okay, here is my idea of "local CQRS execution", i.e. how to restore the functionality of `I*Executor` that we ditched by switching to ASP.NET Core Middlewares.

Basically, the idea is to build the middleware pipeline for CQRS the same way the ASP.NET Core server does, but without the HTTP stuff. This turned out to be pretty easy and working wonderfully. The main problem is the `HttpContext` - we need to create it out of thin air, basically.

Here, I went with a "wrapper" way - i.e. create a `HttpContext` out of the base `HttpContext` from the original call. This is not the only option, fortunately - we can easily make an empty `HttpContext` and just remove `Feature`s that we don't want. This might be problematic if we want to allow all middlewares to be used (which might not make sense because "response" is somewhat undefined in local calls).

When it comes to the "how to build the pipeline", we could go with, I think, two approaches - use the pipeline from `MapRemoteCqrs` or have dedicated local one. I really think we should force users to declare both pipelines differently because they serve different purpose. We could have a way to re-use them, but it should still be visible from the external code. That's why I went with the `WithLocalCommands` that accepts configuration. You can check the initial commit for an idea on how to re-use pipelines across different... modes of execution.

This is PoC, as most of the code is written in random places, and not everything is supported, but it should explain the basic idea.


src/CQRS/LeanCode.CQRS.AspNetCore/Local/LocalHttpContext.cs
```diff
@@ -0,0 +1,63 @@
...
+    public LocalHttpContext(HttpContext inner, IServiceProvider requestServices, CancellationToken requestAborted)
+    {
+        this.inner = inner;
+        this.requestServices = requestServices;
+        this.requestAborted = requestAborted;
+
+        features = new LocalFeatureCollection(inner.Features);
+    }
+
+    public override HttpResponse Response =>
```
marcin-dardzinski: This might be problematic if some handler / middleware would like to use it, e.g setting some headers (cache?). Maybe an empty response would be better.

src/CQRS/LeanCode.CQRS.AspNetCore/Local/MiddlewareBasedLocalCommandExecutor.cs
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    public async Task<CommandResult> RunAsync<T>(
+        HttpContext context,
+        T command,
+        CancellationToken cancellationToken = default
+    )
+        where T : ICommand
+    {
+        await using var scope = serviceProvider.CreateAsyncScope();
```
marcin-dardzinski: I wonder if we should use the outer http context scope to create inner scope here.

src/CQRS/LeanCode.CQRS.AspNetCore/Local/MiddlewareBasedLocalCommandExecutor.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        var localContext = new LocalHttpContext(context, scope.ServiceProvider, cancellationToken);
+
+        localContext.SetCQRSRequestPayload(command);
```
marcin-dardzinski: We have one more problem here. Middlewares often rely on `CQRSRequestMetadata` which currently is accessed through [endpoint routing metadata](https://github.com/leancodepl/corelibrary/blob/v8.0-preview/src/CQRS/LeanCode.CQRS.Execution/HttpContextExtensions.cs#L8). Probably we should store it as a feature as well and make this code and `CQRSMIddleware` set it.

src/CQRS/LeanCode.CQRS.AspNetCore/Local/MiddlewareBasedLocalCommandExecutor.cs
```diff
@@ -0,0 +1,81 @@
...
+        return localPayload.ExecuteAsync(context);
+    }
+}
+
+internal interface ILocalCommandPayload
+{
+    Task ExecuteAsync(HttpContext context);
+}
+
+internal class LocalCommandPayload<T> : ILocalCommandPayload
```
marcin-dardzinski: This probably should be unified with object `ObjectExecutor` somehow.

src/CQRS/LeanCode.CQRS.AspNetCore/ServiceCollectionCQRSExtensions.cs
```diff
@@ -85,4 +85,12 @@ public CQRSServicesBuilder AddCQRSObjects(TypesCatalog contractsCatalog, TypesCa
         );
         return this;
     }
+
+    public CQRSServicesBuilder WithLocalCommands(Action<ICQRSApplicationBuilder> configure)
+    {
+        Services.AddSingleton<Local.ILocalCommandExecutor>(
+            s => new Local.MiddlewareBasedLocalCommandExecutor(s, configure)
```
marcin-dardzinski: Two things there:
1. I wonder if it's good that the same command may have different local and 'remote' pipeline
2. We can have multiple 'remote' pipelines, but only one local?

These are not top priority things, but at some point we'll need to address them.

marcin-dardzinski (commented):
Looks promising, highlighted things that need more love.

lukaszgarstecki:
I wonder if it will be possible to use every C/Q/O as local and remote at the same time? If I get your implementation correctly this is general idea. 

>Here, I went with a "wrapper" way - i.e. create a HttpContext out of the base HttpContext from the original call. This is not the only option, fortunately - we can easily make an empty HttpContext and just remove Features that we don't want. This might be problematic if we want to allow all middlewares to be used (which might not make sense because "response" is somewhat undefined in local calls).

Why do you say `response` is undefined? Do you plan to use some other mechanism to return `CommandResult` or query/operation response? Maybe I didn't get it...
For sure we need a way to define what we want to propagate some metadata that we currently pass outside of http request body.

test/CQRS/LeanCode.CQRS.AspNetCore.Tests/Local/MiddlewareBasedLocalCommandExecutorTests.cs
```diff
@@ -0,0 +1,192 @@
...
+{
+    private const string IsAuthenticatedHeader = "is-authenticated";
+    private static readonly TypesCatalog ThisCatalog = TypesCatalog.Of<LocalCommand>();
+
+    private readonly IHost host;
+    private readonly TestServer server;
+
+    public MiddlewareBasedLocalCommandExecutorTests()
+    {
+        host = new HostBuilder()
```
Saancreed: This makes me wonder if we shouldn't prefer `WebApplication.Create…Builder()` (or even `Host.Create…Builder()`) flavor of host building nowadays.

Well, maybe not here but in `LeanProgram` :see_no_evil:

src/CQRS/LeanCode.CQRS.AspNetCore/Local/LocalHttpContext.cs
```diff
@@ -0,0 +1,63 @@
+using System.Security.Claims;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features;
+
+namespace LeanCode.CQRS.AspNetCore.Local;
+
+internal class LocalHttpContext : HttpContext
```
Saancreed: Is there any particular reason why we need our own `HttpContext` class for this? From the first look it appears that there isn't any critical stuff we need to override so maybe we could get away with a simple helper that just returns a preconfigured context for local execution?

src/CQRS/LeanCode.CQRS.AspNetCore/Local/LocalHttpContext.cs
jakubfijalkowski: I am also wondering how to address it. On one hand, supporting it would be good, on the other - there is no HTTP and we can't rely on that. Using HTTP cache in local exec does not make any sense and I'm somewhat struggling with lying that we support it.

src/CQRS/LeanCode.CQRS.AspNetCore/Local/MiddlewareBasedLocalCommandExecutor.cs
jakubfijalkowski: Won't work. We rely on DI scopes for proper changes containment. Not opening scope here would result in re-using DbContext, which would have very much unintended consequences for the end user (and we still want to make command a transaction boundary).

marcin-dardzinski: I mean use `httpContext.RequestServices.CreateAsyncScope()`, not pass it directly.

src/CQRS/LeanCode.CQRS.AspNetCore/Local/LocalHttpContext.cs
marcin-dardzinski: My point here is that _some_ middlewares / handlers may want to use it and don't know that they're in a local context. It's better to let them write to /dev/null then raise an exception IMO.

Unless you want to be explicit with a local pipeline and make the end user responsible for enforcing it.

jakubfijalkowski: The problem with that is _how_ should that work - we're not in real HTTP context. We can pretend that there is a request because it is read-only, but the same thing for response does not work at all.

The more I think about it, the more I am in favor of explicit local pipeline, where the user is required to define both pipelines separately.

src/CQRS/LeanCode.CQRS.AspNetCore/Local/MiddlewareBasedLocalCommandExecutor.cs
jakubfijalkowski: Ah, got it. There is a separate problem here TBH - some DI containers support hierarchical scopes (even Autofac up until v8 I think) and I'm not sure if we should allow that.

src/CQRS/LeanCode.CQRS.AspNetCore/Local/MiddlewareBasedLocalCommandExecutor.cs
jakubfijalkowski: Yeah, I think that is true.

I'm starting to wonder if we should split the "metadata" about the request from the "metadata" about execution here somehow, so that we don't have to resolve `HandlerType` too early.

src/CQRS/LeanCode.CQRS.AspNetCore/Local/MiddlewareBasedLocalCommandExecutor.cs
jakubfijalkowski: Yeah, this deserves some refactoring. I wanted to contain the changes as much as we can.

src/CQRS/LeanCode.CQRS.AspNetCore/ServiceCollectionCQRSExtensions.cs
jakubfijalkowski: 1. That's something I also wonder myself, but considering how different are these modes, then I am leaning in favor of splitting them (because some middlewares won't make sense in local calls, some won't make sense in remote calls).
2. Yeah, this will need a solution.

test/CQRS/LeanCode.CQRS.AspNetCore.Tests/Local/MiddlewareBasedLocalCommandExecutorTests.cs
jakubfijalkowski: Totally unrelated ;P

src/CQRS/LeanCode.CQRS.AspNetCore/Local/LocalHttpContext.cs
marcin-dardzinski: You cannot just instantiate `HttpResponse` (`DefaultHttpResponse` ?) and have it completely in memory? Maybe a custom null object implementation?

src/CQRS/LeanCode.CQRS.AspNetCore/Local/LocalHttpContext.cs
jakubfijalkowski: The idea here is to have a "derived" context that basically decorates the "original" `HttpContext`, not fully separate context. We probably could work around it with copying necessary `Features`, but having separate class that guards "just enough" is IMO cleaner.

jakubfijalkowski:
> I wonder if it will be possible to use every C/Q/O as local and remote at the same time?

To be hones, I don't know. I also don't want to distinguish between "local" and "remote" too much.

> If I get your implementation correctly this is general idea.

Somewhat yes, but I wouldn't be attached to that.

> Why do you say `response` is undefined? Do you plan to use some other mechanism to return `CommandResult` or query/operation response? Maybe I didn't get it...

In local calls, there is no HTTP, thus there is no response. In the pipeline, the result of command/query/operation is already transported with a dedicated `Feature`, and translated to response in [CQRSMiddleware](https://github.com/leancodepl/corelibrary/blob/v8.0-preview/src/CQRS/LeanCode.CQRS.AspNetCore/Middleware/CQRSMiddleware.cs), this solution uses that mechanism already and I don't plan involving HTTP here (it would add unnecessary overhead).

> For sure we need a way to define what we want to propagate some metadata that we currently pass outside of http request body.

What kind of metadata are you thinking about here? The "input metadata" is already propagated through `Request` (and `User`) which has some drawbacks but solves the problem. There is problem with "output metadata" (e.g. cache headers), but for a local call it somewhat does not make sense to have them either way, so I'm quite willing to sacrifice that.

src/CQRS/LeanCode.CQRS.AspNetCore/Local/MiddlewareBasedLocalCommandExecutor.cs
marcin-dardzinski: What do you mean by 'request metadata`? Command type? Endpoint routing?

If we assume one type can only be at most one of command/query we can query `CQRSObjectsRegistrationSource` to get `CQRSObjectMetadata`. Then we have all the information to finalize the pipeline.

lukaszgarstecki:
>What kind of metadata are you thinking about here? The "input metadata" is already propagated through Request (and User) which has some drawbacks but solves the problem. There is problem with "output metadata" (e.g. cache headers), but for a local call it somewhat does not make sense to have them either way, so I'm quite willing to sacrifice that.

If we pass whole `Request` than this is all I need :smile:

src/CQRS/LeanCode.CQRS.AspNetCore/Local/ILocalCommandExecutor.cs
```diff
@@ -0,0 +1,10 @@
+using LeanCode.Contracts;
+using Microsoft.AspNetCore.Http;
+
+namespace LeanCode.CQRS.AspNetCore.Local;
+
+public interface ILocalCommandExecutor
+{
+    Task<CommandResult> RunAsync<T>(HttpContext context, T command, CancellationToken cancellationToken = default)
```
marcin-dardzinski: What if we don't have any http context at hand? Should the caller create their own fake one? I think at minimum we'd just need DI scope and claims principal to run a command, the rest might be 'faked' for the sake of local execution.

jakubfijalkowski: > we'd just need DI scope and claims principal

`ClaimsPrincipal` will be enough I think. DI scope should not be controlled from the outside, because each call _must_ open a new scope either way (otherwise a lot of things will break).

Also, the more I think about it, the more I am in favor of **not** exposing the `HttpContext` overload. Mocking everything is just _more_ correct, and even giving access to `Request` is wrong (because that is completely different request than the one being run). /cc: @lukaszgarstecki, I said something different earlier.

jakubfijalkowski: Also, implementing `Request` and `Response` as just "return empty data and ignore writes" is more and more OK with me.

src/CQRS/LeanCode.CQRS.AspNetCore/ServiceCollectionCQRSExtensions.cs
jakubfijalkowski: Okay, after some thought, this is what I came up with:
1. Local and remote pipelines are really different. It is up to the user to make it work in both. These can be the same, but it should be a deliberate choice (as, e.g., HTTP cache does not make any sense in local calls).
2. Keyed services to the rescue. I'm thinking of somewhat simplified `IHttpClientFactory` for local executors. Also, I think the key should be the only differentiator and we should allow all objects to go through it _by default_ (otherwise it might get tricky to explain to users how it works).

src/CQRS/LeanCode.CQRS.AspNetCore/Local/ILocalCommandExecutor.cs
jakubfijalkowski: Done, no `HttpContext` in sight.

src/CQRS/LeanCode.CQRS.AspNetCore/Local/LocalHttpContext.cs
jakubfijalkowski: There is a completely separate `LocalCallContext` that works differently now.

src/CQRS/LeanCode.CQRS.AspNetCore/ServiceCollectionCQRSExtensions.cs
jakubfijalkowski: (1) done

src/CQRS/LeanCode.CQRS.AspNetCore/Local/MiddlewareBasedLocalCommandExecutor.cs
jakubfijalkowski: If we don't require `HttpContext`, then we can't use `RequestServices`.

src/CQRS/LeanCode.CQRS.AspNetCore/Local/LocalHttpContext.cs
jakubfijalkowski: Done

src/CQRS/LeanCode.CQRS.AspNetCore/ServiceCollectionCQRSExtensions.cs
jakubfijalkowski: Both done.

test/CQRS/LeanCode.CQRS.AspNetCore.Tests/Local/MiddlewareBasedExecutorIntegrationTests.cs
```diff
@@ -0,0 +1,143 @@
...
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.TestHost;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Xunit;
+
+namespace LeanCode.CQRS.AspNetCore.Tests.Local;
+
+public class MiddlewareBasedExecutorIntegrationTests : IDisposable, IAsyncLifetime
```
Dragemil, line 18: Wouldn't using `WebApplicationFactory` over raw host with `IAsyncLifetime` be more convenient?

test/CQRS/LeanCode.CQRS.AspNetCore.Tests/ServiceProviderRegistrationExtensionsTests.cs
```diff
@@ -0,0 +1,161 @@
...
+        public Task SerializeAsync(
+            Stream utf8Json,
+            object value,
+            Type inputType,
+            CancellationToken cancellationToken
+        ) => throw new NotImplementedException();
+    }
+}
+
+file class ServiceProviderAssertions : ReferenceTypeAssertions<ServiceProvider, ServiceProviderAssertions>
```
Dragemil: That's something I see used for the first time.

test/CQRS/LeanCode.CQRS.AspNetCore.Tests/ServiceProviderRegistrationExtensionsTests.cs
```diff
@@ -0,0 +1,161 @@
...
+
+    private static ServiceProvider BuildProvider(Action<CQRSServicesBuilder>? configure = null)
+    {
+        var collection = new ServiceCollection();
+        var builder = collection.AddCQRS(ThisCatalog, ThisCatalog);
+        configure?.Invoke(builder);
+        return collection.BuildServiceProvider();
+    }
+
+    internal class CustomSerializer : ISerializer
```
marcin-dardzinski: We'd better seal the class as he asks for - I don't want it to be stacking at every PR.

src/CQRS/LeanCode.CQRS.AspNetCore/ServiceCollectionCQRSExtensions.cs
```diff
@@ -88,4 +89,59 @@ public CQRSServicesBuilder AddCQRSObjects(TypesCatalog contractsCatalog, TypesCa
...
+    public CQRSServicesBuilder WithLocalOperations(Action<ICQRSApplicationBuilder> configure)
+    {
+        Services.AddSingleton<Local.ILocalOperationExecutor>(
+            s =>
+                new Local.MiddlewareBasedLocalOperationExecutor(s, s.GetRequiredService<ICQRSObjectSource>(), configure)
+        );
+        return this;
+    }
+
+    public CQRSServicesBuilder WithKeyedLocalCommands(object? serviceKey, Action<ICQRSApplicationBuilder> configure)
```
marcin-dardzinski, line 118: If I understand it correctly, the idea is I can have multiple keyed `ILocalExecutor`s (each with different pipeline, each able to run every command) and it's up to the caller to chose which implementation they wan't to resolve.
 
I find it a bit strange a design decision - I'd expect to have executor which is doing the 'routing' under the hood, but if the latter is harder to achieve then OK.

marcin-dardzinski (commented):
Some documentation entry would be nice...

test/CQRS/LeanCode.CQRS.AspNetCore.Tests/Local/MiddlewareBasedExecutorIntegrationTests.cs
jakubfijalkowski: I would still need to defined `Program` somewhere, or derive from it, I don't think it will result in substantially easier code.

test/CQRS/LeanCode.CQRS.AspNetCore.Tests/ServiceProviderRegistrationExtensionsTests.cs
jakubfijalkowski: This is the first time where I find this useful. ;)

test/CQRS/LeanCode.CQRS.AspNetCore.Tests/ServiceProviderRegistrationExtensionsTests.cs
jakubfijalkowski: I can't make it report it reliably on my dev machine and find about that on CI only.

src/CQRS/LeanCode.CQRS.AspNetCore/ServiceCollectionCQRSExtensions.cs
jakubfijalkowski: > If I understand it correctly, the idea is I can have multiple keyed ILocalExecutors (each with different pipeline, each able to run every command) and it's up to the caller to chose which implementation they wan't to resolve.

Yes. I also don't think one should use `ILocal*Executor` directly - it should be wrapped in some higher-level class that does some checking.

> I find it a bit strange a design decision

This mostly follows the `IHttpClientFactory` way of doing things.

> I'd expect to have executor which is doing the 'routing' under the hood

I really don't want to force "one true way" right now, as I'm not really sure that "routing" is a correct approach here. Also, I do think there are easier, faster and more ergonomic ways of doing things than one big router (i.e. the first line of this comment).

jakubfijalkowski:
> Some documentation entry would be nice...

Next PR. This one is too big already.

test/CQRS/LeanCode.CQRS.AspNetCore.Tests/ServiceProviderRegistrationExtensionsTests.cs
jakubfijalkowski: Fixed

Dragemil (approved):

marcin-dardzinski (approved):

PR #636, opened 2024-01-26
API Explorer integration

jakubfijalkowski:
This PR adds integration to the ASP.NET Core MVC API Explorer. This also enables Swashbuckle and NSwag, thus giving us full OpenAPI & Swagger UI for "free".

It does not work OOTB because the default Minimal APIs integration requires every endpoint to have a [`MethodInfo`](https://github.com/dotnet/aspnetcore/blob/bd016947b6928159814950c06a6e2278be7a256f/src/Mvc/Mvc.ApiExplorer/src/EndpointMetadataApiDescriptionProvider.cs#L63) metadata. It then uses it to generate parameter/response definitions. Unfortunately, this won't work for us as we have a slightly different model. That's why we need to register the endpoints manually.

### DIFF
--- a/src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSApiDescriptionProvider.cs
+++ b/src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSApiDescriptionProvider.cs
@@ -1,0 +1,270 @@
+using System.Collections.Immutable;
+using System.Text;
+using LeanCode.Contracts;
+using LeanCode.CQRS.Execution;
+using Microsoft.AspNetCore.Mvc.Abstractions;
+using Microsoft.AspNetCore.Mvc.ApiExplorer;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
+using Microsoft.AspNetCore.Mvc.ModelBinding.Metadata;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.AspNetCore.Routing.Patterns;
+
+namespace LeanCode.CQRS.AspNetCore.Registration;
+
+internal sealed class CQRSApiDescriptionProvider : IApiDescriptionProvider
+{
+    private const string ApplicationJson = "application/json";
+
+    private readonly EndpointDataSource endpointDataSource;
+
+    public int Order => -1200;
+
+    public CQRSApiDescriptionProvider(EndpointDataSource endpointDataSource)
+    {
+        this.endpointDataSource = endpointDataSource;
+    }
+
+    public void OnProvidersExecuting(ApiDescriptionProviderContext context)
+    {
+        foreach (var endpoint in endpointDataSource.Endpoints)
+        {
+            if (
+                endpoint is RouteEndpoint routeEndpoint
+                && routeEndpoint.Metadata.GetMetadata<CQRSEndpointMetadata>() is { } metadata
+            )
+            {
+                context.Results.Add(CreateApiDescription(routeEndpoint, metadata));
+            }
+        }
+    }
+
+    public void OnProvidersExecuted(ApiDescriptionProviderContext context) { }
+
+    private static ApiDescription CreateApiDescription(RouteEndpoint routeEndpoint, CQRSEndpointMetadata metadata)
+    {
+        var apiDescription = new ApiDescription
+        {
+            HttpMethod = HttpMethod.Post.Method,
+            GroupName = null,
+            RelativePath = BuildPath(routeEndpoint.RoutePattern),
+            ActionDescriptor = new ActionDescriptor
+            {
+                DisplayName = routeEndpoint.DisplayName,
+                RouteValues = { ["controller"] = "CQRS", }, // Required by Swagger (Swashbuckle)
+            },
+        };
+        apiDescription.SupportedRequestFormats.Add(new() { MediaType = ApplicationJson });
+
+        if (metadata.ObjectMetadata.ObjectKind == CQRSObjectKind.Command)
+        {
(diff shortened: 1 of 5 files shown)


marcin-dardzinski (approved):

src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSApiDescriptionProvider.cs
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    public void OnProvidersExecuted(ApiDescriptionProviderContext context) { }
+
+    private static ApiDescription CreateApiDescription(RouteEndpoint routeEndpoint, CQRSEndpointMetadata metadata)
+    {
+        var apiDescription = new ApiDescription
+        {
+            HttpMethod = HttpMethod.Post.Method,
+            GroupName = null,
```
Dragemil, line 48: The endpoints on the UI are grouped by something, on MVC this is by controller ofc, and for our case it would nice to group them by namespace after the project name. Dunno if that's actually the property that would achieve that, but exposing some lambda that would allow defining groups or just straight up implement it as I said could be something.

Dragemil: But that could be in another iteration of this feature.

src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSApiDescriptionProvider.cs
```diff
@@ -0,0 +1,270 @@
...
+
+namespace LeanCode.CQRS.AspNetCore.Registration;
+
+internal sealed class CQRSApiDescriptionProvider : IApiDescriptionProvider
+{
+    private const string ApplicationJson = "application/json";
+
+    private readonly EndpointDataSource endpointDataSource;
+
+    public int Order => -1200;
```
Dragemil, line 20: Oddly specific 🤨

Dragemil (approved):
Another iteration of this feature could also respect comments on the contracts as docs (if it doesn't already?)

src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSApiDescriptionProvider.cs
jakubfijalkowski: Or - random. :)

src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSApiDescriptionProvider.cs
jakubfijalkowski: Although you are correct in the general, this property is misleading - Swashbuckle _filters_ ApiDescriptors based on the `GroupName`. It then uses `ActionDescriptor.RouteValues["controller"]` to do the grouping.

I thought about configuration here but the question is _how_ do to it - we should be able to have some guidelines, and coming up with a sensible one is tough.

jakubfijalkowski:
> Another iteration of this feature could also respect comments on the contracts as docs

I was unable to (quickly) find how to populate the `ApiDescriptor` with comments - it seems as though it is unsupported directly (nor Swashbuckle code says that).

src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSApiDescriptionProvider.cs
Dragemil: One that I came up with: 
Assume we have an endpoint `<host>/(query|command)/<contracts_assembly>/<namespace>/<contract>`,
where `<contracts_assembly>` actually is `<project>.<domain>.Contracts`. Then I think having group names as: `<domain> <namespace>` would be pretty good. Since the (swashbuckle) UI is ordered by groups that'd give us some nice ordering ootb as well.

Dragemil: Another thing is I've just tried it, and the endpoints on the UI are ordered somewhat randomly. I wonder if we could order them, and if so, by `<contracts_assembly>...` would be fine I think.

jakubfijalkowski: We don't control how the UI displays things, unfortunately. The only thing that is given is that the endpoints are grouped by `controller`, which is somewhat opaque string to the tooling (and generating it sensibly is tricky, because most of the time, you want it grouped differently at different "level" of contracts, and probably you would want it grouped differently depending on the goal that you have at the moment).

We also can't order them. From what I found out, it seems that they get sorted by the `RelativePath`, which we can't abuse.

Dragemil: I get it, but what I'm saying that we could allow to implement `controller` value as I said, or at least expose a lambda for doing so, rather than just hard coding it to `CQRS` as it is now.

jakubfijalkowski: The problem that I have - it will be half-baked. If we only allow controlling `controller` and it turns out that some other variable also needs to be set, we will _promise_ some feature that you can't use. And considering that we stabilized the v8, if the API will need to change (e.g. two different fields need to be set), we would need to introduce a breaking change.

It's better to investigate deeper and have something that _works_ from the get go.

PR #639, opened 2024-02-06
Ensure that `ObjectExecutorFactory` resolves the handler that was used to generate metadata

jakubfijalkowski:
Previously, the `ObjectExecutorFactory` resolved `IXHandler<A>` directly from the container. The problem with that approach is that we can't be sure that the `TypeAssembly` that we get in `CQRSObjectsRegistrationSource.AddCQRSObjects` represents _all_ the assemblies that the user scanned for handlers when adding them to DI. For example, it is still possible to have this:

```
Contracts.csproj
   - Command : ICommand

HandlersA.csproj
  - CH1: ICommandHandler<Command>

HandlersB.csproj
  - CH2: ICommandHandler<Command>
```

Then in `App` have:
```
serviceCollection
    .AddCQRS(TypesAssembly.Of<Command>(), TypesAssembly.Of<CH1>())
    .AddCQRSObjects(TypesAssembly.Of<Command>(), TypesAssembly.Of<CH2>());
```

which would result, in the end, in (effectively, the code will be slightly different):
```
serviceCollection.AddTransient<ICommandHandler<Command>, CH1>();
serviceCollection.AddTransient<ICommandHandler<Command>, CH2>();
```

But the `CQRSObjectsRegistrationSource.objects` will have only the `Command, CH1` pair. Thus, the pipeline will _think_ that it executes `CH1`, but `CH2` will be resolved in `ObjectExecutorFactory.ExecuteCommandAsync`.

I admit that this is rather unlikely situation, but when it occurs, it might be very hard to detect. I also don't think this PR makes detection easier, but _at least_ it will resolve `CH1` when it thinks about `CH1`. This _might_ be more important for local execution though.

### DIFF
--- a/src/CQRS/LeanCode.CQRS.AspNetCore/Registration/ServiceCollectionRegistrationExtensions.cs
+++ b/src/CQRS/LeanCode.CQRS.AspNetCore/Registration/ServiceCollectionRegistrationExtensions.cs
@@ -20,9 +20,8 @@ public static class ServiceCollectionRegistrationExtensions
 
     public static void AddCQRSHandler(this IServiceCollection serviceCollection, CQRSObjectMetadata obj)
     {
-        serviceCollection.Add(
-            new ServiceDescriptor(MakeHandlerInterfaceType(obj), obj.HandlerType, ServiceLifetime.Scoped)
-        );
+        serviceCollection.Add(new(MakeHandlerInterfaceType(obj), obj.HandlerType, ServiceLifetime.Scoped));
+        serviceCollection.Add(new(obj.HandlerType, obj.HandlerType, ServiceLifetime.Scoped));
 
         Type MakeHandlerInterfaceType(CQRSObjectMetadata obj)
         {
--- a/src/CQRS/LeanCode.CQRS.AspNetCore/Registration/ServiceCollectionRegistrationExtensions.cs
+++ b/src/CQRS/LeanCode.CQRS.AspNetCore/Registration/ServiceCollectionRegistrationExtensions.cs
@@ -20,8 +20,10 @@ public static class ServiceCollectionRegistrationExtensions
 
     public static void AddCQRSHandler(this IServiceCollection serviceCollection, CQRSObjectMetadata obj)
     {
-        serviceCollection.Add(new(MakeHandlerInterfaceType(obj), obj.HandlerType, ServiceLifetime.Scoped));
         serviceCollection.Add(new(obj.HandlerType, obj.HandlerType, ServiceLifetime.Scoped));
+        serviceCollection.Add(
+            new(MakeHandlerInterfaceType(obj), sp => sp.GetRequiredService(obj.HandlerType), ServiceLifetime.Scoped)
+        );
 
         Type MakeHandlerInterfaceType(CQRSObjectMetadata obj)
         {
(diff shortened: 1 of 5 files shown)


src/CQRS/LeanCode.CQRS.AspNetCore/Registration/ServiceCollectionRegistrationExtensions.cs
```diff
@@ -20,9 +20,8 @@ IEnumerable<CQRSObjectMetadata> cqrsObjects
 
     public static void AddCQRSHandler(this IServiceCollection serviceCollection, CQRSObjectMetadata obj)
     {
-        serviceCollection.Add(
-            new ServiceDescriptor(MakeHandlerInterfaceType(obj), obj.HandlerType, ServiceLifetime.Scoped)
-        );
+        serviceCollection.Add(new(MakeHandlerInterfaceType(obj), obj.HandlerType, ServiceLifetime.Scoped));
+        serviceCollection.Add(new(obj.HandlerType, obj.HandlerType, ServiceLifetime.Scoped));
```
marcin-dardzinski, line 23: DI quirks but maybe make interface registration resolve concrete handler instead of creating a separate instance?

marcin-dardzinski (approved):
```
serviceCollection
    .AddCQRS(TypesAssembly.Of<Command>(), TypesAssembly.Of<CH1>())
    .AddCQRSObjects(TypesAssembly.Of<Command>(), TypesAssembly.Of<CH2>());
```
Interesting corner case, shouldn't `CQRSObjectsRegistrationSource` throw on such cases. If I register command with two handlers there's clearly sth wrong.

src/CQRS/LeanCode.CQRS.AspNetCore/Registration/ServiceCollectionRegistrationExtensions.cs
jakubfijalkowski: I don't really understand that. Nevertheless, the problem is that there can be multiple `Add` for `ICommandHandler<Command>`, and when you resolve, you will get the last one.

jakubfijalkowski:
> Interesting corner case, shouldn't CQRSObjectsRegistrationSource throw on such cases.

It probably should, but it won't solve the problem - the user will still be able to overcome that with two calls to the `AddCQRS` or manual `AddTransient<ICommandHandler<Command>, CH2>()`.

src/CQRS/LeanCode.CQRS.AspNetCore/Registration/ServiceCollectionRegistrationExtensions.cs
marcin-dardzinski: I mean sth like
```
serviceCollection.Add(new(obj.HandlerType, obj.HandlerType, ServiceLifetime.Scoped));
serviceCollection.Add(new(MakeHandlerInterfaceType(obj), sp => sp.GetService(obj.HandlerType)));
```
So that if you ask for interface and concrete type in one scope you'll get the same instance, but as I said, it's just a quirk.

jakubfijalkowski: Ah, got it. That indeed might be better.

jakubfijalkowski: Done.

PR #640, opened 2024-02-06
Unify object and endpoint metadata

jakubfijalkowski:
They describe basically the same thing and I couldn't find a reason to have them separate. Having the executor inside object metadata makes local execution _much_ easier.

The main problem with this one - it is a breaking change, as it removes a public type. Thus, it probably should end up in v8.1.


marcin-dardzinski (approved):
Yeah, `8.1` probably

PR #644, opened 2024-02-09
Update packages and reformat code with newer CSharpier

jakubfijalkowski:
This is mostly a grand reformat, nothing interesting


Saancreed (approved):
You could have split package updates from CSharpier bump and reformatting as separate PRs, in which case the former would be more attractive for reviewers and the latter would be self-approve material.

PR #681, opened 2024-07-11
Switch from Ory.Kratos.Client to LeanCode.Kratos.Client

Saancreed:
~~This won't build successfully until we actually push the package but works on my machine™.~~

The registration and usage changed a bit but it's still the closest we have to old Newtonsoft-based client.


src/Infrastructure/LeanCode.Kratos/KratosAuthenticationHandler.cs
```diff
@@ -42,21 +41,22 @@ protected override async Task<AuthenticateResult> HandleAuthenticateAsync()
...
-            var session = await GetSessionAsync();
+            var response = await GetSessionAsync();
+            var session = response?.Ok();
 
             if (session is null)
             {
                 return AuthenticateResult.NoResult();
             }
-            else if (!session.Active)
+            else if (session.Active != true)
```
Saancreed, line 51: In case you're wondering, this is nullable now.

src/Infrastructure/LeanCode.Kratos/KratosAuthenticationHandler.cs
```diff
@@ -76,49 +76,49 @@ protected override async Task<AuthenticateResult> HandleAuthenticateAsync()
         }
     }
 
-    protected virtual Task<KratosSession?> GetSessionAsync()
+    protected virtual async Task<IToSessionApiResponse?> GetSessionAsync()
     {
         if (Request.Headers.TryGetValue("X-Session-Token", out var token) && !string.IsNullOrEmpty(token))
         {
-            return api.ToSessionAsync(xSessionToken: token);
+            return await api.ToSessionAsync(xSessionToken: token.ToString());
```
Saancreed, line 83: Before, this was a simple `string` parameter so implicit conversion from `StringValues` worked. Now it's `Option<string>` so it's not so easy anymore.

src/Infrastructure/LeanCode.Kratos/KratosAuthenticationHandler.cs
```diff
@@ -76,49 +76,49 @@ protected override async Task<AuthenticateResult> HandleAuthenticateAsync()
         }
     }
 
-    protected virtual Task<KratosSession?> GetSessionAsync()
+    protected virtual async Task<IToSessionApiResponse?> GetSessionAsync()
```
Saancreed, line 79: The generated result type is now an interface that also exposes some additional info about whether the call succeeded or not.

src/Infrastructure/LeanCode.Kratos/KratosAuthenticationHandler.cs
```diff
@@ -76,49 +76,49 @@ protected override async Task<AuthenticateResult> HandleAuthenticateAsync()
...
     {
         var claims = new List<Claim>
         {
-            new(Options.NameClaimType, session.Identity.Id),
-            new("iss", api.Configuration.BasePath),
-            new("iat", ToUnixTimeSecondsString(session.IssuedAt)),
-            new("exp", ToUnixTimeSecondsString(session.ExpiresAt)),
-            new("auth_time", ToUnixTimeSecondsString(session.AuthenticatedAt)),
+            new(Options.NameClaimType, session.Identity!.Id),
+            new("iss", SchemaUrlPathRegex().Replace(session.Identity.SchemaUrl, string.Empty)),
```
Saancreed, line 110: We're now chopping off the schema URL path from the response, which should result in the base URL under which Kratos is publically accessible.

test/Infrastructure/LeanCode.Kratos.Tests/KratosAuthenticationHandlerTests.cs
```diff
@@ -191,14 +227,15 @@ public async Task Maps_session_and_identity_properties_to_principals_claims()
                     c.Add(new(o.RoleClaimType, "developer"));
                 }
 
-                if (s.Identity.MetadataPublic is MetadataPublic { IsAdmin: true })
+                if (s.Identity.MetadataPublic is Option<object> { Value: MetadataPublic { IsAdmin: true } })
```
Saancreed, line 230: This is quite disgusting, we now have to match against the `Option<object>` intermediate thingy.

jakubfijalkowski (approved):

Dragemil (approved):
Seems fine, doesn't seem like an improvement from this side, but maybe the generated client itself will be more convenient for server app usage.

Dragemil (approved):

PR #686, opened 2024-08-12
Bump MassTransit and fix missing method

sivvak:
`HasConsumerInterfaces` method got deleted :see_no_evil:
Using `IsConsumer` instead seems reasonable :smiley:

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -5,7 +5,7 @@
     <ExtensionsVersion>8.0.0</ExtensionsVersion>
 
     <IdentityServerVersion>4.1.2</IdentityServerVersion>
-    <MassTransitVersion>8.1.3</MassTransitVersion>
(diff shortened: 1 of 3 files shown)


Dragemil (approved):
I'm doing it like this in my project and it seems to be fine.

sivvak:
I changed the way Postgres DB gets configured due to the [problems](https://github.com/leancodepl/corelibrary/actions/runs/10351127624/job/28649011961) occuring after MassTransit bump. After some research I discovered potential source of the mentioned problems described on [SO](https://stackoverflow.com/questions/78048420/masstransit-entity-framework-core-more-than-twenty-iserviceprovider-insta). It looks like everything is working fine but please correct me if I am wrong :sweat_smile:

Saancreed (approved):

PR #692, opened 2024-09-02
Retarget analysers from .NETStandard 2.1 to 2.0

Patonymous:
Targeting .NET Standard 2.0 allows us to support `msbuild` used in Visual Studio.

I've attempted not to modify the analysers source code and instead provided implementations of .NET System Corelib functionalities which were missing. Taking them from official implementation wherever possible and providing my own were it was simpler.


src/Tools/LeanCode.CodeAnalysis/NetStandard21Compatibility/System.Diagnostics.CodeAnalysis.NullableAttributes.cs
jakubfijalkowski, line 1: If we target netstandard2.0 in this lib, does this file is ever compiled with `NETSTANDARD2_1`? Do these attributes work?

src/Tools/LeanCode.CodeAnalysis/NetStandard21Compatibility/System.Index.cs
jakubfijalkowski, line 1: This is a heavy dependency and I wonder if it is worth it. ~What would be needed if we remove this and `Range` (roughly)?~

src/Tools/LeanCode.CodeAnalysis/NetStandard21Compatibility/System.Range.cs
```diff
@@ -0,0 +1,133 @@
...
+    public bool Equals(Range other) => other.Start.Equals(Start) && other.End.Equals(End);
+
+    /// <summary>Returns the hash code for this instance.</summary>
+    public override int GetHashCode()
+    {
+#if (!NETSTANDARD2_0 && !NETFRAMEWORK)
+        return HashCode.Combine(Start.GetHashCode(), End.GetHashCode());
+#else
+        return HashHelpers.Combine(Start.GetHashCode(), End.GetHashCode());
+#endif
```
Saancreed: Weird, https://learn.microsoft.com/en-us/dotnet/api/system.hashcode.combine?view=net-8.0#system-hashcode-combine-2(-0-1) suggests that we should be able to use it even in netstandard2.0 projects, in which case this becomes simply

```suggestion
        return HashCode.Combine(Start, End);
```

and lets you get rid of `HashHelpers` altogether.

(But you might need to add a dep on https://www.nuget.org/packages/Microsoft.Bcl.HashCode)

jakubfijalkowski (changes requested):
Okay, I checked what would be needed if we don't want to import the `Index` and `Range` and not include the `NotNull` attributes, and I conclude that it is _not worth it_. This will be literally 8 line change, whereas we import almost 500 lines of code to make the 8 work. Not worth it.

src/Tools/LeanCode.CodeAnalysis/NetStandard21Compatibility/System.Index.cs
Saancreed: I suggest we just add a reference to `Microsoft.Bcl.Memory` and that would be it.

(I also just learned that this package exists because it's fairly new.)

src/Tools/LeanCode.CodeAnalysis/NetStandard21Compatibility/MissingFileMethods.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            throw new ArgumentException("Path is invalid");
+        }
+
+        using var stream = File.OpenWrite(path);
+        var preamble = encoding.GetPreamble();
+        await stream.WriteAsync(preamble, 0, preamble.Length, cancellationToken);
+        if (contents is not null)
+        {
+            var encoded = Encoding.Default.GetBytes(contents);
```
jakubfijalkowski: Shouldn't this be `encoding.GetBytes`, since we write `encoding` preamble?

src/Tools/LeanCode.CodeAnalysis/NetStandard21Compatibility/System.Index.cs
jakubfijalkowski: Or do that, that would be okay, provided that the package would not be in preview (although that is still much better than vendoring these).

jakubfijalkowski: Thinking about it more - I fear that re-defining the types or importing them from Bcl-polyfills package would be problematic. MSBuild loads the analyzers in-proc, thus it will fail if types are re-imported, or if there is a need to load two versions of the same library. :(

Saancreed: It should leave the preview stage next week (or at worst next month) but I'm not sure if we should even care about it, we are using the part of it that I'd expect be stable by now.

About multiple versions: if other analyzers having their own deps is not a problem already then this one shouldn't be either. But let's give it a try and see if anything breaks :slightly_smiling_face:

Saancreed:
> Okay, I checked what would be needed if we don't want to import the `Index` and `Range` and not include the `NotNull` attributes, and I conclude that it is _not worth it_. This will be literally 8 line change, whereas we import almost 500 lines of code to make the 8 work. Not worth it.

I'd try to import `Index` and `Range` from standalone BCL compatibility packages, and the reconsider if nullables are worth it or not. But for them I have no strong preference so we could just get rid of nullables if you think that would make it less clunky.

src/Tools/LeanCode.CodeAnalysis/NetStandard21Compatibility/System.Index.cs
jakubfijalkowski: The main problem with
> if other analyzers having their own deps is not a problem already
that I have is  - we don't know that. Noone uses VS, we tend not to use too many analyzers. :P

However, let's indeed test that case - i.e. let's test if two analyzers with conflicting libraries work.

src/Tools/LeanCode.CodeAnalysis/NetStandard21Compatibility/System.Diagnostics.CodeAnalysis.NullableAttributes.cs
jakubfijalkowski: My question was slightly different - does having these attributes help with anything? Analyzers use them in two places, both of which are only there to make the compiler happy. I would say we should remove this file altogether as it is not worth it.

Patonymous: Yes, they work, because the guard was `#if !NETSTANDARD2_1`.

I didn't want to modify the vendored files more that it was required to handle all the warnings, but now I've decided to removed conditional compilation for simplicity's sake. Unless you'd prefer it to be there

src/Tools/LeanCode.CodeAnalysis/NetStandard21Compatibility/System.Index.cs
Patonymous: I've added the `Microsoft.Bcl.Memory` package reference and will test if it works with `msbuild`

src/Tools/LeanCode.CodeAnalysis/NetStandard21Compatibility/System.Diagnostics.CodeAnalysis.NullableAttributes.cs
Patonymous: > I would say we should remove this file altogether as it is not worth it.

Okay, I will do that. I just though we valued nullability analysis enough to keep it.

jakubfijalkowski: > I just though we valued nullability analysis enough to keep it.

Considering that it was used in two places, it is not worth it. ;P

src/Tools/LeanCode.CodeAnalysis/NetStandard21Compatibility/System.Index.cs
jakubfijalkowski: Remember to test the case of two analyzers with conflicting references, e.g. one uses version X and the other version Y and both are used in a single project. I am pretty sure that this particular package won't be a problem _now_, but it might be when the next version comes up and other analyzers decide to use it.

src/Tools/LeanCode.CodeAnalysis/NetStandard21Compatibility/MissingFileMethods.cs
```diff
@@ -0,0 +1,42 @@
+// Based on https://github.com/dotnet/runtime/blob/5535e31a712343a63f5d7d796cd874e563e5ac14/src/libraries/System.Private.CoreLib/src/System/IO/File.cs
+// as linked by docs on 02.09.2024
+
+using System.Text;
+
+namespace System.IO;
+
+public static class MissingFileMethods
```
Saancreed: One alternative to consider for this would be just

```suggestion
namespace LeanCode.CodeAnalysis;

public static class File
{
    public static void Delete(string path) => global::System.IO.File.Delete(path);
```

which would get picked up instead of `System.IO.Path` where appropriate and avoid the need to call methods here by current (quite awkward) class name.

Saancreed (approved):
I have one more random idea to consider but current revision looks okay too.

src/Tools/LeanCode.CodeAnalysis/NetStandard21Compatibility/MissingFileMethods.cs
Patonymous: If I put it in `LeanCode.CodeAnalysis` namespace it gets picked up in `LeanCode.CodeAnalysis.Tests`, which make use of `File.ReadAllTextAsync`, so I would have to also implement that or add a `using` alias there.
It currently works, because tests still target .NET8.

Patonymous: What do you think of a `global using` alias in `LeanCode.CodeAnalysis`?
It will only be picked up in this project.

PR #700, opened 2024-09-11
Improve tracing across CQRS elements

jakubfijalkowski:
- **Add methods to start activities by convention**
- **Add tracing to common CQRS middlewares**
- **Add entry to CHANGELOG**


marcin-dardzinski (approved):

Dragemil (approved):
Do we set error on the span when we catch exceptions in our middleware or maybe it does happen via aspnet core internals?

jakubfijalkowski:
If the exception is not handled by any of ours (exception translation, mostly), it will be automatically set by the ASP.NET on the HTTP request span.

PR #710, opened 2024-10-02
Allow setting custom type formatters in FCM notifications

### LINKED ISSUE #662: FCM `ToNotificationData` helper may cause data loss with some types
Although documented only to be unable to handle hierarchical/nested data, current implementation of `ToNotificationData` may cause data to be lost during conversion when some specific types are used, e.g. `DateTimeOffset`s are converted to strings like `"04/04/2024 12:34:56 +00:00"` instead of ISO 8601 format, which among other things causes milliseconds to be discarded.

(Also current implementation fails to handle enums that don't fit in `Int32` range and just explodes with a sad exception.)

marcin-dardzinski:
Just a proposal to fix #662. Clients are able to override how types are serialized. 

I didn't provide any defaults, especially for dates as it might be a breaking change for the apps

### DIFF
--- a/src/Infrastructure/LeanCode.Firebase.FCM/NotificationDataConverter.cs
+++ b/src/Infrastructure/LeanCode.Firebase.FCM/NotificationDataConverter.cs
@@ -1,0 +1,58 @@
+using System.Collections.Immutable;
+using System.Reflection;
+
+namespace LeanCode.Firebase.FCM;
+
+public sealed class NotificationDataConverter
+{
+    private const string TypeField = "Type";
+
+    private ImmutableDictionary<Type, Func<object, string>> formatters = ImmutableDictionary<
+        Type,
+        Func<object, string>
+    >.Empty;
+
+    public void FormatUsing<T>(Func<T, string> formatter)
+    {
+        formatters = formatters.Add(typeof(T), o => formatter((T)o));
+    }
+
+    /// <summary>
+    /// Converts POCO object to a notification data dictionary. Does not support hierarchical
+    /// data.
+    /// </summary>
+    public Dictionary<string, string> ToNotificationData(object data)
+    {
+        var type = data.GetType();
+
+        var result = new Dictionary<string, string>() { [TypeField] = type.Name, };
+
+        foreach (var prop in type.GetProperties(BindingFlags.Public | BindingFlags.Instance))
+        {
+            if (prop.Name != TypeField)
+            {
+                var value = prop.GetValue(data);
+
+                if (value is not null)
+                {
+                    var formatter = formatters.GetValueOrDefault(value.GetType(), DefaultValueFormatter);
+                    result.Add(prop.Name, formatter(value));
+                }
+            }
+        }
+
+        return result;
+    }
+
+    private static string DefaultValueFormatter(object value)
+    {
+        if (value is Enum @enum)
+        {
+            return @enum.ToString("D");
+        }
+        else
+        {
+            return value.ToString()!;
+        }
+    }
+}
--- a/src/Infrastructure/LeanCode.Firebase.FCM/NotificationDataConverter.cs
+++ b/src/Infrastructure/LeanCode.Firebase.FCM/NotificationDataConverter.cs
@@ -1,20 +1,16 @@
-using System.Collections.Immutable;
+using System.Collections.Frozen;
 using System.Reflection;
 
 namespace LeanCode.Firebase.FCM;
 
 public sealed class NotificationDataConverter
 {
+    private readonly FrozenDictionary<Type, Func<object, string>> formatters;
     private const string TypeField = "Type";
 
-    private ImmutableDictionary<Type, Func<object, string>> formatters = ImmutableDictionary<
-        Type,
-        Func<object, string>
-    >.Empty;
-
-    public void FormatUsing<T>(Func<T, string> formatter)
+    private NotificationDataConverter(FrozenDictionary<Type, Func<object, string>> formatters)
     {
-        formatters = formatters.Add(typeof(T), o => formatter((T)o));
+        this.formatters = formatters;
     }
 
     /// <summary>
@@ -35,13 +31,21 @@ public sealed class NotificationDataConverter
 
                 if (value is not null)
                 {
-                    var formatter = formatters.GetValueOrDefault(value.GetType(), DefaultValueFormatter);
+                    var formatter = formatters.GetValueOrDefault(GetEffectiveType(value), DefaultValueFormatter);
                     result.Add(prop.Name, formatter(value));
                 }
             }
         }
 
         return result;
+    }
+
+    public static Builder New() => new();
+
+    private static Type GetEffectiveType(object value)
+    {
+        var type = value.GetType();
+        return Nullable.GetUnderlyingType(value.GetType()) ?? type;
     }
 
     private static string DefaultValueFormatter(object value)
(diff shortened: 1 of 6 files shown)


jakubfijalkowski:
I fear this global config. Maybe we should make the class non-static and have the formatters not-global?

marcin-dardzinski:
Wait, who is calling the class? Corelib code or clients directly? 
But yes, we can make it non static and keep some static obsoleted fallback, though I don't see any use cases other than singleton config on app startup.

jakubfijalkowski:
Clients directly, they are responsible for making the underlying message.

Yeah, it will be a singleton most of the time, but it at least be testable. ;P

jakubfijalkowski (approved):

src/Infrastructure/LeanCode.Firebase.FCM/NotificationDataConverter.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        foreach (var prop in type.GetProperties(BindingFlags.Public | BindingFlags.Instance))
+        {
+            if (prop.Name != TypeField)
+            {
+                var value = prop.GetValue(data);
+
+                if (value is not null)
+                {
+                    var formatter = formatters.GetValueOrDefault(value.GetType(), DefaultValueFormatter);
```
Saancreed: I imagine this isn't smart enough to let me handle both `DateTimeOffset` and `DateTimeOffset?` with a single formatter?

src/Infrastructure/LeanCode.Firebase.FCM/NotificationDataConverter.cs
```diff
@@ -0,0 +1,58 @@
...
+public sealed class NotificationDataConverter
+{
+    private const string TypeField = "Type";
+
+    private ImmutableDictionary<Type, Func<object, string>> formatters = ImmutableDictionary<
+        Type,
+        Func<object, string>
+    >.Empty;
+
+    public void FormatUsing<T>(Func<T, string> formatter)
```
Saancreed: I was hoping we would require all formatters to be provided at construction time and `NotificationDataConverter` could then be immutable. Can we do this, using a builder pattern or something?

src/Infrastructure/LeanCode.Firebase.FCM/NotificationDataConverter.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        formatters = formatters.Add(typeof(T), o => formatter((T)o));
+    }
+
+    /// <summary>
+    /// Converts POCO object to a notification data dictionary. Does not support hierarchical
+    /// data.
+    /// </summary>
+    public Dictionary<string, string> ToNotificationData(object data)
```
Saancreed, line 20: Random question and unrelated to problem at hand but why isn't this generic?

```suggestion
    public Dictionary<string, string> ToNotificationData<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties)] T>(T data)
```

jakubfijalkowski: What will we gain by making it generic?

Saancreed: Avoids boxing with value types, one less reflection to fix before AOT complilation works, etc.

Not that they would be some huge benefits but I'm just curious why was the initial decision made to force conversion to `object`. What did we gain by not making it generic?

src/Infrastructure/LeanCode.Firebase.FCM/NotificationDataConverter.cs
jakubfijalkowski: I would not go that way - let's make this class simple. We should re-model how we send pushes  either way (and it seems it will be sooner than later because of other requirements).

src/Infrastructure/LeanCode.Firebase.FCM/NotificationDataConverter.cs
jakubfijalkowski: +1 from me

src/Infrastructure/LeanCode.Firebase.FCM/NotificationDataConverter.cs
marcin-dardzinski: Good catch, will see

src/Infrastructure/LeanCode.Firebase.FCM/NotificationDataConverter.cs
marcin-dardzinski: Any value type will be boxed anyway with `prop.GetValue(data)` call. Btw, this change will make boxing happen per every property, won't it? 
I see no generic gains here.

src/Infrastructure/LeanCode.Firebase.FCM/NotificationDataConverter.cs
marcin-dardzinski: A great overkill but can be done.

src/Infrastructure/LeanCode.Firebase.FCM/NotificationDataConverter.cs
Saancreed: Remodeling would be nice indeed, but while we're still on this converter requiring users of this class to always register two formatters for nullable value types is error-prone. And fixing this is just one

```csharp
if (type.IsConstructedGenericType && type.GetGenericTypeDefinition() == typeof(Nullable<>))
{
    type = type.GenericTypeArguments.Single();
}
```

away.

src/Infrastructure/LeanCode.Firebase.FCM/NotificationDataConverter.cs
Saancreed: > Any value type will be boxed anyway with `prop.GetValue(data)` call.

Ehh, you're right, I totally forgot about this.

src/Infrastructure/LeanCode.Firebase.FCM/NotificationDataConverter.cs
Saancreed: Plus, if it's assumed to be a singleton you could then use `FrozenDictionary` for storage.

src/Infrastructure/LeanCode.Firebase.FCM/NotificationDataConverter.cs
jakubfijalkowski: Yeah, I think you're right. We can go with this I think.

test/Infrastructure/LeanCode.Firebase.FCM.Tests/LeanCode.Firebase.FCM.Tests.csproj
```diff
@@ -1,5 +1,9 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
+    <PropertyGroup>
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
```
Saancreed: Indentation doesn't match the rest of the file :stuck_out_tongue:

test/Infrastructure/LeanCode.Firebase.FCM.Tests/LeanCode.Firebase.FCM.Tests.csproj
```diff
@@ -1,5 +1,9 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
+    <PropertyGroup>
```
jakubfijalkowski: Hehe, indent :P

Saancreed (approved):

PR #724, opened 2024-12-19
Address vulnerabilities and switch to Central Package Management

jakubfijalkowski:
Plus remove StyleCop project, as it should not be there in the first place (and it makes CPM harder to implement ;P).


Directory.Packages.props
```diff
@@ -0,0 +1,141 @@
+<Project>
+  <PropertyGroup>
+    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+
+    <EntityFrameworkCoreVersion>9.0.0</EntityFrameworkCoreVersion>
+    <AspNetCoreVersion>9.0.0</AspNetCoreVersion>
+    <ExtensionsVersion>9.0.0</ExtensionsVersion>
+
+    <IdentityServerVersion>4.1.2</IdentityServerVersion>
+    <MassTransitVersion>8.3.2</MassTransitVersion>
```
Dragemil, line 10: PhatBoyG once wrote somewhere that this version is problematic, but I don't remember why, so we should consider bumping it on some other iteration.

Dragemil (approved):

Directory.Packages.props
jakubfijalkowski: I'll bump the packages once I migrate to xUnit v3

lukaszgarstecki (approved):

PR #725, opened 2024-12-20
Migrate to xUnit 3

jakubfijalkowski:


.github/workflows/core_build_publish.yml
```diff
@@ -90,6 +90,7 @@ jobs:
       - name: Tool Restore
         run: dotnet tool restore
       - name: Check formatting
+        if: false
```
lukaszgarstecki, line 93: Why?

.github/workflows/core_build_publish.yml
```diff
@@ -152,44 +151,40 @@ jobs:
           CORELIB_TESTS_AZURE_BLOB_STORAGE_CONTAINER_NAME: ${{ secrets.CORELIB_TESTS_AZURE_BLOB_STORAGE_CONTAINER_NAME }}
           CORELIB_TESTS_AZURE_TABLE_STORAGE_TABLE_NAME: ${{ secrets.CORELIB_TESTS_AZURE_TABLE_STORAGE_TABLE_NAME }}
       - name: Integration Test - SQL Server
-        run: dotnet test -c Release --logger trx --results-directory ../TestResults
-        working-directory: test/LeanCode.IntegrationTests
+        run: dotnet test -- --explicit on --filter-query '/[(category=integration)&(database!=postgres)]'
```
lukaszgarstecki, line 154: why do you filter here by `postgres` and not in the `Integration Test - Postgres`?

.github/workflows/core_build_publish.yml
```diff
@@ -152,44 +151,40 @@ jobs:
...
           LeanCodeIntegrationTests__Database: postgres
-      - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v3
-        if: always()
 
-      - name: Convert Test Results to jUnit
-        run: find -name '*.trx' -exec trx2junit --output TestResults {} +
-        working-directory: test
+      - name: Process test results
         if: always()
```
lukaszgarstecki, line 165: Isn't it the same as without `if:`?

test/LeanCode.IntegrationTests/PushNotificationTokenStoreTests.cs
```diff
@@ -18,7 +18,7 @@ public PushNotificationTokenStoreTests()
         app = new TestApp();
     }
 
-    [Fact]
+    [IntegrationFact]
```
marcin-dardzinski, line 21: I guess we need to adjust docs on how to run those tests.

.editorconfig
```diff
@@ -193,3 +193,6 @@ dotnet_diagnostic.CA1812.severity = none
 dotnet_diagnostic.CA1724.severity = none
 dotnet_diagnostic.CA1515.severity = none
 dotnet_diagnostic.IDE0058.severity = none
+
+# Disable temporarily
+dotnet_diagnostic.xUnit1051.severity = none
```
Dragemil, line 198: Do you plan to tackle those CTs in another PR?

lukaszgarstecki (approved):

.github/workflows/core_build_publish.yml
jakubfijalkowski: Nope, the default for `if` is "if previous step succeeded", not `always`.

test/LeanCode.Test.Helpers/IntegrationFactAttribute.cs
```diff
@@ -0,0 +1,16 @@
...
+[AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
+public class IntegrationFactAttribute : FactAttribute, ITraitAttribute
+{
+    public IntegrationFactAttribute()
+    {
+        Explicit = true;
+    }
+
+    public virtual IReadOnlyCollection<KeyValuePair<string, string>> GetTraits() =>
+        [new("category", "integration"), new("integration", "true")];
```
marcin-dardzinski, line 15: Why two tags? Shouldn't `category=integration` suffice?

.github/workflows/core_build_publish.yml
jakubfijalkowski: The version used by the project does not work with .NET 9, thus I disabled it temporarily (fixed in next PR).

.github/workflows/core_build_publish.yml
jakubfijalkowski: We support two databases and we have more tests for Postgres than for SQL Server. Here, I filter-out Postgres-specific tests (there's inequality operator).

test/LeanCode.IntegrationTests/PushNotificationTokenStoreTests.cs
jakubfijalkowski: True, will do.

test/LeanCode.Test.Helpers/IntegrationFactAttribute.cs
jakubfijalkowski: Unfortunately, no.

The main culprit is the "explicit" tests. If not filtered out, they are reported as "pending." Thus, the CI reported the incorrect number of tests (because the explicit ones were counted as both "skipped in unit" and "executed in integration"). I didn't want to remove the explicit tag because it allows for a plain `dotnet test` run when running locally.

I went with an additional tag. "Category" was already used for one form of categorization (integration, external, long-running) and I needed another tag that distinguishes "integration"-style tests from unit tests.

I probably should have used better names but TBH I don't have any sensible counterproposals.

test/LeanCode.Test.Helpers/ExternalServiceFactAttribute.cs
```diff
@@ -0,0 +1,39 @@
...
+
+    public void After(MethodInfo methodUnderTest, IXunitTest test) { }
+
+    public void Before(MethodInfo methodUnderTest, IXunitTest test)
+    {
+        var skipReason = SkipIfVariablesNotSet(RequiredEnvVariables.ToHashSet());
+
+        if (skipReason != null)
+        {
+            Assert.Fail(skipReason);
```
Dragemil, line 29: Do we really prefer to fail here rather than actually skip?

test/LeanCode.Test.Helpers/LongRunningFact.cs
```diff
@@ -1,19 +1,16 @@
-using System;
 using Xunit;
```
Dragemil, line 1: Is that namespace even still supported by the v3 version? Don't you have some older xunit creeped in somewhere, that it still compiles?

test/LeanCode.IntegrationTests/PushNotificationTokenStoreTests.cs
jakubfijalkowski: I think I will open another PR with updates to the README, because there is more needed.

.github/workflows/core_build_publish.yml
```diff
@@ -132,16 +133,14 @@ jobs:
...
       - name: Restore
-        run: dotnet restore
-      - name: Test
-        run: dotnet msbuild /t:RunTestsWithCoverage
-        working-directory: test
+        run: |
+          dotnet restore
+          dotnet tool restore
+      - name: Unit tests
+        run: dotnet test -- --explicit on --filter-not-trait 'integration=true'
```
Dragemil, line 141: Do we have explicit tests anywhere or is it some kind of future proofing?

test/LeanCode.Test.Helpers/LongRunningFact.cs
jakubfijalkowski: Yes, the [`FactAttribute`](https://github.com/xunit/xunit/blob/main/src/xunit.v3.core/FactAttribute.cs#L6) is still in `Xunit` namespace. :)

test/LeanCode.Test.Helpers/ExternalServiceFactAttribute.cs
jakubfijalkowski: Yes. You now have to explicitly run them (i.e. pass `--explicit on` to the runner), so failing is IMO more sensible. My thinking here is "you already did something to run them, so let's run them".

.github/workflows/core_build_publish.yml
jakubfijalkowski: All the integration and "external service"-integration tests are explicit.

.editorconfig
jakubfijalkowski: Yes, we have a task for that in the backlog.

test/LeanCode.Test.Helpers/ExternalServiceFactAttribute.cs
Dragemil: Makes sense.

Dragemil (approved):

PR #733, opened 2024-12-30
Remove unused projects

jakubfijalkowski:
The three removed projects should either be migrated to newer versions of the base library (i.e.
migrate to Duende IS) or were not updated in eons (PdfRocket) and should be rewritten. :)


lukaszgarstecki (approved):

lukaszgarstecki:
IMO dropping LeanCode.Mixpanel would be nice. @Saancreed WDYT?

Saancreed:
:shrug:

Whatever, more stuff for me to vendor or delete.

Just when I devendored ExternalIdentityProviders :weary:

jakubfijalkowski:
It stays then. IS4-related things are removed because IS4 is even delisted on Nuget, so it makes more sense (and simplifies version mgmt a lot).

PR #746, opened 2025-01-16
Integration test config rework

jakubfijalkowski:
Two (three) small changes that should make the integration test helpers easier to use:
1. `DbContextInitializer` works with Npgsql now,
2. `ConfigurationOverrides` allows specifying arbitrary config strings now.

Plus, a test connection string management is now in a separate class that loads the connection
string base not from env vars, but from the supplied configuration builder itself. Allows, e.g., to
specify the base connection stirng in `appsettings.json` file.


test/Testing/LeanCode.IntegrationTestHelpers.Tests/TestApp.cs
```diff
@@ -7,13 +7,17 @@
...
+using Serilog.Events;
 
 namespace LeanCode.IntegrationTestHelpers.Tests;
 
 public class TestApp : LeanCodeTestFactory<App.Startup>
 {
-    protected override ConfigurationOverrides Configuration { get; } =
-        new("SqlServer__ConnectionStringBase", "SqlServer:ConnectionString", Serilog.Events.LogEventLevel.Error, false);
+    protected override TestConnectionString ConnectionStringConfig { get; } =
+        new("SqlServer__ConnectionStringBase", "SqlServer:ConnectionString");
```
lukaszgarstecki: Don't we need the same test for postgres?

jakubfijalkowski: This test suite tests if the helpers themselves work. They don't test integration with any particular database (they need _a_ database, but it is not important).

src/Testing/LeanCode.IntegrationTestHelpers/TestConnectionString.cs
```diff
@@ -0,0 +1,48 @@
...
+        string connectionStringKey,
+        IConfiguration parentConfig
+    ) : ConfigurationProvider
+    {
+        public override void Load()
+        {
+            var baseConnStr =
+                parentConfig[connectionStringBaseKey]
+                ?? throw new KeyNotFoundException(
+                    $"Cannot find base connection stirng under key {connectionStringBaseKey}"
```
Dragemil: typo

Dragemil (approved):

test/LeanCode.IntegrationTests/TestDatabaseConfig.cs
```diff
@@ -12,7 +12,7 @@ public abstract class TestDatabaseConfig
 {
     public const string ConfigEnvName = "LeanCodeIntegrationTests__Database";
 
-    public abstract ConfigurationOverrides GetConfigurationOverrides();
+    public abstract TestConnectionString GetConfigurationOverrides();
```
Saancreed:
```suggestion
    public abstract TestConnectionString GetConnectionString();
```

Name doesn't feel appropriate for what it returns now.

jakubfijalkowski: True, will fix

Saancreed (approved):

PR #754, opened 2025-02-03
[CORELIB-75] DI injected logger

wojtek2288:


wojtek2288:
The actual goal of this task was to use DI injected `ILogger<T>` in corelib - closing this one

PR #757, opened 2025-02-07
[CORELIB-75] DI injected logger

wojtek2288:


test/Infrastructure/LeanCode.Kratos.Tests/KratosWebHookHandlerTests.cs
```diff
@@ -144,20 +154,20 @@ Dictionary<string, StringValues> headers
         return httpContext;
     }
 
-    private sealed class KratosWebHookTestHandler : KratosWebHookHandlerBase
+    public sealed class KratosWebHookTestHandler : KratosWebHookHandlerBase
```
wojtek2288, line 157: It's either public or we add [assembly: InternalsVisibleTo("DynamicProxyGenAssembly2")] to project for NSubstitute

jakubfijalkowski:
Let's also discuss this internally in a wider group.

src/CQRS/LeanCode.CQRS.AspNetCore/Middleware/CQRSExceptionTranslationMiddleware.cs
```diff
@@ -40,7 +45,7 @@ public async Task InvokeAsync(HttpContext httpContext)
             activity?.SetTag("error.code", ex.ErrorCode);
 
             var result = WrapInCommandResult(ex);
-            logger.Warning("Command {@Command} is not valid with result {@Result}", cqrsPayload.Payload, result);
+            logger.LogWarning("Command {@Command} is not valid with result {@Result}", cqrsPayload.Payload, result);
```
marcin-dardzinski: Is `{@Command}` syntax an extensions logging standard? I've always though it was a serilog goodie (chat gpt confirms it). So, I sb ever want's to change logger implementation I guess it'd break.

wojtek2288: Hm, right in some loggers `ToString()` might be called here. We probably need to change `@Command` to `Command` and pass `JsonSerializer.Serialize(cqrsPayload.Payload)` as an argument to make it work with any logger.

jakubfijalkowski: I don't think this is a good approach. It will break the "sanitizer" logic that we have and we need to support that.

wojtek2288: Then I'm not sure if we can do anything about it right now other then making sanitization logger agnostic.

Dragemil: So if we'd abstract this way from Serilog we'd lose this functionality, which would be sad. Is there a way to not abstract the Serilog, and still use the DI based loggers? If so, I'd say let's go with it (unless it isn't too ugly), and in the other case I'd stick with what we've done until now.

marcin-dardzinski: If serilog does not provide anything out of the box, I think we can safely register `Serilog.ILogger<T>` as open generic and link it with DI

wojtek2288: Serilog does not provide generic `Serilog.ILogger<T>`. We can use plain `Serilog.ILogger` instead and enrich it with context in constructor using `logger.ForContext<T>()` on DI injected `logger`.

jakubfijalkowski: This will result in a worse experience than we have now, we need to find a way (custom `ILogger`?) to preserve the developer experience.

src/Infrastructure/LeanCode.Logging/IHostBuilderExtensions.cs
```diff
@@ -42,52 +43,58 @@ public static IHostBuilder ConfigureDefaultLogging(
...
+                        loggerConfiguration.WriteTo.Console(formatProvider: CultureInfo.InvariantCulture);
+                    }
+                    else
+                    {
+                        loggerConfiguration.WriteTo.Console(new RenderedCompactJsonFormatter());
+                    }
+
+                    additionalLoggingConfiguration?.Invoke(context, loggerConfiguration);
+
+                    Log.Logger = loggerConfiguration.CreateLogger();
```
jakubfijalkowski: This leaves the static logger available and we want to eliminate it completely.

src/Infrastructure/LeanCode.Logging/IHostBuilderExtensions.cs
```diff
@@ -42,52 +43,58 @@ public static IHostBuilder ConfigureDefaultLogging(
         Action<HostBuilderContext, LoggerConfiguration>? additionalLoggingConfiguration = null
     )
     {
-        return builder.ConfigureLogging(
-            (context, logging) =>
-            {
-                var configuration = context.Configuration;
-                var minLogLevel = configuration.GetValue(MinimumLogLevelKey, LogEventLevel.Verbose);
+        return builder
+            .ConfigureServices(
```
jakubfijalkowski: Also, this should call `AddSerilog`

src/Infrastructure/LeanCode.Logging/IHostBuilderExtensions.cs
wojtek2288: Do we really want to remove it? This would be a breaking change and prevent us from using `Log.ForContext<T>()` in apps and I remember from corelib v8 that we wanted to minimize the number of breaking changes. Maybe it's better to keep both of them available.

jakubfijalkowski: > I remember from corelib v8 that we wanted to minimize the number of breaking changes.
This is the kind of breaking change that I am willing to do. "Reverting" that on the app side is a matter of two-three lines.

Also, we can add `preserveStaticLogger` flag (defaulting to false), just like `AddSerilog` does if we want to do.

src/CQRS/LeanCode.CQRS.AspNetCore/Middleware/CQRSMiddleware.cs
```diff
@@ -1,23 +1,29 @@
...
 
-    private readonly ILogger logger = Log.ForContext<CQRSMiddleware>();
+    private readonly ILogger<CQRSMiddleware> logger;
 
     private readonly CQRSMetrics metrics;
     private readonly ISerializer serializer;
     private readonly RequestDelegate next;
 
-    public CQRSMiddleware(CQRSMetrics metrics, ISerializer serializer, RequestDelegate next)
+    public CQRSMiddleware(
```
jakubfijalkowski, line 19: One last (not priority, not mandatory) request - let's put the logger as a last parameter (I think this is also a convention in the .NET examples, but might be wrong here).

src/Infrastructure/LeanCode.Logging/IHostBuilderExtensions.cs
```diff
@@ -83,10 +92,18 @@ public static IHostBuilder ConfigureDefaultLogging(
...
+                    Log.Logger = logger;
+                }
+
+                logging.Services.AddSingleton<Serilog.ILogger>(logger);
+                logging.Services.AddSingleton(typeof(ILogger<>), typeof(Logger<>));
 
-                logging.AddConfiguration(configuration.GetSection(SystemLoggersEntryName));
-                logging.AddSerilog();
+                logging.AddConfiguration(context.Configuration.GetSection(SystemLoggersEntryName));
+                logging.AddSerilog(logger, dispose: true);
```
jakubfijalkowski, line 106: I think I know why we haven't used [AddSerilog](https://github.com/serilog/serilog-extensions-hosting/blob/962e7214eb7dd5fb3b2e4347189baf5c9d260d9a/src/Serilog.Extensions.Hosting/SerilogServiceCollectionExtensions.cs#L129-L216) in the `ConfigureServices`, but let's be sure - why?

test/CQRS/LeanCode.CQRS.AspNetCore.Tests/Local/MiddlewaresForLocalExecutionTests.cs
```diff
@@ -72,19 +75,26 @@ public async Task Security_middleware_works()
 
     public static ILocalCommandExecutor BuildWith(Action<ICQRSApplicationBuilder> configure)
     {
-        var serviceCollection = new ServiceCollection();
-        serviceCollection.AddMetrics();
-        serviceCollection.AddScoped<IMiddlewareFactory>(sp => new MiddlewareFactory(sp));
-        serviceCollection.AddScoped<LocalHandlerMiddleware>();
-        serviceCollection.AddScoped<CQRSMetrics>();
-        serviceCollection.AddScoped<ICommandValidatorResolver, CommandValidatorResolver>();
+        var hostBuilder = Host.CreateDefaultBuilder()
```
jakubfijalkowski, line 78: This is wrong. We don't want to start host here, in any way.

test/Infrastructure/LeanCode.AuditLogs.Tests/AuditLogsIntegrationTests.cs
```diff
@@ -202,7 +204,12 @@ IRegistrationContext context
 
     internal sealed class StubAuditLogStorage : IAuditLogStorage
     {
-        private readonly ILogger logger = Log.ForContext<StubAuditLogStorage>();
+        private readonly Logging.ILogger<StubAuditLogStorage> logger;
+
+        public StubAuditLogStorage()
+        {
+            logger = Substitute.For<Logging.ILogger<StubAuditLogStorage>>();
```
jakubfijalkowski, line 211: We probably should have a `NullLogger<T>` implementation. Using NSubstitute for this gives a wrong impression that it is used anyhow.

test/Infrastructure/LeanCode.Logging.Tests/LoggerInjectionTests.cs
```diff
@@ -0,0 +1,77 @@
...
+    private static string GetLogOutput(Action log)
+    {
+        using var stringWriter = new StringWriter();
+
+        Console.SetOut(stringWriter);
+
+        log();
+
+        return stringWriter.ToString();
+    }
```
jakubfijalkowski, line 59: This will make the test output broken. We _can't_ override console stream.

jakubfijalkowski (changes requested):
A couple of changes are needed:
1) We must not start `Host` in tests that just configure `ServiceCollection`. This has much more consequences (it really starts a host).
2) We need to split how we configure the logger, because currently it is very awkward (esp. for tests). We need to leave `ConfigureDefaultLogger` as-is, but we need to introduce `AddLeanCodeLogger` that configures the `ILogger<T>` factory (+ Serilog).
3) We need `NullLogger<T>`
4) We probably should split the `LeanCode.Logging` package into two - one that has the interface (+ `NullLogger<T>`), and the other that has the configuration methods. Now, we have quite unexpected dependencies from e.g. from LeanCode.CQRS.Seciruty to Serilog.Sinks.Console.

jakubfijalkowski:
Superseded by #775

PR #758, opened 2025-02-07
Silence xUnit1051 forever

wojtek2288:


wojtek2288:
Oops, not all places

jakubfijalkowski (commented):
The more I look at it, the more I am in favor of silencing it forever. This does not add any substantial benefit (the tests are quick, so even if we need to interrupt them, doing it in-between tests is almost as good. And the readability is just bonkers - they introduce soo much clutter to already quite cluttered tests that I don't see any sensible way out.

jakubfijalkowski (approved):

PR #769, opened 2025-03-13
Open trace on periodic execution

Dragemil:
Not sure if we want to use the same `ActivitySource`. It seems like it was implicitly only for CQRS (even though nothing explicitly suggests it). We could have dedicated activity sources for various packages and have `AddLeanCodeTelemetry(...)` register all `"LeanCode.CoreLibrary"` and `"LeanCode.CoreLibrary.*"` activities.

Aaalso `AddLeanCodeTelemetry()` and `AddLeanCodeMetrics()` names are kind of outside of the norm (both should be `AddLeanCodeInstrumentation`)

### DIFF
--- a/src/Infrastructure/LeanCode.PeriodicService/PeriodicHostedService.cs
+++ b/src/Infrastructure/LeanCode.PeriodicService/PeriodicHostedService.cs
@@ -1,3 +1,4 @@
+using LeanCode.OpenTelemetry;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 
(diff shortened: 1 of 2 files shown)


jakubfijalkowski:
> It seems like it was implicitly only for CQRS (even though nothing explicitly suggests it).

It was only used there, but the idea was to have a single source.

> Aaalso `AddLeanCodeTelemetry()` and `AddLeanCodeMetrics()` names are kind of outside of the norm (both should be `AddLeanCodeInstrumentation`)

Let's change that then (in a separate PR). :)

src/Infrastructure/LeanCode.PeriodicService/PeriodicHostedService.cs
```diff
@@ -41,6 +42,7 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
     )]
     private async Task<TimeSpan> ExecuteOnceAsync(int executionNo, CancellationToken stoppingToken)
     {
+        using var activity = LeanCodeActivitySource.Start($"periodic action - {typeof(TAction).Name}");
```
jakubfijalkowski:
```suggestion
        using var activity = LeanCodeActivitySource.StartExecution("periodic action", typeof(TAction).Name);
```

Dragemil:
> Let's change that then (in a separate PR). :)

I'd obsolete the old names rather than remove them, though.

jakubfijalkowski (approved):

PR #771, opened 2025-03-13
Always return constructed type from `VersionSupportQH`

jakubfijalkowski:
The query should never return null (as it is stated in contracts), thus let's assume that the wrong version is "always newer".

### DIFF
--- a/src/Infrastructure/LeanCode.ForceUpdate/LeanCode.ForceUpdate.Services/CQRS/VersionSupportQH.cs
+++ b/src/Infrastructure/LeanCode.ForceUpdate/LeanCode.ForceUpdate.Services/CQRS/VersionSupportQH.cs
@@ -5,7 +5,7 @@ using VersionSupport = LeanCode.ForceUpdate.Contracts.VersionSupport;
 
 namespace LeanCode.ForceUpdate.Services.CQRS;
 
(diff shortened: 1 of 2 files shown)


src/Infrastructure/LeanCode.ForceUpdate/LeanCode.ForceUpdate.Services/CQRS/VersionSupportQH.cs
```diff
@@ -24,22 +24,29 @@ VersionHandler versionHandler
         this.versionHandler = versionHandler;
     }
 
-    public async Task<VersionSupportDTO?> ExecuteAsync(HttpContext context, VersionSupport query)
+    public async Task<VersionSupportDTO> ExecuteAsync(HttpContext context, VersionSupport query)
     {
+        var (minimum, current) = GetVersions(query.Platform);
```
Patonymous: Won't this result in an exception when `query.Platform` is invalid? I'm pointing this out, because looking at code below the intended behavior was to handle this gracefully

jakubfijalkowski: Great catch, you are totally right. Will change that.

Patonymous (approved):

PR #773, opened 2025-04-02
Filter out outbox and inbox related SQL logs

Dragemil:


src/Infrastructure/LeanCode.Logging/IHostBuilderExtensions.cs
```diff
@@ -1,16 +1,23 @@
...
 using Serilog.Events;
+using Serilog.Filters;
 using Serilog.Formatting.Compact;
 
 namespace LeanCode.Logging;
 
 public static class IHostBuilderExtensions
 {
+    private static readonly SearchValues<string> OutboxInboxTablesToFilterOut = SearchValues.Create(
+        new[] { "InboxState", "OutboxState", "OutboxMessage" },
```
jakubfijalkowski: I would make it configurable—sometimes, you might have "normal" tables named "MyOutboxMessages" or something similar.

src/Infrastructure/LeanCode.Logging/IHostBuilderExtensions.cs
```diff
@@ -90,4 +98,22 @@ public static IHostBuilder ConfigureDefaultLogging(
             }
         );
     }
+
+    private static LoggerConfiguration FilterOutSqlLogsWithOutboxOrInboxTables(
```
jakubfijalkowski: Let's make it public so one can use it without the rest of the defaults.

marcin-dardzinski (approved):

PR #774, opened 2025-04-04
Merge tracing into default CQRS middleware and record exceptions where possible [CORELIB-191]

Dragemil:


marcin-dardzinski (approved):
Just add it to changelog

PR #775, opened 2025-05-14
DI-injected logger, attempt 2

Saancreed:
Supersedes https://github.com/leancodepl/corelibrary/pull/757.

I'll probably squash all that in the end, but for now I kept my own commit separate for easier reviewing.

The idea is to have one slim package that references Serilog and has only interfaces, adapters and null logger, with the rest moved to a new package. Naming suggestions are welcome; `LeanCode.Logging.Loggers` is kind of horrible.

TODO: reorder constructor parameters to match the logger-is-last convention?


src/Infrastructure/LeanCode.Logging/LeanCodeLoggingBuilderExtensions.cs
```diff
@@ -0,0 +1,29 @@
+using LeanCode.Logging;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Extensions.DependencyInjection;
+
+public static class LeanCodeLoggingBuilderExtensions
+{
+    public static ILoggingBuilder ClearLeanCodeLoggers(this ILoggingBuilder builder)
```
jakubfijalkowski: YAGNI?

test/CQRS/LeanCode.CQRS.AspNetCore.Tests/Middleware/CQRSValidationMiddlewareTests.cs
```diff
@@ -36,6 +40,11 @@ public CQRSValidationMiddlewareTests()
     protected override void ConfigureServices(IServiceCollection services)
     {
         services.AddSingleton(_ => validatorResolver);
+        services.AddLogging(logging =>
+        {
+            logging.AddProvider(NullLoggerProvider.Instance);
```
jakubfijalkowski: Shouldn't this be handled by the `TryAddNullLeanCodeLogger`? method?

jakubfijalkowski: Or even shouldn't this be omitted, as we are supposed to use our `ILogger<T>`?

src/Infrastructure/LeanCode.Logging/LeanCodeLoggingBuilderExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        builder.Services.RemoveAll(typeof(LeanCode.Logging.ILogger<>));
+
+        return builder;
+    }
+
+    public static ILoggingBuilder TryAddContextualLeanCodeLogger(this ILoggingBuilder builder)
+    {
+        builder.Services.TryAddSingleton(typeof(LeanCode.Logging.ILogger<>), typeof(ContextualLogger<>));
```
jakubfijalkowski: Transient, not singleton - otherwise the loggers will stay instantiated even if the class using them will be long gone (think rarely invoked handlers).

src/Infrastructure/LeanCode.Logging.Loggers/LeanCodeLoggingBuilderExtensions.cs
```diff
@@ -0,0 +1,75 @@
...
+namespace Microsoft.Extensions.DependencyInjection;
+
+public static class LeanCodeLoggingBuilderExtensions
+{
+    public static Serilog.ILogger AddLeanCodeLogging(
+        this ILoggingBuilder logging,
+        IConfiguration configuration,
+        string projectName,
+        string appName,
+        bool formatConsoleLogsAsJson,
```
jakubfijalkowski: What's the purpose?

jakubfijalkowski (commented):
About the `Leancode.Logging.Loggers` - `LeanCode.Logging.AspNetCore`, as the main use case is to integrate the LeanCode.Logging package with ASP.NET (analogous to `LeanCode.CQRS.AspNetCore`).

src/Infrastructure/LeanCode.Logging/LeanCodeLoggingBuilderExtensions.cs
Saancreed: It could be useful for cases where we are handed a preconfigured host builder and then we want to silence its log output before starting it. Perhaps in integration tests?

But it's nothing that can't be achieved with manual manipulation of the service collection so the value added is indeed minimal… I'll remove it and then consider reintroducing it once I actually find a place where I could use it.

test/CQRS/LeanCode.CQRS.AspNetCore.Tests/Middleware/CQRSValidationMiddlewareTests.cs
Saancreed: I think omitted would be the way to go, but I wasn't sure if our middlewares happen to pull in other classes that use classic .NET logger (or not). I'll remove it if it isn't needed.

src/Infrastructure/LeanCode.Logging/LeanCodeLoggingBuilderExtensions.cs
Saancreed: Wouldn't that cause DI validation to complain if we try to resolve this logger in a service registered as singleton?

And even if not, wouldn't Scoped be better than Transient? I don't see a reason to instanciate the logger for the same context multiple times per request.

src/Infrastructure/LeanCode.Logging.Loggers/LeanCodeLoggingBuilderExtensions.cs
Saancreed: The code I moved this from registered a different log formatter for console sink that depended on the hosting environment being Development or not. This is the same thing, only with the awareness of something like the host even existing removed.

jakubfijalkowski: Ahh, got it, that indeed is better.

src/Infrastructure/LeanCode.Logging/LeanCodeLoggingBuilderExtensions.cs
jakubfijalkowski: Good point, dunno how to tackle the singleton problem. :( Nevertheless, the main usecase is "quickly instantiate logger for a very short time", so would go with it either way (which also would work the same way as previously when it comes to GC/instantiation).

Ad. Scoped/Transient - yep, you're right. I forgot the nomenclature.

jakubfijalkowski (approved):

src/Infrastructure/LeanCode.ViewRenderer.Razor/RazorViewRendererServiceCollectionExtensions.cs
```diff
@@ -10,7 +10,9 @@ public static IServiceCollection AddRazorViewRenderer(
         RazorViewRendererOptions config
     )
     {
-        services.TryAddSingleton<IViewRenderer>(_ => new RazorViewRenderer(config));
-        return services;
+        return services.AddSingleton<IViewRenderer>(sp => new RazorViewRenderer(
+            config,
+            sp.GetRequiredService<ILogger<RazorViewRenderer>>()
+        ));
```
lukaszgarstecki, line 16: seems out of scope

test/Infrastructure/LeanCode.Logging.AspNetCore.Tests/LoggerInjectionTests.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        injectedLogger.Error("Three");
+
+        evt = Sink.LogEvent.Value;
+
+        Assert.Equal(LogEventLevel.Error, evt.Level);
+        Assert.Equal("Three", evt.MessageTemplate.Text);
+    }
+
+    internal class SingleLogEventCapturerSink : ILogEventSink
```
lukaszgarstecki: :eyes:

lukaszgarstecki (approved):

src/Infrastructure/LeanCode.ViewRenderer.Razor/RazorViewRendererServiceCollectionExtensions.cs
Saancreed: Indeed, but it was a small inconsistency I wanted to fix along the way.

test/Infrastructure/LeanCode.Logging.AspNetCore.Tests/LoggerInjectionTests.cs
Saancreed: Yeah, might as well.

PR #777, opened 2025-06-16
MCP server tools

Saancreed:


src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSObjectCollection.cs
```diff
@@ -0,0 +1,97 @@
...
+            if (ValidateContractType(contract) is not { } objectKind)
+            {
+                continue;
+            }
+
+            yield return (contract, objectKind);
+        }
+    }
+
+    private static CQRSObjectKind? ValidateContractType(TypeInfo type)
```
jakubfijalkowski: It does not validate, it maps type to object kind.

Also, I'm not so sure if "parse, don't validate" works here that well (when reading the code).

src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSObjectCollection.cs
```diff
@@ -0,0 +1,97 @@
...
+            .Where(t => IsCommand(t) || IsQuery(t) || IsOperation(t));
+
+        foreach (var contract in contracts)
+        {
+            if (ValidateContractType(contract) is not { } objectKind)
+            {
+                continue;
+            }
+
+            yield return (contract, objectKind);
```
jakubfijalkowski:
```suggestion
            if (ValidateContractType(contract) is { } objectKind)
            {
                 yield return (contract, objectKind);
            }
```
This is both shorter and simpler.

src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSObjectCollection.cs
```diff
@@ -0,0 +1,97 @@
...
+        var isQuery = type.ImplementedInterfaces.Count(i => IsGenericType(i, typeof(IQuery<>))) == 1;
+        var isOperation = type.ImplementedInterfaces.Count(i => IsGenericType(i, typeof(IOperation<>))) == 1;
+
+        return (isCommand, isQuery, isOperation) switch
+        {
+            (true, false, false) => CQRSObjectKind.Command,
+            (false, true, false) => CQRSObjectKind.Query,
+            (false, false, true) => CQRSObjectKind.Operation,
+            _ => null,
+        };
```
jakubfijalkowski: This is one of the places where I think pattern matching does not work. Simple `if...else if...else if...else` will be simpler to read.

src/Infrastructure/LeanCode.AI.McpServer/McpServerBuilderExtensions.cs
```diff
@@ -0,0 +1,133 @@
...
+                        SerializerOptions = serializerOptions,
+                    }
+                )
+            );
+        }
+
+        return builder.WithTools(tools);
+    }
+
+    public static IMcpServerBuilder WithToolsFromCommands(
```
jakubfijalkowski: What's the `executorType`?

src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSObjectCollection.cs
```diff
@@ -0,0 +1,97 @@
+using System.Collections.Frozen;
+using System.Reflection;
+using LeanCode.Components;
+using LeanCode.Contracts;
+using LeanCode.CQRS.Execution;
+
+namespace LeanCode.CQRS.AspNetCore.Registration;
+
+public class CQRSObjectCollection : ICQRSObjectSource
```
jakubfijalkowski: After reading this reafactor, I don't like how this looks like now.

The responsibilities of `CQRSObjectCollection` and `CQRSObjectsRegistrationSource` are quite unclear. This is even more pronounced because `CQRSObjectCollection` seemingly randomly allows some methods to be overriden, some methods to be protected and some even public. We need to work on that (what goes where), otherwise maintaining the code will be hard.

If I was going to refactor this, I would do something like this:
1) `CQRSObjectCollection` should be responsible for assembly scanning and type mapping. It should support adding a single type, and adding the whole assembly. It should have enough metadata about the objects so that you could use it, without the need to pass `TypesCatalog` along the way. It should basically expose two main methods (+ helpers) - `AddCQRSObject(CQRSObjectMetadata)` and `AddCQRSObjects(TypesCatalog)` - and two properties - the `Objects` list and cache. It probably should work on contracts only and should ignore the handlers. It then could be sealed. Probably, it also should be named `CQRSContractsCollection` as it won't really be able to keep `CQRSObjectMetadata` since it won't know about the handlers.
2) `CQRSObjectsRegistrationSource` is the place that maps objects from `CQRSObjectCollection` to handlers and maintains `CQRSObjectMetadata`s. I probably would re-work the API quite substantially to better work with the two-pass approach (i.e. scan contracts first, then find handlers).

If that won't work for you, I would just duplicate the scanning code.

src/Infrastructure/LeanCode.AI/CQRSAIFunction.cs
```diff
@@ -0,0 +1,209 @@
...
+using Microsoft.Extensions.AI;
+using Microsoft.Extensions.DependencyInjection;
+using static LeanCode.AI.Contracts.McpToolHints;
+using IOException = System.InvalidOperationException;
+
+namespace LeanCode.AI;
+
+public abstract class CQRSAIFunction(Type cqrsObjectType) : AIFunction
+{
+    public enum JsonSchemaSource
```
jakubfijalkowski: This will be helluva hard to understand. Let's pick one and just go with it. If it ends up not working correctly, we will switch centrally here. There is no point in maintaining both if we don't have a sensible use-case.

src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSObjectCollection.cs
```diff
@@ -0,0 +1,97 @@
...
+
+            yield return (contract, objectKind);
+        }
+    }
+
+    private static CQRSObjectKind? ValidateContractType(TypeInfo type)
+    {
+        var isCommand = type.ImplementedInterfaces.Count(i => i == typeof(ICommand)) == 1;
+        var isQuery = type.ImplementedInterfaces.Count(i => IsGenericType(i, typeof(IQuery<>))) == 1;
+        var isOperation = type.ImplementedInterfaces.Count(i => IsGenericType(i, typeof(IOperation<>))) == 1;
```
sivvak: Why is this a separate logic from the methods below?

src/Infrastructure/LeanCode.AI/CQRSAIFunction.cs
```diff
@@ -0,0 +1,209 @@
...
+
+    public Type CQRSObjectType { get; } = cqrsObjectType;
+    public McpToolAttribute? Metadata { get; } = cqrsObjectType.GetCustomAttribute<McpToolAttribute>();
+    public string? Title { get; protected init; }
+    public McpToolHints ToolHints { get; protected init; }
+
+    protected abstract McpToolHints DefaultHints { get; }
+}
+
+public abstract class CQRSAIFunction<TExecutor, TObject, TResult> : CQRSAIFunction
```
jakubfijalkowski: Do we need to know the `TObject` type here? Can't we work on `object`s? We can't really do anything with the type either way.

src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSObjectCollection.cs
```diff
@@ -0,0 +1,97 @@
...
+        var contracts = contractsCatalog
+            .Assemblies.SelectMany(a => a.DefinedTypes)
+            .Where(t => IsCommand(t) || IsQuery(t) || IsOperation(t));
+
+        foreach (var contract in contracts)
+        {
+            if (ValidateContractType(contract) is not { } objectKind)
+            {
+                continue;
+            }
```
sivvak: Don't we double check very similar conditions here and in the `Where` above?

src/Infrastructure/LeanCode.AI.Contracts/McpToolAttribute.cs
```diff
@@ -0,0 +1,13 @@
...
+
+namespace LeanCode.AI.Contracts;
+
+[ExcludeFromContractsGeneration]
+[AttributeUsage(AttributeTargets.Class, Inherited = false, AllowMultiple = false)]
+public sealed class McpToolAttribute(string description) : Attribute
+{
+    public string Description { get; } = description;
+    public string? Name { get; init; }
+    public string? Title { get; init; }
```
sivvak, line 11: What's the difference between those two?

src/Infrastructure/LeanCode.AI.Contracts/McpToolHints.cs
```diff
@@ -0,0 +1,16 @@
...
+[ExcludeFromContractsGeneration]
+[Flags]
+public enum McpToolHints
+{
+    None = 0,
+    Default = 1,
+
+    Destructive = 0b10,
+    Idempotent = 0b100,
+    OpenWorld = 0b1000,
```
sivvak, line 14: What does it mean exactly?

src/Infrastructure/LeanCode.AI/CQRSAIFunction.cs
```diff
@@ -0,0 +1,209 @@
...
+        var jso = JsonSerializerOptions;
+        var services = arguments.Services ?? throw new IOException("Arguments have no services.");
+        var nodes = arguments.Select(kvp => new KeyValuePair<string, JsonNode?>(
+            kvp.Key,
+            kvp.Value is JsonElement je ? je.Deserialize<JsonNode>(jso) : null
+        ));
+        var jsonObject = new JsonObject(nodes);
+        var cqrsObject = jsonObject.Deserialize<TObject>(jso) ?? throw new UnreachableException();
+
+        return InvokeCoreAsync(arguments, services.GetRequiredService<TExecutor>(), cqrsObject, cancellationToken);
```
jakubfijalkowski: If you move the `GetRequiredService` to `InvokeCoreAsync`, then you could ditch the `TExecutor` type from this class.

Also, we should open a scope before resolving `TExecutor`.

src/Infrastructure/LeanCode.AI.McpServer/McpServerBuilderExtensions.cs
jakubfijalkowski: We need to have a simple way of configuring this. Currently, deciding what to call/what to pass here would be quite tricky. The changes that I would want for v1 is to have a _simple_ way of registering all three:
1) `WithCommandsFromContracts<TExecutor>(TypesCatalog contracts, JsonSerializerOptions? serializerOptions = null)`,
2) `WithQueriesFromContracts<TExecutor>(TypesCatalog contracts, JsonSerializerOptions? serializerOptions = null)`,
3) `WithOperationsFromContracts<TExecutor>(TypesCatalog contracts, JsonSerializerOptions? serializerOptions = null)`,
where `TExecutor` is constrained to the corresponding classes.

After we re-work the `HttpExecutor`s, we probably could simplify this (a single executor for all?) but for now, it IMO would be simpler, compared to the random `Type executorType` that doesn't really tell you anything.

src/Infrastructure/LeanCode.AI/CQRSAIFunction.cs
```diff
@@ -0,0 +1,209 @@
...
+using System.Text.Json.Schema;
+using LeanCode.AI.Contracts;
+using LeanCode.Contracts;
+using LeanCode.CQRS.RemoteHttp.Client;
+using Microsoft.Extensions.AI;
+using Microsoft.Extensions.DependencyInjection;
+using static LeanCode.AI.Contracts.McpToolHints;
+using IOException = System.InvalidOperationException;
+
+namespace LeanCode.AI;
```
jakubfijalkowski: Why the split between `LeanCode.AI` and `LeanCode.AI.McpServer`? What do we get by having two assemblies with a single file in each?

src/Infrastructure/LeanCode.AI.Contracts/McpToolHints.cs
Saancreed: > Whether this tool may interact with an "open world" of external entities. If true, the tool may interact with an unpredictable or dynamic set of entities (like web search). If false, the tool's domain of interaction is closed and well-defined (like memory access).

src/Infrastructure/LeanCode.AI.Contracts/McpToolAttribute.cs
Saancreed: Name is the programmatic identifier of the tool that should be unique and restricted to a closed set of characters (I believe it's something like `[A-Za-z0-9_]`). Title is a human-readable name that doesn't have such restrictions.

https://modelcontextprotocol.io/docs/concepts/tools#tool-definition-structure

src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSObjectCollection.cs
Saancreed: Similar but not exactly the same. The `Where` above checks if the type has any number of implemented CQRS interfaces, this one verifies that it implements exactly one.

Yes, it's a bit weird, but [original code does the same thing](https://github.com/leancodepl/corelibrary/blob/v9.0-preview/src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSObjectsRegistrationSource.cs#L33-L47).

src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSObjectCollection.cs
Saancreed: Really? How is
```csharp
        return (isCommand, isQuery, isOperation) switch
        {
            (true, false, false) => CQRSObjectKind.Command,
            (false, true, false) => CQRSObjectKind.Query,
            (false, false, true) => CQRSObjectKind.Operation,
            _ => null,
        };
```
less readable than
```csharp
        if (isCommand && !isQuery && !isOperation)
        {
            return CQRSObjectKind.Command;
        }
        else if (!isCommand && isQuery && !isOperation)
        {
            return CQRSObjectKind.Query;
        }
        else if (!isCommand && !isQuery && isOperation)
        {
            return CQRSObjectKind.Operation;
        }
        else
        {
            return null;
        }
```
I very much disagree with six `!` being more readable than pattern matching.

src/Infrastructure/LeanCode.AI.McpServer/McpServerBuilderExtensions.cs
Saancreed: It configures what subtype of `HttpCommandsExecutor` should be retrieved from DI to execute the remote CQRS call.

But your comment made me realize I can move it to generic type parameters, so I'll do just that.

src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSObjectCollection.cs
Saancreed: The proposed refactor sounds good, but it's quite a bit of extra work.

I think I'll just revert it and write my own contract enumeration directly in `McpServer`, which should allow me to drop the dependency on CQRS package.

src/Infrastructure/LeanCode.AI/CQRSAIFunction.cs
Saancreed: Fair, I'll just remove it.

src/Infrastructure/LeanCode.AI/CQRSAIFunction.cs
Saancreed: We need it to somehow call `GetAsync`, so we'd need this type to construct appropriate generic method one way or another.

src/Infrastructure/LeanCode.AI/CQRSAIFunction.cs
Saancreed: For the first one, sure, I can do it, but it probably won't give us too many benefits.

For the second one, I initially planned to rely on [McpServer creating a scope for us](https://github.com/modelcontextprotocol/csharp-sdk/blob/2bcff315f8839a1187e8a61f1b110578230ce0b8/src/ModelContextProtocol.Core/Server/McpServer.cs#L542-L548) but sure, I can create my own scope if you think it's a good idea.

src/Infrastructure/LeanCode.AI.McpServer/McpServerBuilderExtensions.cs
Saancreed: Right, will do.

Plus, I intend to replace `JsonSerializerOptions? serializerOptions = null` with some `Action<stuff, McpServerToolCreateOptions>? configureAction = null`… serializer options probably won't be customized that often anyway.

src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSObjectCollection.cs
lukaszgarstecki: IMO check invalid state first and if else will look better

src/Infrastructure/LeanCode.AI/CQRSAIFunction.cs
jakubfijalkowski: > For the first one, sure, I can do it, but it probably won't give us too many benefits.

One less generic to pass ;P I'm not insisting on that, I agree it won't change that much (except probably you could resolve the executor using the constructor then, but that's not much of a change).

> For the second one, I initially planned to rely on [McpServer creating a scope for us](https://github.com/modelcontextprotocol/csharp-sdk/blob/2bcff315f8839a1187e8a61f1b110578230ce0b8/src/ModelContextProtocol.Core/Server/McpServer.cs#L542-L548) but sure, I can create my own scope if you think it's a good idea.

Ah, it makes sense. If the services provider is already scoped, then we can just use it, sure.

src/Infrastructure/LeanCode.AI/CQRSAIFunction.cs
Saancreed: I wanted to keep the dependency on `ModelContextProtocol`, which is a preview package with a specific implementation, to be as isolated as possible. I'd reconsider it when it's out of preview and we decide to stick with it as the only MCP server implementation we wish to use, but it could very much be a YAGNI moment for me :see_no_evil:

src/Infrastructure/LeanCode.AI/CQRSAIFunction.cs
jakubfijalkowski: Ugh, right, we need the `TResult`. `TObject` might be skipped (the executor accepts `IQuery<TResult>`), but the result is needed to deserialize things.

I would probably tinker with this so that we could drop it (because, in the end, we _don't_ need to deserialize to a concrete type - `JsonObject` would work), but I also don't think it is worth doing right now.

I think you can ignore this comment.

src/CQRS/LeanCode.CQRS.AspNetCore/Registration/CQRSObjectCollection.cs
jakubfijalkowski: Yeah, then let's go with the faster way. It won't be that hard to maintain (but I'll leave a task in Jira with a reference to this comment to refactor sometime later).

jakubfijalkowski (approved):

PR #780, opened 2025-09-04
Add some documentation for TimestampTz type

Saancreed:


docs/domain/timestamp_tz/index.md
```diff
@@ -0,0 +1,74 @@
...
+    public class ErrorCodes
+    {
+        public const int InvalidTimeZoneId = 10001;
+    }
+}
+```
+
+should satisfy those needs, with mapping between the two being trivial.
+
+Because EF currently does not support including properties of nested objects in index definitions, indexing timestamps stored in `TimestampTz` requires building indexes manually, using [`MigrationBuilder.Sql` API](https://learn.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.migrations.migrationbuilder.sql), for example:
```
Patonymous: Not necessarily `MigrationBuilder.Sql`. Hand-written `MigrationBuilder.CreateIndex` also works from my experience (albeit in SQL Server)

Saancreed: Hmm, okay, I suppose that could work too, unless one wanted to create an index on some expression instead of a column, like I did in the example below.

Patonymous: I guess in that case you're right.
Also is index like the one in the example even legal? Isn't `at time zone` non-deterministic?

Saancreed: I also thought it would be illegal, then I couldn't find anything in the docs that would suggest it is, then I tried running the following on my local instance and it just worked™:

```sql
create temporary table events (
    i int primary key,
    ts timestamptz,
    tz text
) on commit drop;

create index ix_events on events ((ts at time zone tz));

insert into events (i, ts, tz) values
(1, '2025-09-04T15:00:00Z', 'Europe/Warsaw'),
(2, '2025-09-04T15:00:00Z', 'Asia/Tokyo'),
(3, '2025-09-04T15:00:00Z', 'America/Los_Angeles');

explain analyze select *, ts at time zone tz as lts
from events
order by ts at time zone tz;
```

So… :shrug:?

docs/domain/timestamp_tz/index.md
```diff
@@ -0,0 +1,74 @@
...
+
+Additionally, if database-side support for `LocalTimestampWithoutOffset` is desired, use `TimestampTzExpressionInterceptorDbContextOptionsBuilderExtensions.AddTimestampTzExpressionInterceptor(this DbContextOptionsBuilder)` to install a query expression interceptor that will replace reads of that property with equivalent expression that EF will be able to translate to SQL.
+
+When reading, usually converting to `LocalTimestampWithOffset` and returning only that to clients might be sufficient. When doing so, remember that `LocalTimestampWithOffset` cannot be evaluated on Postgres side, so avoid using this property in queries in contexts where its value would participate in some computation, e.g. sorting.
+
+On the other hand, when writing and when replicating saved inputs for readback, it might be desirable to have both properties in the contract. In such case, a DTO like
```
jakubfijalkowski, line 40: Should we standardize this, maybe?

docs/domain/timestamp_tz/index.md
```diff
@@ -0,0 +1,74 @@
...
+* Support for different calendar systems is out of scope.
+
+## FAQ
+
+### Q: When to use `DateTime`, `DateTimeOffset` and `TimestampTz` in domain entities?
+
+A: In Postgres, use `DateTimeOffset` (mapped to `timestamp with time zone` column type) to represent an instant without associated time zone (e.g. various `CreatedAt` timestamps generated on the backend). Use `TimestampTz` to persist an instant in context of some time zone (e.g. in a system aggregating news, it could be a time of an event that took place in some location). Use `DateTime` mapped to `timestamp without time zone` column type _together with_ `TimestampTz` to represent the local time of some event, often one meant to start in the future, or for indexing purposes. Avoid using `DateTime` mapped to `timestamp with time zone` columns: it has no benefits over using `DateTimeOffset`.
```
jakubfijalkowski, line 74: Won't +5 offset be problematic when saved in `timestamp with time zone`? How that will behave?

docs/domain/timestamp_tz/index.md
Saancreed: I can see the appeal, but I wouldn't consider each project maintaining it's own DTO to be that big of an overhead. Such a DTO doesn't have any place to live yet (LeanCode.DomainModels.Contracts?), and I don't think it's a good idea to put in separate packages DTOs with error codes that would effectively have to be globally reserved for them.

docs/domain/timestamp_tz/index.md
Saancreed: Well, you can't save it there, Npgsql should (IIRC) refuse to write it, which the first paragraph calls out:

> […] created to address the inability to persist a timestamp in particular time zone in a single column in Postgres.

But I will reword it to specify that "in particular time zone" also means "with some non-zero offset".

docs/domain/timestamp_tz/index.md
jakubfijalkowski: Makes sense, thanks!

docs/domain/timestamp_tz/index.md
jakubfijalkowski: Good idea! Yeah, when reading the last paragraph I already forgot was on the first line of the document. ;P

jakubfijalkowski (approved):

PR #783, opened 2025-09-24
Add enum dto consistency analyzer

piotrek999:
Should I add a helper with common methods (e.g. FindBaseEnum, FindEnumInCompilation, GetExcludedMembers andAreEnumValuesEqual)?

### DIFF
--- a/src/Tools/LeanCode.CodeAnalysis/Analyzers/EnumDtoConsistencyAnalyzer.cs
+++ b/src/Tools/LeanCode.CodeAnalysis/Analyzers/EnumDtoConsistencyAnalyzer.cs
@@ -296,39 +296,37 @@ public class EnumDtoConsistencyAnalyzer : DiagnosticAnalyzer
 
     private static bool AreEnumValuesEqual(object? value1, object? value2)
     {
-        if (value1 == null && value2 == null)
+        if (value1 is null && value2 is null)
         {
             return true;
         }
-        if (value1 == null || value2 == null)
+        if (value1 is null || value2 is null)
         {
             return false;
         }
 
-        // This handles cases where enums have different underlying types (byte, short, int, etc.)
-        if (value1.GetType() != value2.GetType())
+        var type1 = value1.GetType();
+        var type2 = value2.GetType();
+
+        if (!type1.IsEnum || !type2.IsEnum)
         {
-            try
-            {
-                var converted1 = Convert.ToInt64(value1, CultureInfo.InvariantCulture);
-                var converted2 = Convert.ToInt64(value2, CultureInfo.InvariantCulture);
-                return converted1 == converted2;
-            }
-            catch (FormatException)
-            {
-                return false;
-            }
-            catch (OverflowException)
-            {
-                return false;
-            }
-            catch (InvalidCastException)
-            {
-                return false;
-            }
+            return value1.Equals(value2);
         }
 
-        return value1.Equals(value2);
+        try
+        {
+            var underlyingValue1 = Convert.ToInt64(value1, CultureInfo.InvariantCulture);
+            var underlyingValue2 = Convert.ToInt64(value2, CultureInfo.InvariantCulture);
+            return underlyingValue1 == underlyingValue2;
+        }
+        catch (OverflowException)
+        {
+            return false;
+        }
+        catch (InvalidCastException)
+        {
+            return false;
+        }
     }
 
     private sealed class EnumConsistencyAnalysis
--- a/src/Tools/LeanCode.CodeAnalysis/Analyzers/EnumDtoConsistencyAnalyzer.cs
+++ b/src/Tools/LeanCode.CodeAnalysis/Analyzers/EnumDtoConsistencyAnalyzer.cs
@@ -1,0 +1,342 @@
+using System.Collections.Immutable;
+using System.Globalization;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace LeanCode.CodeAnalysis.Analyzers;
+
+[DiagnosticAnalyzer(LanguageNames.CSharp)]
+public class EnumDtoConsistencyAnalyzer : DiagnosticAnalyzer
+{
+    private const string Category = "Design";
+    private const string DtoSuffix = "DTO";
+    private const string IgnoreEnumDtoAttributeName = "LeanCode.CodeAnalysis.IgnoreEnumDtoAttribute";
+    private const string ExcludeMembersAttributeName = "LeanCode.CodeAnalysis.ExcludeMembersAttribute";
+    private const string IgnoreEnumValueAttributeName = "LeanCode.CodeAnalysis.IgnoreEnumValueAttribute";
+    private const string EnumValueCorrespondsAttributeName = "LeanCode.CodeAnalysis.EnumValueCorrespondsAttribute";
+
+    private static readonly DiagnosticDescriptor Rule = new DiagnosticDescriptor(
+        DiagnosticsIds.EnumDtoShouldMatchBaseEnum,
+        "DTO enum should match its base enum",
+        "DTO enum '{0}' should have the same members as its base enum '{1}'. Missing: [{2}]. Extra: [{3}]. Value mismatches: [{4}].",
+        Category,
+        DiagnosticSeverity.Warning,
+        isEnabledByDefault: true,
+        description: "DTO enums should have the same members (names and values) as their corresponding base enums, unless explicitly configured with attributes."
+    );
+
+    public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = [Rule];
+
+    public override void Initialize(AnalysisContext context)
+    {
+        context.EnableConcurrentExecution();
+        context.ConfigureGeneratedCodeAnalysis(
+            GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics
+        );
+        context.RegisterSymbolAction(AnalyzeEnum, SymbolKind.NamedType);
+    }
+
+    private static void AnalyzeEnum(SymbolAnalysisContext context)
+    {
+        var enumSymbol = (INamedTypeSymbol)context.Symbol;
+
+        if (enumSymbol.TypeKind != TypeKind.Enum)
+        {
+            return;
+        }
+
+        if (!enumSymbol.Name.EndsWith(DtoSuffix, StringComparison.Ordinal))
+        {
+            return;
+        }
+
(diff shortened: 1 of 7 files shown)


test/Tools/LeanCode.CodeAnalysis.Tests/Analyzers/EnumDtoConsistencyAnalyzerTests.cs
```diff
@@ -0,0 +1,370 @@
...
+    Cancelled = 3
+}
+
+public enum StatusDTO
+{
+    None = 0,
+    InProgress = 1,
+    Completed = 2,
+    Cancelled = 3,
+    Deleted = 4
```
Saancreed: Question: would this still report a diagnostic even if `Deleted` was marked with `ExcludeFromContractsGeneration`? If yes, maybe it shouldn't?

src/Tools/LeanCode.CodeAnalysis/EnumDtoAttributes.cs
```diff
@@ -0,0 +1,76 @@
...
+/// <summary>
+/// Attribute that marks an enum ending with "DTO" to be completely ignored
+/// by the enum DTO consistency analyzer.
+/// </summary>
+/// <remarks>
+/// This attribute can only be applied to enums whose names end with "DTO".
+/// When applied, the enum will not be checked for consistency with its base enum.
+/// </remarks>
+[AttributeUsage(AttributeTargets.Enum, Inherited = false, AllowMultiple = false)]
+public sealed class IgnoreEnumDtoAttribute : Attribute { }
```
Saancreed:
```suggestion
public sealed class IgnoreEnumDtoAttribute : Attribute;
```

src/Tools/LeanCode.CodeAnalysis/EnumDtoAttributes.cs
```diff
@@ -0,0 +1,76 @@
...
+/// <summary>
+/// Attribute that marks a specific enum value in a DTO enum to be ignored
+/// during consistency checking with the base enum.
+/// </summary>
+/// <remarks>
+/// This attribute can only be applied to enum values within enums whose names end with "DTO".
+/// When applied, the analyzer will not require a corresponding value in the base enum.
+/// </remarks>
+[AttributeUsage(AttributeTargets.Field, Inherited = false, AllowMultiple = false)]
+public sealed class IgnoreEnumValueAttribute : Attribute { }
```
Saancreed:
```suggestion
public sealed class IgnoreEnumValueAttribute : Attribute;
```

test/Tools/LeanCode.CodeAnalysis.Tests/Analyzers/EnumDtoConsistencyAnalyzerTests.cs
```diff
@@ -0,0 +1,370 @@
...
+using LeanCode.CodeAnalysis.Tests.Verifiers;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Xunit;
+
+namespace LeanCode.CodeAnalysis.Tests.Analyzers;
+
+public class EnumDtoConsistencyAnalyzerTests : DiagnosticVerifier
+{
+    private const string AttributeDefinitions =
+        @"
```
Saancreed: Why aren't you using `"""` for multiline strings? This isn't C# 10 anymore :stuck_out_tongue:

test/Tools/LeanCode.CodeAnalysis.Tests/Analyzers/EnumDtoConsistencyAnalyzerTests.cs
```diff
@@ -0,0 +1,370 @@
...
+}
+
+public enum StatusDTO
+{
+    None = 0,
+    [LeanCode.CodeAnalysis.EnumValueCorresponds(Status.InProgress)]
+    InProgress = 2,
+    [LeanCode.CodeAnalysis.EnumValueCorresponds(Status.Completed)]
+    Completed = 3,
+    [LeanCode.CodeAnalysis.EnumValueCorresponds(Status.Cancelled)]
```
Saancreed: Do remember that this is something you'll never see in practice. Contracts projects won't have references do domain projects, so at best you can do something like

```suggestion
    [LeanCode.CodeAnalysis.EnumValueCorresponds(3)]
```

Saancreed (commented):
Would it be possible to report a diagnostic if an enum is declared without explicit values assigned?

And out of curiosity, how well does this handle enums with multiple members having the same value? Maybe it deserves some tests to showcase it.

src/Tools/LeanCode.CodeAnalysis/Analyzers/EnumDtoConsistencyAnalyzer.cs
```diff
@@ -0,0 +1,342 @@
...
+        {
+            return true;
+        }
+        if (value1 == null || value2 == null)
+        {
+            return false;
+        }
+
+        // This handles cases where enums have different underlying types (byte, short, int, etc.)
+        if (value1.GetType() != value2.GetType())
```
Patonymous: Does `GetType` actually return `int`, `short`, etc. here? Not `TestEnum` and `TestEnumDTO`? I thought it's necessary to use `Enum.GetUnderlyingType`

test/Tools/LeanCode.CodeAnalysis.Tests/Analyzers/EnumDtoConsistencyAnalyzerTests.cs
Patonymous, line 1: I feel like tests concerning `EnumValueCorrespondsAttribute` are pretty similar, do we need all of them?

src/Tools/LeanCode.CodeAnalysis/CodeActions/SynchronizeEnumDtoCodeAction.cs
Patonymous, line 1: Some helper functions here are the same as in `EnumDtoConsistencyAnalyzer.cs`, can't we reuse them?

Patonymous (approved):
Code looks good, I have some general questions and I'm also interested in answers to previous @Saancreed's comments

src/Tools/LeanCode.CodeAnalysis/EnumDtoAttributes.cs
```diff
@@ -0,0 +1,78 @@
+namespace LeanCode.CodeAnalysis;
```
jakubfijalkowski, line 1: I think we can't keep these attributes in this project. `LeanCode.CodeAnalysis` is referenced as [global package](https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management#global-package-references), which are private-by-default (i.e. only in dev/build), thus they must not be referenced in the code. And attributes inevitably will be.

src/Tools/LeanCode.CodeAnalysis/Analyzers/EnumDtoConsistencyAnalyzer.cs
```diff
@@ -0,0 +1,340 @@
...
+    private static bool HasIgnoreEnumDtoAttribute(INamedTypeSymbol enumSymbol)
+    {
+        return enumSymbol
+            .GetAttributes()
+            .Any(attr => attr.AttributeClass?.GetFullNamespaceName() == IgnoreEnumDtoAttributeName);
+    }
+
+    private static INamedTypeSymbol? FindBaseEnum(INamedTypeSymbol dtoEnum, string baseEnumName)
+    {
+        var sameNamespaceEnum = dtoEnum
```
jakubfijalkowski, line 93: Can this even happen in real project? We will always compare Contracts enum with Domain enum, thus the namespace will be different.

src/Tools/LeanCode.CodeAnalysis/Analyzers/EnumDtoConsistencyAnalyzer.cs
```diff
@@ -0,0 +1,340 @@
...
+
+        if (sameNamespaceEnum is not null)
+        {
+            return sameNamespaceEnum;
+        }
+
+        return FindEnumInCompilation(dtoEnum.ContainingAssembly.GlobalNamespace, baseEnumName);
+    }
+
+    private static INamedTypeSymbol? FindEnumInCompilation(INamespaceSymbol namespaceSymbol, string enumName)
```
jakubfijalkowski, line 105: This searches through namespaces, not compilations, the name is slightly misleading. Moreover, I don't think this will work - as far as I get how `dtoEnum.ContainingAssembly` works, it will only return the _contracts assembly_. And we have to find an enum in a completely different assembly. Moreover, one that is not referenced by the contracts at all.

I wonder if it is possible to implement this check. Contracts _don't know anything_ about the domain, the domain is only (happens to be) in the same solution. Roslyn does not work on solutions, it works on assemblies/projects. Thus, it _cannot know_ about the domain enum. Or am I missing something?

@Saancreed , I think you are the most well-versed in Roslyn. Is my thinking right?

src/Tools/LeanCode.CodeAnalysis/EnumDtoAttributes.cs
Saancreed: Would `[Conditional("…")]` on these attributes solve this perhaps? :thinking:

src/Tools/LeanCode.CodeAnalysis/Analyzers/EnumDtoConsistencyAnalyzer.cs
Saancreed: > @Saancreed, I think you are the most well-versed in Roslyn.

Uhm, citation needed? :see_no_evil:

> Is my thinking right?

I also think it's not possible as long as we talk in terms of analysing just contracts or just domain projects. On the other hand, at some point we'll be building some project that references both, at which point I _think_ it should be possible to do a pass over all known/referenced types when both contract and domain enums are visible. Maybe this should be the point where we run the analyzer and report violations.

But of course that makes `[Conditional("…")]` idea unviable :slightly_frowning_face:

src/Tools/LeanCode.CodeAnalysis/EnumDtoAttributes.cs
jakubfijalkowski: I think the condition would be hard to come up with :P

src/Tools/LeanCode.CodeAnalysis/Analyzers/EnumDtoConsistencyAnalyzer.cs
jakubfijalkowski: That... might work! Thx!

PR #786, opened 2025-11-07
Caching prep - write responses in middlewares & add ADR about cache support for local execution 

Dragemil:
This PR aims to prepare our CQRS middleware stack, to introduce ASP.NET Core's [Output Caching middleware](https://learn.microsoft.com/en-us/aspnet/core/performance/caching/output?view=aspnetcore-9.0) as a possible part of our CQRS middleware pipeline.

That cache operates on `HttpContext.Response` so we needed to change some CQRS middleware resposibilities. When CQRS middleware short circuits the pipeline:

- before: it MUST attach an `ExecutionResult` to the `HttpContext` features
- now: it MUST populate the `HttpContext.Response` with proper `Body` and `StatusCode`, and still SHOULD attach an `ExecutionResult` to the `HttpContext` features

`ExecutionResult` is still needed if the middleware were to be supported in our CQRS local execution pipeline. All existing CQRS middlewares are attaching it as before, so they are still supported locally. But Output Caching middleware won't do that, so it won't be supported in local pipeline, as per ADR present in this PR.

This PR is just a first part of caching support, no caching support OOTB is introduced here yet. I'll stack another PR on top of this one.


adrs/2025-11-05_output-cache-vs-local-cqrs.md
jakubfijalkowski, line 1: Nit: `adrs/DATE_name.md`, making two folders for a single file is too much IMO :P

src/CQRS/LeanCode.CQRS.AspNetCore/Middleware/CQRSExceptionTranslationMiddleware.cs
```diff
@@ -45,10 +45,11 @@ public async Task InvokeAsync(HttpContext httpContext)
             activity?.SetTag("error.code", ex.ErrorCode);
 
             var result = WrapInCommandResult(ex);
+            metrics.CQRSFailure(CQRSMetrics.ValidationFailure);
```
jakubfijalkowski, line 48: Unnecessary move or was this deliberate?

jakubfijalkowski (approved):
Some nitpicks, but overall LGTM.

Let's wait with merging this - this introduces breaking changes, and I don't think we should add them to v9. @michal12334 works on v10 already, so at the beginning of the next week, we should have the branches/versions cleaned up.

src/CQRS/LeanCode.CQRS.AspNetCore/Middleware/CQRSExceptionTranslationMiddleware.cs
Dragemil: It's consistent with other middlewares changes, I thought its reasonable to have all observability calls next to each other?

PR #791, opened 2025-11-14
Integrate Output Caching middleware into CQRS

Dragemil:
The best way to find out how it works is to read the provided docs.


src/CQRS/LeanCode.CQRS.AspNetCore/ServiceCollectionCQRSExtensions.cs
```diff
@@ -4,13 +4,18 @@
 using LeanCode.CQRS.AspNetCore.Registration;
 using LeanCode.CQRS.AspNetCore.Serialization;
 using LeanCode.CQRS.Execution;
+using LeanCode.CQRS.OutputCaching;
```
jakubfijalkowski: Nit: Unnecessary?

src/CQRS/LeanCode.CQRS.OutputCaching/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+        bool registerOutputCache = true
+    )
+    {
+        return services.AddCQRSOutputCache(catalog.Assemblies, registerOutputCache);
+    }
+
+    public static IServiceCollection AddCQRSOutputCache(
+        this IServiceCollection services,
+        IEnumerable<Assembly> assemblies,
+        bool registerOutputCache = true
```
jakubfijalkowski: Do we need this flag? Can't we _always_ call `AddOutputCache`? AFAIK all the `IServiceCollection` extensions provided by ASP.NET are re-entrant, i.e. they use `TryAdd...` underneath, so they won't register things twice.

src/CQRS/LeanCode.CQRS.OutputCaching/Registration/CQRSOutputCacheOptionsConfigurator.cs
```diff
@@ -0,0 +1,28 @@
...
+    public void Configure(OutputCacheOptions options)
+    {
+        foreach (var policyType in registry.PolicyForObject.Values)
+        {
+            var name = CQRSOutputCachePolicyName.For(policyType);
+            RegisterPolicy(options, name, policyType);
+        }
+    }
+
+    protected virtual void RegisterPolicy(OutputCacheOptions options, string policyName, Type policyType)
```
jakubfijalkowski: `protected virtual` for tests?

test/CQRS/LeanCode.CQRS.OutputCaching.Tests/CQRSOutputCacheOptionsConfiguratorTests.cs
```diff
@@ -0,0 +1,43 @@
...
+            .RegisteredPolicies.Should()
+            .BeEquivalentTo(
+                [
+                    (CQRSOutputCachePolicyName.For(typeof(TestQueryOCP)), typeof(TestQueryOCP)),
+                    (CQRSOutputCachePolicyName.For(typeof(TestOperationOCP)), typeof(TestOperationOCP)),
+                ]
+            );
+    }
+
+    private sealed class RecordingConfigurator(CQRSOutputCacheRegistry registry)
```
jakubfijalkowski, line 32: TBH, if we want to test that loop inside of configurator, I think accessing the internal properties with policies in tests is a better option. This approach overcomes that `internal NamedPolicies` with much more code than _just_ `public List<Policy> GetNamedPolicies(this OutputCacheOptions opts) => typeof(OutputCacheOptions).GetProperty("NamedPolicies").GetValue(opts)`, but without any substantial benefit.

src/CQRS/LeanCode.CQRS.OutputCaching/Registration/CQRSOutputCachePolicyName.cs
```diff
@@ -0,0 +1,10 @@
+namespace LeanCode.CQRS.OutputCaching.Registration;
+
+internal static class CQRSOutputCachePolicyName
+{
+    public static string For(Type objectType)
+    {
+        ArgumentNullException.ThrowIfNull(objectType);
```
jakubfijalkowski: Unneeded

src/CQRS/LeanCode.CQRS.OutputCaching/Registration/CQRSOutputCacheRegistry.cs
```diff
@@ -0,0 +1,91 @@
+using System.Reflection;
+using LeanCode.Contracts;
+using LeanCode.CQRS.Execution;
+using LeanCode.CQRS.OutputCaching.BasePolicies;
+using Microsoft.AspNetCore.OutputCaching;
+
+namespace LeanCode.CQRS.OutputCaching.Registration;
+
+internal sealed class CQRSOutputCacheRegistry : ICQRSEndpointMetadataProvider
```
jakubfijalkowski, line 9: `public` first, then `private`/`internal`s.

src/CQRS/LeanCode.CQRS.OutputCaching/Registration/CQRSOutputCacheRegistry.cs
```diff
@@ -0,0 +1,91 @@
...
+    {
+        return policyType
+            .GetInterfaces()
+            .Where(i => i.IsConstructedGenericType && i.GetGenericTypeDefinition() == typeof(ICQRSOutputCachePolicy<>));
+    }
+
+    internal void Register(Type contractType, Type policyType)
+    {
+        ArgumentNullException.ThrowIfNull(contractType);
+        ArgumentNullException.ThrowIfNull(policyType);
```
jakubfijalkowski: No, please, no. We have NRT for that ;P

src/CQRS/LeanCode.CQRS.OutputCaching/BasePolicies/CQRSOperationOutputCachePolicy.cs
```diff
@@ -0,0 +1,14 @@
...
+using Microsoft.AspNetCore.Http;
+
+namespace LeanCode.CQRS.OutputCaching.BasePolicies;
+
+public abstract class CQRSOperationOutputCachePolicy<TOperation, TResult>
+    : CQRSOutputCachePolicy<TOperation>,
+        ICQRSOperationOutputCachePolicy<TOperation, TResult>
+    where TOperation : IOperation<TResult>
+{
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1000")]
```
jakubfijalkowski: I don't think we should ignore this, plus `Justification` is missing .

src/CQRS/LeanCode.CQRS.OutputCaching/BasePolicies/CQRSOutputCachePolicy.cs
```diff
@@ -0,0 +1,29 @@
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.OutputCaching;
+
+namespace LeanCode.CQRS.OutputCaching.BasePolicies;
+
+[SuppressMessage("Microsoft.Naming", "CA1725")]
```
jakubfijalkowski: Why?

src/CQRS/LeanCode.CQRS.OutputCaching/BasePolicies/ICQRSOperationOutputCachePolicy.cs
```diff
@@ -0,0 +1,15 @@
+using LeanCode.Contracts;
+using LeanCode.CQRS.Execution;
+using Microsoft.AspNetCore.Http;
+
+namespace LeanCode.CQRS.OutputCaching.BasePolicies;
+
+public interface ICQRSOperationOutputCachePolicy<TOperation, TResult> : ICQRSOutputCachePolicy<TOperation>
```
jakubfijalkowski: I don't think we need this interface. It does not add any safety, especially since you can use `ICQRSOutputCachePolicy` either way.

src/CQRS/LeanCode.CQRS.OutputCaching/BasePolicies/ICQRSQueryOutputCachePolicy.cs
```diff
@@ -0,0 +1,15 @@
+using LeanCode.Contracts;
+using LeanCode.CQRS.Execution;
+using Microsoft.AspNetCore.Http;
+
+namespace LeanCode.CQRS.OutputCaching.BasePolicies;
+
+public interface ICQRSQueryOutputCachePolicy<TQuery, TResult> : ICQRSOutputCachePolicy<TQuery>
```
jakubfijalkowski: Same here

src/CQRS/LeanCode.CQRS.OutputCaching/BasePolicies/CQRSOperationOutputCachePolicy.cs
```diff
@@ -0,0 +1,14 @@
+using LeanCode.Contracts;
+using Microsoft.AspNetCore.Http;
+
+namespace LeanCode.CQRS.OutputCaching.BasePolicies;
+
+public abstract class CQRSOperationOutputCachePolicy<TOperation, TResult>
```
jakubfijalkowski: And I think we should get rid of these two base classes, with the same argument as for `ICQRSOperationOutputCachePolicy`.

src/CQRS/LeanCode.CQRS.OutputCaching/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+        TypesCatalog catalog,
+        bool registerOutputCache = true
+    )
+    {
+        return services.AddCQRSOutputCache(catalog.Assemblies, registerOutputCache);
+    }
+
+    public static IServiceCollection AddCQRSOutputCache(
+        this IServiceCollection services,
+        IEnumerable<Assembly> assemblies,
```
jakubfijalkowski: The `TypesCatalog` one should be the only overload. We introduced that type to _not_ have to deal with raw assemblies.

docs/cqrs/output_caching/index.md
```diff
@@ -0,0 +1,552 @@
...
+| --- | ----------- | ----------- |
+| LeanCode.CQRS.OutputCaching | [![NuGet version (LeanCode.CQRS.OutputCaching)](https://img.shields.io/nuget/vpre/LeanCode.CQRS.OutputCaching.svg?style=flat-square&logo=nuget)](https://www.nuget.org/packages/LeanCode.CQRS.OutputCaching) | Output caching policies |
+
+## Why Output Caching?
+
+- **Performance improvement:** Significantly reduces response times for frequently requested data by serving cached responses.
+- **Reduced database load:** Minimizes database queries for cacheable data, allowing your system to handle more concurrent users.
+- **Bandwidth savings:** Supports HTTP conditional requests (ETags, Last-Modified) to minimize data transfer.
+- **Type-safe policies:** Define caching behavior per query/operation with full access to **deserialized** request and response payloads, not just HTTP-level data.
+- **Correct pipeline integration:** The middleware is automatically placed at the right position in the CQRS pipeline, ensuring cache policies have access to the fully processed CQRS request object.
```
jakubfijalkowski: Nit: this does not describe _why_ output cache. This describes the implementation. Would remove.

docs/cqrs/output_caching/index.md
```diff
@@ -0,0 +1,552 @@
...
+
+```csharp
+public override void ConfigureServices(IServiceCollection services)
+{
+    // Register CQRS
+    services.AddCQRS(TypesCatalog.Of<ExampleQuery>(), TypesCatalog.Of<ExampleHandler>());
+
+    // Register output caching with automatic policy discovery
+    services.AddCQRSOutputCache(
+        catalog: TypesCatalog.Of<ExampleQuery>(),
```
jakubfijalkowski: nit:

```suggestion
        catalog: TypesCatalog.Of<ExampleHandler>(), // Catalog with the policies
```

docs/cqrs/output_caching/index.md
```diff
@@ -0,0 +1,552 @@
...
+
+The `AddCQRSOutputCache` method:
+
+- Automatically discovers all cache policies in the specified assemblies
+- Registers them with the DI container
+- Configures the ASP.NET Core OutputCache middleware to use them
+- Adds endpoint metadata to apply caching to the appropriate CQRS endpoints
+
+!!! warning "Cache Policies Required"
+    **Important:** Queries and operations are **only cached** if they have a corresponding cache policy class that implements `ICQRSOutputCachePolicy<T>` for their respecting type `T`. The `AddCQRSOutputCache` method scans the provided assemblies for policy implementations. Without a policy, a query/operation will execute normally even if `.CacheOutput()` is in the pipeline.
```
jakubfijalkowski:
```suggestion
    **Important:** Queries and operations are **cached only if** they have a corresponding cache policy class that implements `ICQRSOutputCachePolicy<T>` for their respective type `T`. The `AddCQRSOutputCache` method scans the provided assemblies for policy implementations. Without a policy, a query/operation will execute normally even if `.CacheOutput()` is in the pipeline.
```

docs/cqrs/output_caching/index.md
```diff
@@ -0,0 +1,552 @@
...
+```
+
+!!! danger "Do NOT Use app.UseOutputCache() for CQRS Endpoints"
+    You **must not** call `app.UseOutputCache()` globally when using output caching with CQRS endpoints. The `.CacheOutput()` method in the CQRS pipeline automatically adds the OutputCache middleware at the correct position in the pipeline - **after** the CQRS request payload has been deserialized and is available to cache policies. Adding `app.UseOutputCache()` globally would add the middleware in the wrong place (before CQRS deserialization), preventing cache policies from accessing the request payload and breaking the caching functionality.
+
+!!! tip "CacheOutput() Placement"
+    The `.CacheOutput()` method should typically be called after security middlewares. This ensures that only authorized requests may get cached responses, so that reduces a class of potential bugs.
+
+!!! tip "Mixed Scenarios: CQRS + Other Endpoints"
+    If you need output caching for **both** CQRS endpoints and non-CQRS endpoints (e.g., minimal APIs, controllers), use `app.UseWhen()` to apply the middleware **only** to non-CQRS endpoints:
```
jakubfijalkowski: TBH, instead of `UseWhen`, I would suggest [Map](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.mapextensions.map?view=aspnetcore-9.0#microsoft-aspnetcore-builder-mapextensions-map(microsoft-aspnetcore-builder-iapplicationbuilder-microsoft-aspnetcore-http-pathstring-system-action((microsoft-aspnetcore-builder-iapplicationbuilder)))), and suggest `UseWhen` only as an last-resort alternative.

docs/cqrs/output_caching/index.md
```diff
@@ -0,0 +1,552 @@
...
+        {
+            context.AllowCacheStorage = false;
+        }
+
+        return ValueTask.CompletedTask;
+    }
+}
+```
+
+## Policy Lifecycle Methods
```
jakubfijalkowski, line 305: I'm not sure if we should describe how the Output Cache works. We should link to official docs.

docs/cqrs/output_caching/index.md
```diff
@@ -0,0 +1,552 @@
...
+    );
+}
+```
+
+### Global Cache Settings
+
+Configure default settings for all cache policies using `IConfigureOptions<OutputCacheOptions>`:
+
+```csharp
+services.Configure<OutputCacheOptions>(options =>
```
jakubfijalkowski: `AddOutputCache` should take this (optional) configurator.

docs/cqrs/output_caching/index.md
```diff
@@ -0,0 +1,552 @@
...
+- **Test cache behavior:** Verify that cache invalidation and variation work as expected.
+
+### ❌ Don't
+
+- **Don't use `app.UseOutputCache()` globally:** Never add this middleware to the main pipeline when you have CQRS endpoints. Use `.CacheOutput()` in the CQRS pipeline instead. If you need caching for non-CQRS endpoints, use `app.UseWhen()` to apply it conditionally.
+- **Don't cache commands:** Commands change state and caching them is not supported.
+- **Don't cache operations with side effects that must execute every time:** If an operation has important side effects (e.g., logging, analytics, notifications) that need to happen on every call, it should not be cached. Only cache operations where the primary value is the returned data.
+- **Don't cache user-specific data without varying by user:** Always vary by user ID for personalized data.
+- **Don't set very long expiration times:** Stale data can cause confusion and bugs.
+- **Don't cache sensitive data:** Be careful with responses containing PII or sensitive information.
```
jakubfijalkowski: I miss the `why` here.

docs/cqrs/output_caching/index.md
```diff
@@ -0,0 +1,552 @@
...
+- **Don't assume local execution with output cache works:** Output caching only works with HTTP requests.
```
jakubfijalkowski: And if I assume that, what will happen?

src/CQRS/LeanCode.CQRS.OutputCaching/BasePolicies/CQRSOperationOutputCachePolicy.cs
```diff
@@ -0,0 +1,14 @@
...
+    : CQRSOutputCachePolicy<TOperation>,
+        ICQRSOperationOutputCachePolicy<TOperation, TResult>
+    where TOperation : IOperation<TResult>
+{
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1000")]
+    public static TResult? GetResultPayload(HttpContext context) =>
```
jakubfijalkowski: The more I think about it, the more these `GetRequesPayload`/`GetResultPayload` should be extension methods on the `HttpContext` or even `OutputCacheContext` (or [extension properties](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods), although not sure if properties can be generic xd). `GetResultPayload(context.HttpContext)` is no shorter/easier to write than `context.GetRequest<MyQuery>()`.

jakubfijalkowski (commented):
I have a problem with all the `internal`s exposed for tests. We shouldn't be testing internal implementations; we should be testing (public) behaviors. Otherwise, any change to the code would require us to rewrite the tests. [Tests shouldn't test the internals](https://github.com/leancodepl/leanpipe/pull/16#discussion_r1322746785), they should test public behaviors.

test/CQRS/LeanCode.CQRS.OutputCaching.Tests/CQRSOutputCacheOptionsConfiguratorTests.cs
Dragemil: Oh okay. Yeah it wasn't easy to avoid reflection here, but if it's acceptable, then sure.

src/CQRS/LeanCode.CQRS.OutputCaching/BasePolicies/CQRSOperationOutputCachePolicy.cs
Dragemil: I don't see why is that a problem. What we have here is quite convenient, I'd say.

src/CQRS/LeanCode.CQRS.OutputCaching/BasePolicies/CQRSOutputCachePolicy.cs
Dragemil: We have to break _some_ rule here. It's either that we shouldn't change the name of the interface parameter (`CA1725`) or that the `CancellationToken` should be named `cancellationToken` - our rule.

src/CQRS/LeanCode.CQRS.OutputCaching/BasePolicies/ICQRSOperationOutputCachePolicy.cs
Dragemil: We'll it does expose the convenience extension for getting the result, if you don't want to use the base class.

src/CQRS/LeanCode.CQRS.OutputCaching/ServiceCollectionExtensions.cs
Dragemil: Tbh I've never really understood why we actually have the `TypesCatalog`. And I still don't I guess? Alright though.

docs/cqrs/output_caching/index.md
Dragemil: orr maybe let's rename the section to _"Why Output Caching in CQRS?"_ ?

docs/cqrs/output_caching/index.md
Dragemil: Why though?

docs/cqrs/output_caching/index.md
Dragemil: We link to the official docs as well. TBH this is not explained in the official docs IIRC. Also, this output caching middleware seems pretty stable, hasn't been touched since 8.0.

docs/cqrs/output_caching/index.md
Dragemil: The local pipeline will fail if the response is returned from cache.

docs/cqrs/output_caching/index.md
Dragemil: Well it's not a hard `don't` after all. I'll remove this.

src/CQRS/LeanCode.CQRS.OutputCaching/BasePolicies/CQRSOperationOutputCachePolicy.cs
Dragemil: I guess they could be there as well, but why make the developer type the required type if it can be easily inferred?

src/CQRS/LeanCode.CQRS.OutputCaching/ServiceCollectionExtensions.cs
Dragemil: You're probably right

Dragemil:
> I have a problem with all the `internal`s exposed for tests. We shouldn't be testing internal implementations; we should be testing (public) behaviors. Otherwise, any change to the code would require us to rewrite the tests. [Tests shouldn't test the internals](https://github.com/leancodepl/leanpipe/pull/16#discussion_r1322746785), they should test public behaviors.

Well, I think it's generally right, but it's not some golden rule. It's sometimes convenient to make sure something works on a lower level. It's not like it's the first time in the corelib.

But let's say fine, and what would you do in this particular case? Remove all unit tests and stick only to the integration tests? I'd say it's more convenient to make sure proper assembly scanning works on a lower level, and not pollute the integration tests which test the general functionality.

test/CQRS/LeanCode.CQRS.OutputCaching.Tests/CQRSOutputCacheOptionsConfiguratorTests.cs
jakubfijalkowski: Yeah, tests are the place where _some_ rules can be broken.

docs/cqrs/output_caching/index.md
jakubfijalkowski: I would note that in the text - better to reiterate a couple of times. :P

docs/cqrs/output_caching/index.md
jakubfijalkowski: That's a good point - let's leave it then.

src/CQRS/LeanCode.CQRS.OutputCaching/ServiceCollectionExtensions.cs
jakubfijalkowski: As a wrapper for scanning multiple assemblies - _extending_ it so that it is functional (because, in the end, it lacks the _scanning_ capabilities) is in the roadmap, but it always got deprioritized.

src/CQRS/LeanCode.CQRS.OutputCaching/BasePolicies/CQRSOutputCachePolicy.cs
jakubfijalkowski: Aah, so that thing has been renamed. I personally prefer the CA1725 and would stick to it, silencing our rule.

docs/cqrs/output_caching/index.md
jakubfijalkowski: Maybe? The thing I'm referring to is that the last two points (missed that one previously) describe a totally different property than the first three.
"Performance ...", "Reduced ...", "Bandwidth ..." describe advantages of the caching, whereas "type-safe ..."/"correct ..." describe the properties of this particular implementation, not output caching. I wouldn't put them in the same category.

src/CQRS/LeanCode.CQRS.OutputCaching/BasePolicies/CQRSOperationOutputCachePolicy.cs
jakubfijalkowski: The "easily inferred" part makes the callsite awkward, so IMO that is not worth it.

src/CQRS/LeanCode.CQRS.OutputCaching/BasePolicies/ICQRSOperationOutputCachePolicy.cs
jakubfijalkowski: Too many things to do the same. This small convenience does not justify multiple options.

jakubfijalkowski:
> It's sometimes convenient to make sure something works on a lower level.

It is always convenient, but not worth the tradeoffs - you're making your implementation detail _very rigid_ for virtually no benefit. The tests of the internal part are very... simple. And much longer than the code they are testing. It is _not worth it_ to keep.

> Remove all unit tests and stick only to the integration tests? I'd say it's more convenient to make sure proper assembly scanning works on a lower level, and not pollute the integration tests which test the general functionality.

You can still test the same properties (if the policies are registered and if they match the handlers) using only public interfaces/classes and the final properties, without going into internal implementation.

You can "just" call `AddCQRSOutputCache` and see if proper policies are registered in the DI, then construct the DI and see if endpoint metadata gets populated correctly. It will be mostly the same amount of code without using any internal class.

src/CQRS/LeanCode.CQRS.OutputCaching/BasePolicies/CQRSOperationOutputCachePolicy.cs
Dragemil: Okay, I think I can roll with that.

test/CQRS/LeanCode.CQRS.OutputCaching.Tests/CQRSOutputCacheOptionsConfiguratorTests.cs
Dragemil: TBH now I remember that the registered policies are some internal aspnet core types as well, so more reflection would be needed, so I think current approach is better after all.
(https://github.com/dotnet/aspnetcore/blob/main/src/Middleware/OutputCaching/src/Policies/TypedPolicy.cs)

Orr maybe I shouldn't test at that level at all.

docs/cqrs/output_caching/index.md
Dragemil: Divided those categories.

jakubfijalkowski (approved):

PR #792, opened 2025-11-24
Add output caching observability & adjust CQRS approach to middleware spans

Dragemil:


src/CQRS/LeanCode.CQRS.AspNetCore/Middleware/CQRSExceptionTranslationMiddleware.cs
```diff
@@ -35,13 +35,13 @@ public async Task InvokeAsync(HttpContext httpContext)
             throw new InvalidOperationException("CQRSExceptionTranslationMiddleware may be used only for commands.");
         }
 
+        using var activity = LeanCodeActivitySource.StartMiddleware("ExceptionTranslation");
         try
         {
             await next(httpContext);
         }
         catch (CommandExecutionInvalidException ex)
         {
```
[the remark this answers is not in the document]
Dragemil: That one is kinda interesting, tbh both options are valid now, what do you think @jakubfijalkowski?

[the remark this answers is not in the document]
jakubfijalkowski: I don't think this is a sensible suggestion - the span should indicate that the middleware ran, not that it did some work. The same applies to other changes. I believe we should extend the spans to be present _always_, even if the middleware does virtually no work.

src/CQRS/LeanCode.CQRS.AspNetCore/Middleware/CQRSOutputCachingObservabilityMiddleware.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    public async Task InvokeAsync(HttpContext httpContext)
+    {
+        if (GetOutputCachingPolicyName(httpContext) is not { } policyName)
+        {
+            await next(httpContext);
+            return;
+        }
+
+        using var activity = LeanCodeActivitySource.StartMiddleware(CQRSOutputCachingConsts.MiddlewareName);
```
jakubfijalkowski: I think we should open the span at the beginning and "just" set that there is no caching policy tag.

src/CQRS/LeanCode.CQRS.AspNetCore/Middleware/CQRSSecurityMiddleware.cs
```diff
@@ -27,10 +27,17 @@ public async Task InvokeAsync(HttpContext context)
...
-        var user = context.User;
 
-        if (customAuthorizers.Count > 0 && !(user.Identity?.IsAuthenticated ?? false))
+        if (customAuthorizers.Count == 0)
         {
+            await next(context);
+            return;
+        }
+
+        using var activity = LeanCodeActivitySource.StartMiddleware("Security");
```
jakubfijalkowski: Ditto

src/CQRS/LeanCode.CQRS.AspNetCore/Middleware/CQRSValidationMiddleware.cs
```diff
@@ -32,32 +32,31 @@ public async Task InvokeAsync(HttpContext httpContext, ICommandValidatorResolver
         }
 
         var validator = resolver.FindCommandValidator(cqrsMetadata.ObjectType);
+        if (validator is null)
+        {
+            await next(httpContext);
+            return;
+        }
+
+        using var activity = LeanCodeActivitySource.StartMiddleware("Validation");
```
jakubfijalkowski: Ditto

src/CQRS/LeanCode.CQRS.OutputCaching/BasePolicies/CQRSOutputCachePolicy.cs
```diff
@@ -1,24 +1,75 @@
...
+        if (!context.EnableOutputCaching)
+        {
+            return;
+        }
+
+        var activity = Activity.Current;
+
+        if (
+            activity is null
+            || !activity.OperationName.Contains(
```
jakubfijalkowski, line 48: Hmm... What usecase does this cover?

Dragemil: Rather niche? That somebody uses somehow just the output caching middleware without its provided observability part introduced in this PR (skipping our extension method), and then we'd attach these tags on some random span.

src/CQRS/LeanCode.CQRS.AspNetCore/Middleware/CQRSExceptionTranslationMiddleware.cs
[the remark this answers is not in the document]
Dragemil: Yeah I guess, the other way we might encounter some weird unsatisfiable cases like this one.

src/CQRS/LeanCode.CQRS.OutputCaching/BasePolicies/CQRSOutputCachePolicy.cs
jakubfijalkowski: Ahh, got it. Makes sense. I would consider removing that and assuming it is used correctly, but I don't have any strong opinions here.

src/CQRS/LeanCode.CQRS.AspNetCore/Middleware/CQRSOutputCachingObservabilityMiddleware.cs
```diff
@@ -17,14 +17,15 @@ public CQRSOutputCachingObservabilityMiddleware(RequestDelegate next)
...
+        if (GetOutputCachingPolicyName(httpContext) is { } policyName)
         {
-            await next(httpContext);
-            return;
+            activity?.SetTag("output_cache.policy", CQRSOutputCachePolicyName.GetObjectTypeString(policyName));
+        }
+        else
+        {
+            activity?.SetTag("output_cache.policy_present", false);
```
jakubfijalkowski, line 28: Let's also set this (to `true`) in the success case, for consistency's sake.

src/CQRS/LeanCode.CQRS.AspNetCore/Middleware/CQRSSecurityMiddleware.cs
```diff
@@ -23,18 +23,19 @@ public CQRSSecurityMiddleware(CQRSMetrics metrics, RequestDelegate next, ILogger
...
     {
+        using var activity = LeanCodeActivitySource.StartMiddleware("Security");
         var cqrsMetadata = context.GetCQRSObjectMetadata();
         var payload = context.GetCQRSRequestPayload();
 
         var customAuthorizers = AuthorizeWhenAttribute.GetCustomAuthorizers(cqrsMetadata.ObjectType);
 
         if (customAuthorizers.Count == 0)
         {
+            activity?.SetTag("authorizers_present", false);
```
jakubfijalkowski: The name is somewhat inconsistent with the output cache, where we have `output_cache.sometag` and here it is just `sometag`. I would say let's pick one and stick to it (I dunno which is more common in the wild).

jakubfijalkowski: Plus, the same "consistency" argument as in output cache applies to this flag being present in both paths.

jakubfijalkowski (approved):

Dragemil:
Uhoh, seems like some tests are flaky as they pass on my end.

PR #799, opened 2025-12-01
Bump remaining packages

Dragemil:


Dragemil:
Ugh, another flaky test, created bug for that.

Directory.Packages.props
```diff
@@ -2,15 +2,15 @@
...
+    <EntityFrameworkCoreVersion>10.0.0</EntityFrameworkCoreVersion>
     <AspNetCoreVersion>10.0.0</AspNetCoreVersion>
     <ExtensionsVersion>10.0.0</ExtensionsVersion>
 
-    <MassTransitVersion>8.5.5</MassTransitVersion>
+    <MassTransitVersion>8.5.7</MassTransitVersion>
 
     <!-- Do not bump these dependencies if you don't want to force users to use newer .NET Core SDK -->
     <!-- Keep the major.minor values at exactly the one listed here: https://github.com/dotnet/roslyn/blob/main/docs/wiki/NuGet-packages.md -->
-    <CodeAnalysisVersion>4.12.0</CodeAnalysisVersion>
```
jakubfijalkowski, line 13: Which SDK version will this work with? And have we tested if that works (because they _looove_ to lie a little)?

Dragemil: Apparently the same used in the whole CoreLib:
> Version 5.0 includes C# 14.0 (Visual Studio 2026 version 18.0, .NET 10)

What approaches do we have to testing this? Wouldn't `LeanCode.CodeAnalysis.Tests` fail if something goes wrong?

Dragemil: And MS.Build 18.0 is being supported by the 10.0.1xx SDK.

jakubfijalkowski: > What approaches do we have to testing this? Wouldn't LeanCode.CodeAnalysis.Tests fail if something goes wrong?

Unfortunately, it probably won't - the tests are run in a slightly different environment than what the analyzers run on. :( The best way to test that is to use [this very short guide](https://github.com/leancodepl/corelibrary/blob/v10.0-preview/src/Tools/LeanCode.CodeAnalysis/local-debug.txt) or create a local NuGet feed and put the prebuilt package there.

> And MS.Build 18.0 is being supported by the 10.0.1xx SDK.

I had problems with that some time ago, especially on mixed-version-installs (i.e. having two 8.0.XXX SDKs), so let's do a quick test just to be sure. We can even make it "live", i.e. merge this and fix it only if the analyzer stops working.

Also, let's test in different setups - .NET CLI + Rider at least. I believe Rider might have some other SDK selectors implemented.

jakubfijalkowski (approved):

Directory.Packages.props
Dragemil: Aight, seems fine in both cases when tested locally

PR #808, opened 2025-12-16
Migrate solution to slnx

Dragemil:


LeanCode.CoreLibrary.slnx
jakubfijalkowski, line 1: While we're at it - can we ditch the platforms and per-platform config? I don't think it adds anything to CoreLib and was there just because some older tooling added it.

jakubfijalkowski (approved):

PR #809, opened 2025-12-17
String backed strongly typed IDs

Dragemil:
Closes [CORELIB-196](https://leancode.atlassian.net/browse/CORELIB-196)

[CORELIB-196]: https://leancode.atlassian.net/browse/CORELIB-196?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/domain/id/index.md
```diff
@@ -87,6 +91,21 @@ public readonly partial record struct VeryLongEmployeeId;
 // The `VeryLongEmployeeId` will have format `employee_(guid)`, with `New` using `Guid.NewGuid` as random source.
 ```
 
+```cs
+[TypedId(TypedIdFormat.RawString, MaxValueLength = 100)]
+public readonly partial record struct ExternalId;
+
+// The `ExternalId` wraps any string up to 100 characters. Exposes `MaxLength` static property.
```
jakubfijalkowski, line 98: Shouldn't we expose `MaxValueLength` to make it consistent with the prefixed variant?

Dragemil: Not sure about this, the `Value` part of the name doesn't make much sense in this context, and consistency between types of ids doesn't really give us much. It is consistent with the attribute property name though. TBH I could see us going with both ways, and I've already backed down from the name you're proposing, but I can bring it back, if you think it makes more sense.

jakubfijalkowski (approved):

docs/domain/id/index.md
jakubfijalkowski: Naah, if you thought it out, then I agree with it - this was my initial impression, but I haven't put too much thought into it.

Dragemil:
Another flaky test, added a bug

src/Domain/LeanCode.DomainModels.EF/PropertiesConfigurationBuilderExtensions.cs
```diff
@@ -54,15 +54,34 @@ this PropertiesConfigurationBuilder<TId?> builder
...
+    }
+
     public static PropertiesConfigurationBuilder<TId> ArePrefixedTypedId<TId>(
         this PropertiesConfigurationBuilder<TId> builder
     )
         where TId : struct, IPrefixedTypedId<TId>
     {
         return builder
             .HaveConversion<PrefixedTypedIdConverter<TId>, PrefixedTypedIdComparer<TId>>()
-            .HaveMaxLength(TId.RawLength)
```
jakubfijalkowski, line 64: I would leave the `AreFixedLength` for the back-compat purposes or ensure that the resulting migration would work for PKs on both SQL Server & Postgres.

jakubfijalkowski (approved):

PR #810, opened 2025-12-19
Do not cache unsuccesful responses by default

Dragemil:
Closes [CORELIB-390](https://leancode.atlassian.net/browse/CORELIB-390)

[CORELIB-390]: https://leancode.atlassian.net/browse/CORELIB-390?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/CQRS/LeanCode.CQRS.OutputCaching/BasePolicies/CQRSOutputCachePolicy.cs
+++ b/src/CQRS/LeanCode.CQRS.OutputCaching/BasePolicies/CQRSOutputCachePolicy.cs
@@ -1,4 +1,5 @@
 using System.Diagnostics;
+using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.OutputCaching;
 
 namespace LeanCode.CQRS.OutputCaching.BasePolicies;
@@ -28,8 +29,20 @@ public abstract class CQRSOutputCachePolicy<TObject> : ICQRSOutputCachePolicy<TO
     }
 
     /// <inheritdoc />
+    /// <remarks>
+    /// Unsuccessful responses with status codes other than 2xx are not cached by default.
+    /// </remarks>
     public virtual ValueTask ServeFromCacheAsync(OutputCacheContext context, CancellationToken cancellation)
     {
+        if (
+            context.HttpContext.Response.StatusCode
+            is < StatusCodes.Status200OK
+                or >= StatusCodes.Status300MultipleChoices
+        )
+        {
+            context.AllowCacheStorage = false;
+        }
+
         RecordSpanTagsAndMetrics(true, context);
         return ValueTask.CompletedTask;
     }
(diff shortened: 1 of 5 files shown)


src/CQRS/LeanCode.CQRS.OutputCaching/BasePolicies/CQRSOutputCachePolicy.cs
```diff
@@ -22,19 +29,33 @@ public async ValueTask CacheRequestAsync(OutputCacheContext context, Cancellatio
         }
     }
 
+    /// <inheritdoc />
+    /// <remarks>
+    /// Unsuccessful responses with status codes other than 2xx are not cached by default.
```
jakubfijalkowski: Nit: 3xx is a successful response :)

src/CQRS/LeanCode.CQRS.OutputCaching/BasePolicies/CQRSOutputCachePolicy.cs
```diff
@@ -22,19 +29,33 @@ public async ValueTask CacheRequestAsync(OutputCacheContext context, Cancellatio
...
+        {
+            context.AllowCacheStorage = false;
+        }
+
         RecordSpanTagsAndMetrics(false, context);
         return ValueTask.CompletedTask;
     }
 
-    private static void RecordSpanTagsAndMetrics(bool servingFromCache, OutputCacheContext context)
+    protected static void RecordSpanTagsAndMetrics(bool servingFromCache, OutputCacheContext context)
```
jakubfijalkowski, line 58: Why the change?

Dragemil: I've figured if someone wants to override the `ServeResponseAsync` without running the base version, cause they'd like to cache the >=300 responses, they would still have the possibility to record the span tags and metrics.

src/CQRS/LeanCode.CQRS.OutputCaching/BasePolicies/CQRSOutputCachePolicy.cs
Dragemil: Kinda true, but its specified later what is meant by that and HTTP literally calls 200<=x<300 [successful responses](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status#successful_responses).

jakubfijalkowski: Then it is all about the semantics of what "unsuccessful" means. If going that direction, then there are no "unsuccessful" status codes at all, only informational, successful, redirects, client & server errors :P

I am pretty sure that someone will understand "unsuccessful" as 4xx & 5xx and will be very surprised when they find out that is not true (because redirects should be cached). :P

Let's remove the `unsuccessful` word completely and call it a day.

---

This was a very nice yak shaving, I have to say.

jakubfijalkowski (approved):

PR #821, opened 2026-01-09
[CORELIB-392] RelationalTypeMapping for typed ids

marcin-dardzinski:
This will enable us to use typed ids in raw SQL queries. The mapping plugin, with some additional work can replace the conventions completely, though right now they coexist

Also added a helper for registering conventions for all typed ids in the assembly.

### DIFF
--- a/src/Domain/LeanCode.DomainModels.EF/DbContextOptionsBuilderExtensions.cs
+++ b/src/Domain/LeanCode.DomainModels.EF/DbContextOptionsBuilderExtensions.cs
@@ -1,3 +1,4 @@
+using LeanCode.DomainModels.Ids;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.EntityFrameworkCore.Storage;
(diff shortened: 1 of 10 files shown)


src/Domain/LeanCode.DomainModels.EF/DbContextOptionsBuilderExtensions.cs
```diff
@@ -0,0 +1,75 @@
...
+        ((IDbContextOptionsBuilderInfrastructure)builder).AddOrUpdateExtension(
+            new TypedIdDbContextOptionsExtension(new PostgresTypedIdStoreTypeProvider())
+        );
+
+        return builder;
+    }
+
+    /// <summary>
+    /// Automatically registers TypeMappingPlugin for all TypedIds.
+    /// Supports types implementing IPrefixedTypedId and IRawTypedId with any backing type (int, long, Guid).
```
jakubfijalkowski: Slightly misleading from what I see - the string types are also supported.

jakubfijalkowski:
Does this replace the `Configure*Id` methods?

marcin-dardzinski:
> Does this replace the `Configure*Id` methods?

The relational type mapping? At this moment not completely, both are needed, though I think it should be possible to get rid of conventions with some more effort?

jakubfijalkowski (approved):

PR #829, opened 2026-01-23
Add JsonSerializerOptions to AuditLogs

piotrek999:


lukaszgarstecki (approved):

src/Infrastructure/LeanCode.AuditLogs/AuditLogsExtensions.cs
```diff
@@ -6,10 +7,15 @@ public static class AuditLogsExtensions
...
-        AzureBlobAuditLogStorageConfiguration config
+        AzureBlobAuditLogStorageConfiguration config,
+        JsonSerializerOptions? jsonSerializerOptions = null
     )
     {
         services.AddSingleton(config);
+        if (jsonSerializerOptions is not null)
+        {
+            services.AddSingleton(jsonSerializerOptions);
+        }
```
Saancreed, line 20: Not sure if this is a good idea, it may conflict with existing registration of a different JSO. Maybe use a keyed service instead?

Saancreed (approved):

PR #842, opened 2026-02-26
Create ServiceToService package to standardize S2S calls authorization

Dragemil:
An attempt to standardize service to service authentication and authorization, using for authorization the same patterns that are used for the end client calls. Comments regarding anything are welcome.

The authentication heavily relies on some infrastructural entity (eg. an Istio service mesh) to enforce that incoming requests from all allowed services have a particular header with appropriate value attached, and I know this is possible. Without it (or with some misconfiguration of it), it's obviously very insecure, as such thing could be very easily spoofed.

On properly configured infrastructure, a request without the appropriate header would be short circuited with 403 status by the proxy, never reaching the destination service. It is also assumed that the external ingress will cut out and provide correct `LNCD-Caller-Id` headers.

This implementation is mainly concerned by the authorization on quite granular, application level, while also providing some defense in depth, thanks to not accepting requests from unknown services, even though it is mainly the infrastructure responsibility.

It works out of the box on the local envs, it's just not secure at all there, but we don't need the security there.

Some accompanying infrastructure with this would look a bit like this:
https://github.com/leancodepl/terraform-kubernetes-cluster/pull/43
https://github.com/leancodepl/exampleapp/pull/386
https://github.com/leancodepl/exampleapp-deployment/pull/37

### DIFF
--- a/src/Infrastructure/LeanCode.ServiceToService/Incoming/ServiceToServiceAuthenticationHandler.cs
+++ b/src/Infrastructure/LeanCode.ServiceToService/Incoming/ServiceToServiceAuthenticationHandler.cs
@@ -36,12 +36,7 @@ public partial class ServiceToServiceAuthenticationHandler(
                 return Task.FromResult(AuthenticateResult.Fail("Multiple S2S API key headers are not allowed."));
             }
 
-            if (Options.RejectMissingS2SApiKey)
-            {
-                return Task.FromResult(AuthenticateResult.Fail("Missing S2S API key header."));
-            }
-
-            return Task.FromResult(AuthenticateResult.NoResult());
+            return Task.FromResult(AuthenticateResult.Fail("Missing S2S API key header."));
         }
 
         if (!VerifyApiKey(s2SApiKey, Options.S2SApiKey))
@@ -64,12 +59,7 @@ public partial class ServiceToServiceAuthenticationHandler(
                 return Task.FromResult(AuthenticateResult.Fail("Multiple caller identity headers are not allowed."));
             }
 
-            if (Options.RejectMissingCallerId)
-            {
-                return Task.FromResult(AuthenticateResult.Fail("Missing caller identity header."));
-            }
-
-            return Task.FromResult(AuthenticateResult.NoResult());
+            return Task.FromResult(AuthenticateResult.Fail("Missing caller identity header."));
         }
 
         if (!Options.CallerRoles.TryGetValue(callerId, out var roles))
--- a/src/Infrastructure/LeanCode.ServiceToService/Incoming/ServiceToServiceAuthenticationHandler.cs
+++ b/src/Infrastructure/LeanCode.ServiceToService/Incoming/ServiceToServiceAuthenticationHandler.cs
@@ -1,0 +1,72 @@
+using System.Security.Claims;
+using System.Text.Encodings.Web;
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace LeanCode.ServiceToService.Incoming;
+
+/// <summary>
+/// Authenticates callers based on <c>LNCD-Caller-Id</c>.
+/// Security depends on infrastructure-level header enforcement (for example service mesh or ingress middleware).
+/// If the header is not enforced and can be set by arbitrary clients, caller identity can be spoofed.
+/// </summary>
+public partial class ServiceToServiceAuthenticationHandler(
+    IOptionsMonitor<ServiceToServiceAuthenticationOptions> options,
+    ILoggerFactory logger,
+    UrlEncoder encoder
+) : AuthenticationHandler<ServiceToServiceAuthenticationOptions>(options, logger, encoder)
+{
+    protected override Task<AuthenticateResult> HandleAuthenticateAsync()
+    {
+        if (
+            !Request.Headers.TryGetValue(ServiceToServiceDefaults.CallerIdHeaderName, out var values)
+            || string.IsNullOrWhiteSpace(values.ToString())
+        )
+        {
+            if (Options.RejectMissingCallerId)
+            {
+                LogMissingCallerIdHeader(Logger, ServiceToServiceDefaults.CallerIdHeaderName);
+                return Task.FromResult(AuthenticateResult.Fail("Missing caller identity header."));
+            }
+
+            return Task.FromResult(AuthenticateResult.NoResult());
+        }
+
+        var callerId = values.ToString();
+
+        if (!Options.CallerRoles.TryGetValue(callerId, out var roles))
+        {
+            if (Options.RejectUnknownCallers)
+            {
+                LogUnknownServiceCaller(Logger, callerId);
+                return Task.FromResult(AuthenticateResult.Fail($"Unknown service caller: {callerId}"));
+            }
+
+            return Task.FromResult(AuthenticateResult.NoResult());
+        }
+
+        var claims = new List<Claim> { new(Options.NameClaimType, callerId) };
+        claims.AddRange(roles.Select(role => new Claim(Options.RoleClaimType, role)));
+
+        var identity = new ClaimsIdentity(claims, Scheme.Name, Options.NameClaimType, Options.RoleClaimType);
+        var ticket = new AuthenticationTicket(new(identity), Scheme.Name);
+
+        LogAuthenticatedServiceCaller(Logger, callerId, string.Join(", ", roles));
+
+        return Task.FromResult(AuthenticateResult.Success(ticket));
+    }
+
+    [LoggerMessage(EventId = 1, Level = LogLevel.Warning, Message = "Missing {Header} header")]
+    private static partial void LogMissingCallerIdHeader(ILogger logger, string header);
+
+    [LoggerMessage(EventId = 2, Level = LogLevel.Warning, Message = "Unknown service caller: {CallerId}")]
+    private static partial void LogUnknownServiceCaller(ILogger logger, string callerId);
+
+    [LoggerMessage(
+        EventId = 3,
+        Level = LogLevel.Debug,
+        Message = "Authenticated s2s caller {CallerId} with roles [{Roles}]"
+    )]
+    private static partial void LogAuthenticatedServiceCaller(ILogger logger, string callerId, string roles);
+}
--- a/src/Infrastructure/LeanCode.ServiceToService/Incoming/ServiceToServiceAuthenticationHandler.cs
+++ b/src/Infrastructure/LeanCode.ServiceToService/Incoming/ServiceToServiceAuthenticationHandler.cs
@@ -1,4 +1,6 @@
+using System.Runtime.InteropServices;
 using System.Security.Claims;
+using System.Security.Cryptography;
 using System.Text.Encodings.Web;
 using Microsoft.AspNetCore.Authentication;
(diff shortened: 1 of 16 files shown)


src/Infrastructure/LeanCode.ServiceToService/Incoming/ServiceToServiceAuthenticationHandler.cs
```diff
@@ -0,0 +1,72 @@
...
+            if (Options.RejectUnknownCallers)
+            {
+                LogUnknownServiceCaller(Logger, callerId);
+                return Task.FromResult(AuthenticateResult.Fail($"Unknown service caller: {callerId}"));
+            }
+
+            return Task.FromResult(AuthenticateResult.NoResult());
+        }
+
+        var claims = new List<Claim> { new(Options.NameClaimType, callerId) };
```
marcin-dardzinski, line 76: This allows only to set name / role claims on the generated identity. Maybe we should allow more control here, e.g. instead `CallerRoles` we should have, `TryCreateClient(string clientId, out ClaimsIdentity)` or sth. 

With a helper / default method that just reads the roles from a dict.

src/Infrastructure/LeanCode.ServiceToService/Outgoing/ServiceToServiceHttpClientExtensions.cs
```diff
@@ -0,0 +1,13 @@
+using Microsoft.Extensions.DependencyInjection;
+
+namespace LeanCode.ServiceToService.Outgoing;
+
+public static class ServiceToServiceHttpClientExtensions
+{
+    public static IHttpClientBuilder AddCallerIdentity(this IHttpClientBuilder builder, string serviceName)
```
marcin-dardzinski: Maybe `AddServiceToServiceCallerIdentity`? 

Also, unless `ServiceToServiceDefaults` are really _defaults_ and not _consts_. I'd do sth like this `AddIdentity(IHttpClientBuilder builder, string callerId, string headerName = ServiceToServiceDefaults.CallerIdHeaderName)`

marcin-dardzinski (commented):
> Without it (or with some misconfiguration of it), it's obviously very insecure, as such thing could be very easily spoofed.

So is there anything the app can do about it now, or we just pray it's configured? 

> It works out of the box on the local envs, it's just not secure at all there, but we don't need the security there.

Istio on local clusters? :biting_lip:

src/Infrastructure/LeanCode.ServiceToService/Outgoing/ServiceToServiceHttpClientExtensions.cs
Dragemil: Fine for the name.

Well, the `ServiceToServiceDefaults` sometimes are _defaults_ (authorization scheme) and sometimes are _consts_ (header name).

src/Infrastructure/LeanCode.ServiceToService/Incoming/ServiceToServiceAuthenticationHandler.cs
Dragemil: Yeah, but would we ever need anything more? I guess maybe, but I'm trying to think of some examples.

Dragemil:
> So is there anything the app can do about it now, or we just pray it's configured?

Well' that's the main point of this RFC, cause I don't have any better idea. The good thing is that it's actually very easy to have a secure by default Istio setup, that will lean into disallowing requests you haven't explicitly allowed.

> Istio on local clusters? 🫦

Hell no.
I mean, we could, but I wouldn't absolutely go this way. The local envs should be more frictionless, rather than secure.

Dragemil:
> So is there anything the app can do about it now, or we just pray it's configured?

I wonder if this S2S auth shouldn't also check for some project wide secret, and only if it is correct, then check for the well  known header values. It would mitigate the ease of shooting one in the foot at least, in terms of just providing a header from the outside of the ingress that wouldn't be stripped.

jakubfijalkowski:
Maybe we can see if the Istio is running and the namespace is managed by Istio and fail if that is not the case? Plus allow that on development environments.

Dragemil:
> Maybe we can see if the Istio is running and the namespace is managed by Istio and fail if that is not the case? Plus allow that on development environments.

You would need to give the workload some additional k8s RBAC to check on the namespace label, and maybe some other roles to check whether some Istio pods are running. And it'd still be just a heuristic, since you probably wouldn't check if there is a default deny policy on the namespace (but if you'd check for that, then that would stop being just a heuristic I think).

So I think it's quite more complex solution, while still is just a heuristic, so I think i like the project x env api key more, since it's simpler and would work simply on the local env.

src/Infrastructure/LeanCode.ServiceToService/Incoming/ServiceToServiceAuthenticationHandler.cs
marcin-dardzinski: We might want to inject some other information here. E.g. if the s2s is a result of a user action, we might want to propagate calling user id in the claims. 

This is not impossible without it though, e.g. we may just read it directly from headers

src/Infrastructure/LeanCode.ServiceToService/Incoming/ServiceToServiceAuthenticationHandler.cs
```diff
@@ -0,0 +1,149 @@
...
+    {
+        value = string.Empty;
+        multipleValues = false;
+
+        if (!Request.Headers.TryGetValue(headerName, out var headerValues))
+        {
+            return false;
+        }
+
+        if (headerValues.Count > 1)
```
marcin-dardzinski: Nit: can the rest of this method just be a pattern match?

marcin-dardzinski (approved):

src/Infrastructure/LeanCode.ServiceToService/Incoming/ServiceToServiceAuthenticationHandler.cs
Dragemil: It is kind of awkward to pattern match against that.

src/Infrastructure/LeanCode.ServiceToService/Incoming/ServiceToServiceAuthenticationHandler.cs
Dragemil: We'll tackle this once we'll see what other projects actually need.

src/Infrastructure/LeanCode.ServiceToService/Incoming/ServiceToServiceAuthenticationHandler.cs
Dragemil: But I've managed something

src/Infrastructure/LeanCode.ServiceToService/Incoming/ServiceToServiceCallerRolesValidator.cs
```diff
@@ -0,0 +1,41 @@
...
+            errors.Add("S2SApiKey must be configured when ValidateS2SApiKeyAtStartup is enabled.");
+        }
+
+        if (options.ValidateCallerRolesAtStartup)
+        {
+            var availableRoles = roleRegistry.All.Select(role => role.Name).ToHashSet(StringComparer.Ordinal);
+            var unknownRoles = options
+                .CallerRoles.Values.SelectMany(roles => roles)
+                .Distinct(StringComparer.Ordinal)
+                .Where(role => !availableRoles.Contains(role))
```
Saancreed:
```suggestion
                .Except(availableRoles)
```

src/Infrastructure/LeanCode.ServiceToService/Incoming/ServiceToServiceAuthenticationHandler.cs
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    [LoggerMessage(EventId = 1, Level = LogLevel.Warning, Message = "Missing or multiple {Header} headers")]
+    private static partial void LogMissingOrMultipleS2SApiKeyHeader(ILogger logger, string header);
+
+    [LoggerMessage(EventId = 2, Level = LogLevel.Warning, Message = "Invalid S2S API key")]
+    private static partial void LogInvalidS2SApiKey(ILogger logger);
+
+    [LoggerMessage(EventId = 3, Level = LogLevel.Warning, Message = "Missing or multiple {Header} headers")]
+    private static partial void LogMissingOrMultipleCallerIdHeader(ILogger logger, string header);
```
Saancreed, line 120: This is only ever called with a single `const` value for `header`, same as `LogMissingOrMultipleS2SApiKeyHeader`. Can we perhaps merge them into a single helper?

src/Infrastructure/LeanCode.ServiceToService/Incoming/ServiceToServiceAuthenticationHandler.cs
```diff
@@ -0,0 +1,141 @@
...
+            {
+                return Task.FromResult(AuthenticateResult.Fail("Multiple caller identity headers are not allowed."));
+            }
+
+            if (Options.RejectMissingCallerId)
+            {
+                return Task.FromResult(AuthenticateResult.Fail("Missing caller identity header."));
+            }
+
+            return Task.FromResult(AuthenticateResult.NoResult());
```
Saancreed: What's the use case for this? I can't quite think of a situation where the caller has a valid api key but failed to include their identity in the request. I'd say it's a protocol error and should always result in a failure.

I'd say the same about `RejectMissingS2SApiKey` above because I'm not sure how you'd hit it `ForwardDefaultSelector` being defined the way it is, but oh well.

src/Infrastructure/LeanCode.ServiceToService/Incoming/ServiceToServiceAuthenticationHandler.cs
Dragemil: Thought about that, but then I thought that it's good to have this decoupled.

src/Infrastructure/LeanCode.ServiceToService/Incoming/ServiceToServiceAuthenticationHandler.cs
marcin-dardzinski: I've always had the impression that `NoResult` is the usual way to go, as it does not restrict other handlers from authorizing the request. `Fail` on the other hand is stricter and does not allow other potential handlers to authorize.

Dragemil: I agree with redundancy of `RejectMissingCallerId` option. I actually also agree with that `RejectMissingS2SApiKey` is also redundant, even though I think I can think of a case, where a project ditches the provided policy scheme, and uses one of their own (maybe selecting some service identity to be treated as ingress), but its too exotic to support this.

Dragemil: > I've always had the impression that NoResult is the usual way to go, as it does not restrict other handlers from authorizing the request. Fail on the other hand is stricter and does not allow other potential handlers to authorize.

There are no other handlers in ASP.NET, ever. Once you're authenticating with an authentication policy, it gets the final say. Only place to pick an authn handler is the policy scheme.

Saancreed (approved):

PR #856, opened 2026-03-18
Add error name

michal12334:


jakubfijalkowski (approved):
LGTM, but let's wait for v10 (we should finalize it) and start with v11

PR #891, opened 2026-06-17
Mention alternative docker compose quickstart method in docs

tomaszlaskowskiLC:
As agreed, example-app docker compose local dev option should be mentioned.

### DIFF
--- a/docs/quick_start/index.md
+++ b/docs/quick_start/index.md
@@ -4,8 +4,8 @@ LeanCode CoreLibrary offers flexibility to suit different needs. Whether you wan
 
 !!! Prerequisites info
     Prior to proceeding with this tutorial, ensure you have the following requirements installed: [kubectl](https://kubernetes.io/docs/tasks/tools/),
-    [K3D](https://k3d.io/#installation), [Helm](https://helm.sh/docs/intro/quickstart/), [Tilt](https://docs.tilt.dev/install.html), [Terraform](https://www.terraform.io/), [dotnet](https://dotnet.microsoft.com/en-us/download), [git](https://git-scm.com/).
-    If you want to use the Docker Compose based setup instead of the local Kubernetes cluster, install [Docker](https://docs.docker.com/get-docker/).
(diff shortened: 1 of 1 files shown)


docs/quick_start/index.md
```diff
@@ -5,6 +5,7 @@ LeanCode CoreLibrary offers flexibility to suit different needs. Whether you wan
 !!! Prerequisites info
     Prior to proceeding with this tutorial, ensure you have the following requirements installed: [kubectl](https://kubernetes.io/docs/tasks/tools/),
     [K3D](https://k3d.io/#installation), [Helm](https://helm.sh/docs/intro/quickstart/), [Tilt](https://docs.tilt.dev/install.html), [Terraform](https://www.terraform.io/), [dotnet](https://dotnet.microsoft.com/en-us/download), [git](https://git-scm.com/).
+    If you want to use the Docker Compose based setup instead of the local Kubernetes cluster, install [Docker](https://docs.docker.com/get-docker/).
```
Saancreed: Docker is technically required in both cases, as we're running k3s inside it in the first case anyway. But this path requires Compose plugin too, which is sometimes distributed separately. On Arch, it's a distinct `docker-compose` package, for example.

And for anyone crazy enough, this can also work with `podman` and `podman-compose`, although it requires some manual tweaks in Compose files. Not sure if it's worth pointing out, but The Compose Specification doesn't technically require Docker, it's just the reference implementation 🙂

(Also, shouldn't this also mention `az` / Azure CLI, required for access to ACR with our certificates?)

docs/quick_start/index.md
```diff
@@ -143,3 +144,12 @@ tilt up projectname-integration_tests
 [Seq]: https://datalust.co/seq
 [SignalR]: https://dotnet.microsoft.com/en-us/apps/aspnet/signalr
 [Traefik]: https://traefik.io/traefik/
+
+## Docker Compose
+
+As an alternative to the local Kubernetes setup, `example-app` also ships Docker Compose files in `backend/dev/compose`. From a generated project, you can start that environment with:
```
Saancreed: `example-app` → `exampleapp`

Saancreed (approved):

PR #892, opened 2026-09-10
Multi-target contracts projects (v10)

Saancreed:
Plus a minor fix to VersionSupport, see the commits. Allows one to stop depending on long-deprecated `net6.0` TFM.

I'll port this to v11 branch afterwards.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -26,7 +26,7 @@
     <ProjectReference
       Include="../../Tools/LeanCode.CodeAnalysis/LeanCode.CodeAnalysis.csproj"
       Condition="'$(MSBuildProjectName)' != 'LeanCode.CodeAnalysis'
-                                And '$(TargetFramework)' == 'net9.0'"
+                                And '$(TargetFramework)' != 'netstandard2.0'"
     >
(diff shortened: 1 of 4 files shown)


Komoszek (approved):

src/Directory.Build.props
```diff
@@ -26,7 +26,7 @@
     <ProjectReference
       Include="../../Tools/LeanCode.CodeAnalysis/LeanCode.CodeAnalysis.csproj"
       Condition="'$(MSBuildProjectName)' != 'LeanCode.CodeAnalysis'
-                                And '$(TargetFramework)' == 'net9.0'"
+                                And '$(TargetFramework)' != 'netstandard2.0'"
```
mishioo, line 29: Only not the particular version of netstandard is enough here? 🤔 2.1 is fine?

mishioo (approved):

ZoltePudeleczko (approved):

src/Directory.Build.props
Saancreed: Yes, netstandard2.1 is fine because our CodeAnalysis package itself targets netstandard2.1. Only referencing it in projects targeting older netstandards is a problem.

PR #893, opened 2026-09-10
Multi-target contracts projects (v11)

Saancreed:
Frontport of https://github.com/leancodepl/corelibrary/pull/892 (is that even a word? :thinking:), except with legacy `net6.0` swapped for current `net11.0`.


Saancreed:
Right, obviously this depends on us actually starting to use .NET 11 :see_no_evil:

tomaszlaskowskiLC (approved):

PR #896, opened 2026-09-15
Update to .NET 11

Saancreed:
Unfortunately, the Npgsql shenanigans which require us to pull packages from daily .NET package feed make me very reluctant to merge this before upstream resolves this in a reasonable manner. See the "Update integration tests to .NET 11 and Npgsql.EFCore to rc.1" and "Update Extensions version to satisfy EFCore requirements" for details.


src/Infrastructure/LeanCode.Firebase.FCM/FCMClient.cs
```diff
@@ -62,9 +62,9 @@ public virtual async Task SendToUserAsync(
         CancellationToken cancellationToken = default
     )
     {
-        message.Tokens = await tokenStore.GetTokensAsync(userId, cancellationToken);
+        message.Fids = await tokenStore.GetTokensAsync(userId, cancellationToken);
```
tomaszlaskowskiLC, line 65: TIL, that Token is getting obsolete. 😛 

Anyway, it's not obvious that you can pass `Tokens` as `Fids`.
If we would like to keep "Tokens" everywhere else and use `Fids` property in this client, we should add comment with an explanation here and mention it in the [docs](https://leancode-corelibrary.readthedocs.io/external_integrations/push_notifications_fcm/).

Also, let's ensure there is no changes required on the client side (I mean mobile apps 😛).

src/Infrastructure/LeanCode.Firebase/FirebaseConfiguration.cs
```diff
@@ -28,7 +28,7 @@ public static FirebaseApp Prepare(string? cfg, string name)
             return FirebaseApp.Create(
                 new AppOptions
                 {
-                    Credential = GoogleCredential.FromJson(cfg),
+                    Credential = CredentialFactory.FromJson<ServiceAccountCredential>(cfg).ToGoogleCredential(),
```
tomaszlaskowskiLC, line 31: NIT: technically this narrows accepted credentials to service accounts (although it should be the way-to-go here).
Might be worth noting in the docs.

Directory.Packages.props
```diff
@@ -97,36 +103,36 @@
     <PackageVersion Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="$(ExtensionsVersion)" />
 
     <PackageVersion Include="Microsoft.Extensions.Telemetry.Testing" Version="8.0.0-rc.2.23510.2" />
```
tomaszlaskowskiLC, line 105: No bump for this one?

test/LeanCode.IntegrationTests/PostgresOnlyFactAttribute.cs
```diff
@@ -3,11 +3,21 @@
 
 namespace LeanCode.IntegrationTests;
 
-public sealed class PostgresOnlyFactAttribute(
-    [CallerFilePath] string? sourceFilePath = null,
-    [CallerLineNumber] int sourceLineNumber = 0
-) : IntegrationFactAttribute(sourceFilePath, sourceLineNumber)
+public sealed class PostgresOnlyFactAttribute : IntegrationFactAttribute
 {
+    public static bool IsPostgres => Environment.GetEnvironmentVariable(TestDatabaseConfig.ConfigEnvName) == "postgres";
```
tomaszlaskowskiLC, line 8: NIT: I guess `IsPostgres` belongs more to `TestDatabaseConfig`.

src/Domain/LeanCode.DomainModels.EF/TimestampTzExpressionInterceptor.cs
```diff
@@ -44,7 +49,15 @@ protected override Expression VisitMember(MemberExpression node)
             return Expression.Call(
                 null,
                 ConvertDateTimeBySystemTimeZoneIdMethod,
-                Expression.Property(Expression.Property(node.Expression, UtcTimestampProperty), UtcDateTimeProperty),
+                // Npgsql 11 translates UtcDateTime as a no-op which retains DateTimeOffset as the SQL
+                // expression's CLR type. ToUniversalTime retags it as DateTime, which the outer translator requires.
```
tomaszlaskowskiLC, line 53: You can also link your issue here (shows blocker explicitly): https://github.com/npgsql/efcore.pg/issues/3912. 😄

tomaszlaskowskiLC (commented):
@Saancreed I agree on waiting until the package issue is resolved. Is there expected timeline for that?

Directory.Packages.props
Saancreed: This is already the newest version: https://www.nuget.org/packages/Microsoft.Extensions.Telemetry.Testing

I'm not sure what we need this for, but the package appears to be kind of abandoned. Didn't want to touch this during the bump though.

test/LeanCode.IntegrationTests/PostgresOnlyFactAttribute.cs
Saancreed: Sure, can move it there.

src/Domain/LeanCode.DomainModels.EF/TimestampTzExpressionInterceptor.cs
Saancreed: IIRC I committed that before I made this issue :upside_down_face:

src/Infrastructure/LeanCode.Firebase/FirebaseConfiguration.cs
Saancreed: Noting, yes. Notice how the changelog entries for v11 are still missing, it will be a good place to put it there once it actually materializes.

src/Infrastructure/LeanCode.Firebase.FCM/FCMClient.cs
Saancreed: Ah, good point. I initially found https://firebase.google.com/docs/projects/manage-installations#retrieving-an-fcm-registration-token which made me think "ok, so it's fine to keep using `FirebaseMessaging.getInstance().token`", but it appears that it's kinda misleading and this isn't actually related to migration to FIDs.

We may want to keep using `Tokens` for some time, as it's documented to also [accept FIDs during the migration period](https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages#Message)… whatever that is.

Saancreed:
> Is there expected timeline for that?

Nothing confirmed yet, but I'd expect rc.2 to address that. If not, we'll have to wait for 11.0.0 to be released.

